PR #43, opened 2024-03-05
Add external-dns(v0.14.0) as a dependency 

### LINKED ISSUE #47: Add external-dns as a dependency
**What**

Based on the results of this [poc task](https://github.com/Kuadrant/dns-operator/issues/26) , we will be adopting external-dns as the means in which we will manage DNS internally within the DNS Operator. This will become the base off which we implement multi cluster dns support and adopt new providers.

**How**

Add external-dns (latest released version v0.14.0) as a go module dependency and update the dnsrecord controller and dns provider logic to align with an external-dns approach to dns management.

mikenairn:
Adds external-dns (latest released version v0.14.0) as a go module dependency and updates the dnsrecord controller and dns provider logic to align with an external-dns approach to dns management.

### Updates

**dnsrecord changes:**
* Replace `v1alpha1.Endpoint` with externaldns Endpoint
* Add `GetRootDomain` method to `DNSRecord`, returns the shortest domain that is shared across all `spec.Endpoints` dns names.

**provider changes:**

* Remove Ensure and Delete and instead include `externaldnsprovider.Provider` in `provider.Provider` interface.
* Add `provider.Config` struct to hold common provider configuration, currently it only handles filters. Update provider Factory constructors accept the config as input.
* Set DomainFilter, ZoneTypeFilter and ZoneIDFilter in provider implementations from `provider.Config`.

**AWS provider changes:**

* Implement `AdjustEndpoints` to change our generic provider specific fields for weight and geo into aws ones understood by external-dns.

**Google provider changes:**

* Copy external-dns google provider
* Adds implementations of funtions required to translate endpoints and resourceRecordSets into the different states.
  - endpointsToProviderFormat - converts a list of endpoints into a google specific format.
  - endpointsFromResourceRecordSets - converts a list of `ResourceRecordSet` into endpoints (google format).
  - resourceRecordSetFromEndpoint - converts an endpoint(google format) into a `ResourceRecordSet`.

**controller changes:**

* Update `DNSRecord` controller to use applyChanges which follows an external-dns process to determine a set of changes(Create/Update/Delete) that are required to be made against the provider.
* Use noop registry, no owner is currently set or checked for any records.
* Setup provider filters:
  - DomainFilter = root domain for endpoints determined by calling dnsRecord.GetRootDomain()
  - ZoneTypeFilter = "", public or private zones considered.
  - ZoneIDFilter = set to single hosted zone id (managedZone.Status.ID)
* Use sync policy, all changes will be applied (Create, Update and Deletes)
* Use plan with above and using the current zone(plan.Current) and spec endpoints(plan.Desired) as input.

**additional changes:**

downgrade gatewayapiv1 to v0.7.1. This is required since the external dns aws provider is including the external dns source package which has a dependency on Gateway API v0.7.1. Downgrading is less than ideal, but dns-operator shouldn't have a dependency on gateway api, and the health check code that is currently causing it will likely be removed soon anyway in favour of provider health checks. There is also an opportunity to change external-dns to prevent it including the source package since it's including here for no reason other than getting a version
https://github.com/kubernetes-sigs/external-dns/blob/master/provider/aws/session.go#L72

### Validation

As part of the POC several example dnsrecords were created to run through the various different scenarios on Google and AWS. These were all specific to my own accounts so are not added here. I did add them to the [poc branch](https://github.com/Kuadrant/dns-operator/tree/multicluster_external_dns/hack/dnsrecords) though so can be used, with modification, for testing here. 

A [test script](https://github.com/Kuadrant/dns-operator/blob/multicluster_external_dns/hack/test-multicluster.sh) is also available in this branch to step through each of the scenarios manually.

A [test suite](https://github.com/Kuadrant/dns-operator/issues/46) based on these will be created in the very near future, 

Note: The tests for "multicluster with no control plane" are not expected to work currently

closes #47


internal/controller/dnshealthcheckprobe_controller.go
```diff
@@ -15,7 +15,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/log"
-	gatewayapiv1 "sigs.k8s.io/gateway-api/apis/v1"
+	gatewayapiv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
```
mikenairn, line 18: See description for the reason why this is downgraded

api/v1alpha1/dnsrecord_types.go
```diff
@@ -182,6 +101,37 @@ const (
...
+	}
+
+	for idx := range dnsNames {
+		if !strings.HasSuffix(dnsNames[idx], domain) {
+			return "", fmt.Errorf("inconsitent domains, got %s, expected suffix %s", dnsNames[idx], domain)
+		}
+	}
+
+	return domain, nil
+}
```
mikenairn, line 133: We might want to replace this by adding a new `rootDomain` field to the dnsrecord spec, but this will work for now.

internal/provider/aws/aws.go
```diff
@@ -72,34 +76,64 @@ func NewProviderFromSecret(ctx context.Context, s *v1.Secret) (provider.Provider
...
+		}
 
-func (p *Route53DNSProvider) Delete(record *v1alpha1.DNSRecord, managedZone *v1alpha1.ManagedZone) error {
-	return p.change(record, managedZone, deleteAction)
+		if prop, ok := ep.GetProviderSpecificProperty(v1alpha1.ProviderSpecificGeoCode); ok {
+			ep.DeleteProviderSpecificProperty(v1alpha1.ProviderSpecificGeoCode)
+			if provider.IsISO3166Alpha2Code(prop) || prop == "*" {
+				ep.WithProviderSpecific(providerSpecificGeolocationCountryCode, prop)
+			} else {
+				ep.WithProviderSpecific(providerSpecificGeolocationCountryCode, prop)
```
mikenairn, line 128: @mikenairn This is wrong

api/v1alpha1/dnsrecord_types.go
```diff
@@ -182,6 +101,37 @@ const (
...
+// Validates that all endpoints share an equal root domain and returns an error if they don't.
+func (s *DNSRecord) GetRootDomain() (string, error) {
+	domain := ""
+	dnsNames := []string{}
+	for idx := range s.Spec.Endpoints {
+		dnsNames = append(dnsNames, s.Spec.Endpoints[idx].DNSName)
+	}
+	for idx := range dnsNames {
+		if domain == "" {
+			domain = dnsNames[0]
```
philbrookes, line 114: should this be `domain = dnsNames[idx]`?

mikenairn: Yeah,  can probably just be:

```.go
for idx := range dnsNames {
	if domain == "" || len(domain) > len(dnsNames[idx]) {
		domain = dnsNames[idx]
	}
}
```

Will update.

api/v1alpha1/dnsrecord_types.go
```diff
@@ -182,6 +101,33 @@ const (
 	DefaultGeo string = "default"
 )
 
+// GetRootDomain returns the shortest domain that is shared across all spec.Endpoints dns names.
+// Validates that all endpoints share an equal root domain and returns an error if they don't.
+func (s *DNSRecord) GetRootDomain() (string, error) {
```
philbrookes, line 106: doesn't this just get us the longest shared hostname?

e.g. with hosts like:
```
app1.pb.domain.com.
app2.pb.domain.com.
app3.pb.domain.com.
```
We'd get not value? Where we'd expect `pb.domain.com` right?

Doesn't it make more sense to work up the first hostname sections until we find the longest section that every host has? I.e. in the example above, they all have `com`, `domain.com`, `pb.domain.com` with that being the longest value that every host has, return that?

mikenairn: Yeah, likely not going to work if hosts are of identical length, will rework. That said i don't want to spend too much time on it as i think we will just replace it with a spec update tbh, just don't want to do that here.

mikenairn: Hmm actually, the current logic is correct, or at least correct with what i was thinking. 

We are trying to find a single entry that is the start point, not work out the longest shared hostname. We are assuming that all the records in this record are related and that they all share a dns name that is also defined in this record set.

In your example we would expect there to be an entry for `pb.domain.com`, or you create 3 dns records.

I suppose it depends what we are trying to support with our DNSRecord, i was looking at it from the POV that a  kuadrant DNSRecord is defining a record set from beginning to end and if you have multiple endpoints one of them *must* be the start of the chain. :thinking:

mikenairn:
> Could you put the samples you showed us in the demo somewhere, so I can try this out locally?

Added a "Validation" section to the end of the description. Has links to the example DNSRecords i was using.

api/v1alpha1/dnsrecord_types.go
philbrookes: hmm yeah, that's true, so I suppose this functions for a DNS Record created by Kuadrant, but won't for like a weird manual DNS Record... but then in that scenario there might not even be a "root" it could just be a set of unrelated A records, right?

This is making me wonder if the "Root Host" is actually a Kuadrant concept, and not a dns-operator concept, and so doesn't belong here?

mikenairn: We have a follow up issue now to add a "rootHost" to the DNSRecord spec the https://github.com/Kuadrant/dns-operator/issues/50. This will allow the enforcement of the desired behaviour here, but also allow a DNSRecord to be used for other purposes that do not require a shared "rootHost".

philbrookes (approved):
/lgtm

PR #54, opened 2024-03-11
remove health checker

### LINKED ISSUE #25: DNS Operator uses excessive CPU
When deploying and measuring CPU usage, we see the DNSOperator using 200m as a baseline. This is quite a bit more than other controllers. 
Investigate and debug what maybe causing the dns operator to consume higher levels of CPU

### LINKED ISSUE #45: Remove provider agnostic health check monitor
**What**

Remove all the code only required for provider agnostic heath checks. We will not be supporting these as they are currently implemented and are falling back to provider health checks (AWS only) for the initial release.

https://github.com/Kuadrant/dns-operator/blob/main/cmd/main.go#L87-L88

This is also currently causing the high CPU usage so should also close https://github.com/Kuadrant/dns-operator/issues/25

maleck13:
closes #45
closes #25


PROJECT
```diff
@@ -27,12 +27,3 @@ resources:
...
   version: v1alpha1
-- api:
-    crdVersion: v1
-    namespaced: true
-  controller: true
-  domain: kuadrant.io
-  kind: DNSHealthCheckProbe
-  path: github.com/kuadrant/dns-operator/api/v1alpha1
-  version: v1alpha1
-version: "3"
```
mikenairn, line 38: Add this back

PROJECT
```diff
@@ -27,12 +27,4 @@ resources:
...
+version: "3"
```
mikenairn, line 30: You are removing the newline here

mikenairn:
@maleck13 Can you run `go mod tidy`, it should remove the gateway-api direct dependency.

mikenairn:
@maleck13 You also need to remove https://github.com/Kuadrant/dns-operator/blob/main/config/manifests/bases/dns-operator.clusterserviceversion.yaml#L18-L23. For some reason `operator-sdk generate kustomize manifests -q` doesn't remove it, but if you empty the array and re-run it will be as expected :man_shrugging: .

mikenairn (changes requested):
Remove https://github.com/Kuadrant/dns-operator/blob/main/config/manifests/bases/dns-operator.clusterserviceversion.yaml#L18-L23. For some reason `operator-sdk generate kustomize manifests -q` doesn't remove it, but if you empty the array and re-run it will be as expected 🤷‍♂️ .

Run `go mod tidy`

mikenairn (approved):

PR #56, opened 2024-03-11
registry: Use txt registry for owned records

### LINKED ISSUE #48: Update default registry to the TXT Registry
**What**

Update the external dns registry that we use from the [NooP](https://github.com/Kuadrant/dns-operator/pull/43/files#diff-adacb6a4589c65c2d853c0f59ca02d6b8fb50bb388b41de1d13e99b0d128cfe1R244) Registry to the [TXT](https://github.com/Kuadrant/dns-operator/blob/multicluster_external_dns/internal/controller/dnsrecord_controller.go#L263) registry. This will enforce this regardless of single vs multicluster, it will always be used.

Future work might allow this to be configurable, but for now we will always use the TXT registry.

**How**

Update the dnsrecord controller as described above. The TXT registry in the current release for some reason is creating two sets of TXT records, old and new, we don't want this as it's only increasing the amount of records we are creating for no good reason. Figure out if it can be configured not to (i couldn't see an obvious way), if not we will have to fork (we will have to later anyway) and remove the old TXT creation logic.

A registry in ExternalDNS requires an owner id to be set in order to function. This owner id is normally set explicitly on a controller instance and is used by the configured registry for all records created by that ExternalDNS instance. In our case a DNSRecord is essentially supplying all configuration needed on a per record set(Endpoint array) basis, as such in order to use a registry we need an `ownerID` configured somewhere(DNSRecord.spec?) to use with it.

**Option 1**

Add a new optional field `ownerID` to the DNSRecord spec.

The use of a registry is optional based on the existence of an `ownerID` in the DNSRecord.Spec. If no `ownerID` is specified the Noop Registry is used(Current behaviour), if one is set we use a TXT registry by default with the owner id from the `DNSRecord.Spec.OwnerID` field.

Consideration needs to be given to what happens if that ID changes. We possibly shouldn't allow it to be once you create it.

**Option 2**

Generate the `ownerID` in the DNS operator based on the same logic planned [here](https://github.com/Kuadrant/kuadrant-operator/issues/452) and always use the TXT Registry.

Note: It's not actually essential that the same ID is used in cluster specific endpoint generation in the kuadrant operator as the ID used to represent an owner in the DNS operator, but it probably is preferable. 

**Requires:**

* https://github.com/Kuadrant/dns-operator/pull/43

mikenairn:
Use the TXT Registry when an `ownerID` is specified for a record being processed.  Adds an optional field `ownerID` to the DNSRecord spec. This is immutable since changing it after initial creation will likely not result in desirable behaviour.

Requires:
* https://github.com/Kuadrant/external-dns/pull/1

closes #48


internal/controller/dnsrecord_controller.go
```diff
@@ -81,6 +80,9 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		logger.Info("Removing Finalizer", "name", DNSRecordFinalizer)
 		controllerutil.RemoveFinalizer(dnsRecord, DNSRecordFinalizer)
 		if err = r.Update(ctx, dnsRecord); client.IgnoreNotFound(err) != nil {
+			if apierrors.IsConflict(err) {
```
philbrookes, line 83: Do you think it's worth creating a generalised handler for this, in the DNS Operator, to help us reduce the number of calls to the API? Something like:
```
func PersistentUpdate(ctx, client, object, tries) error {

    err := client.Update(ctx, object)

    while err != nil && count < tries {
        count++

        // get fresh copy of object and generate patch

        err := client.Patch(ctx, patch)
    }
}
```

philbrookes: Without some mechanism like that, every requeue could result in a few hits to the DNS Provider API, right?

mikenairn: Possibly, but a bit out of scope for these particular changes. The change here is just preventing a conflict being logged as an error (the logs get littered with conflict errors that really are not that important) and doesn't cause any more or less API calls. We do this in a few places, but was missing here.

We did talk about some mechanism like you describe to limit API calls in particular cases, conflicts being one of the main ones, but it's probably it's own task as part of the [dnsrecord lifecycle](https://github.com/Kuadrant/dns-operator/labels/dnsrecord%20lifecycle) issues no?.

philbrookes (approved):
lgtm

PR #59, opened 2024-03-15
add owner ID to DNS Record spec

philbrookes:


mikenairn:
@philbrookes You need to run `make bundle` and commit the changes

bundle.Dockerfile
```diff
@@ -6,7 +6,7 @@ LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
 LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=dns-operator
 LABEL operators.operatorframework.io.bundle.channels.v1=alpha
-LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.33.0
+LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.34.1
```
mikenairn, line 9: This shouldn't be changing We need to update that operator-sdk target to make sure it's always the same [version](https://github.com/Kuadrant/dns-operator/blob/main/Makefile#L51) being used. Surprised the CI passed with this tbh.

mikenairn (approved):
Looks fine. Obviously we would be looking for tests normally , but since I'm adding some in a PR coming soon and i think we can go with it.

bundle.Dockerfile
philbrookes: I've updated this to stay on 1.33.0, so this can be merged. I'll take a look into why this bled over from the kuadrant-operator separately.

mikenairn (approved):

Issue #36, opened 2024-03-01
Re-queue validation intermittently


philbrookes:
To ensure consistency across all of our clusters, we will need to occasionally queue a verification check to ensure the zone looks as expected for the local cluster.

To do this, when a verification reconcile finishes successfully (thereby setting the count of failed verifications in the status to 0), it should be requeued for 15 minutes later to re-verify and update the DNS Record status if required.

Update status with the next queued time and the time it was queued and valid length.

Requeue time should have a sane default (e.g. 15 minutes) and be possible to override as a flag to the runtime.

When observedGeneration == generation and the time it was queued is younger than valid length old - do not go to DNS Provider API actions, only verify metadata, e.g. finalizers.

- [ ] Delay added with default value in main.go
- [ ] Status updated with next queued reconcile
- [ ] Flag to override this value to a custom value
- [ ] Tests added / updated

Issue #38, opened 2024-03-01
Schedule removal of finalizer from DNS Records


philbrookes:
When a DNS Record is being removed, the following must be successfully completed before the finalizer can be removed:
- Remove the local clusters records and targets from the relevant zone
- Requeue 
- Ensure no changes need to be planned

Only once these actions have all resolved should the finalizer be removed, and the DNS Record allowed to be deleted.

PR #66, opened 2024-03-27
 GH-36 dns record lifecycle

makslion:
After applying changes to the DNS provider ruqueue DNS Record for validation; mark the Record as ready only after the validation success. 
Also, reconcile records with a default timeout to ensure the validity of the content.

### DIFF
--- a/internal/controller/dnsrecord_controller.go
+++ b/internal/controller/dnsrecord_controller.go
@@ -18,7 +18,6 @@ package controller
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"strings"
 	"time"
@@ -49,6 +48,7 @@ const (
 var (
 	defaultRequeueTime    time.Duration
 	validationRequeueTime = time.Second * 5
+	noRequeueDuration     = time.Duration(0)
 	validFor              time.Duration
 	reconcileStart                    = metav1.Time{}
 	Clock                 clock.Clock = clock.RealClock{}
@@ -113,7 +113,7 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		reason = "ValidationError"
 		message = fmt.Sprintf("validation of DNSRecord failed: %v", err)
 		setDNSRecordCondition(dnsRecord, string(v1alpha1.ConditionTypeReady), metav1.ConditionFalse, reason, message)
-		return r.updateStatus(ctx, previous, dnsRecord, "0s")
+		return r.updateStatus(ctx, previous, dnsRecord, noRequeueDuration)
 	}
 
 	// Publish the record
@@ -122,7 +122,7 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		reason = "ProviderError"
 		message = fmt.Sprintf("The DNS provider failed to ensure the record: %v", provider.SanitizeError(err))
 		setDNSRecordCondition(dnsRecord, string(v1alpha1.ConditionTypeReady), metav1.ConditionFalse, reason, message)
-		return r.updateStatus(ctx, previous, dnsRecord, "0s")
+		return r.updateStatus(ctx, previous, dnsRecord, noRequeueDuration)
 	}
 	// success
 	dnsRecord.Status.ObservedGeneration = dnsRecord.Generation
@@ -130,12 +130,8 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	return r.updateStatus(ctx, previous, dnsRecord, requeueAfter)
 }
 
-func (r *DNSRecordReconciler) updateStatus(ctx context.Context, previous, current *v1alpha1.DNSRecord, requeueAfter string) (reconcile.Result, error) {
-	requeueDuration, err := time.ParseDuration(requeueAfter)
-	if err != nil {
-		return ctrl.Result{}, errors.New("error parsing duration while setting requeue time")
-	}
-	current.Status.QueuedFor = metav1.NewTime(reconcileStart.Add(requeueDuration))
+func (r *DNSRecordReconciler) updateStatus(ctx context.Context, previous, current *v1alpha1.DNSRecord, requeueAfter time.Duration) (reconcile.Result, error) {
+	current.Status.QueuedFor = metav1.NewTime(reconcileStart.Add(requeueAfter))
 
 	if !equality.Semantic.DeepEqual(previous.Status, current.Status) {
 		updateError := r.Status().Update(ctx, current)
@@ -144,7 +140,7 @@ func (r *DNSRecordReconciler) updateStatus(ctx context.Context, previous, curren
 		}
 		return ctrl.Result{}, updateError
 	}
-	return ctrl.Result{RequeueAfter: requeueDuration}, nil
+	return ctrl.Result{RequeueAfter: requeueAfter}, nil
 }
 
 // SetupWithManager sets up the controller with the Manager.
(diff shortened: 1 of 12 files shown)


internal/controller/dnsrecord_controller.go
```diff
@@ -73,7 +81,7 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	dnsRecord := previous.DeepCopy()
 
 	if dnsRecord.DeletionTimestamp != nil && !dnsRecord.DeletionTimestamp.IsZero() {
-		if err := r.deleteRecord(ctx, dnsRecord); err != nil {
+		if err = r.deleteRecord(ctx, dnsRecord); err != nil {
 			logger.Error(err, "Failed to delete DNSRecord")
 			return ctrl.Result{}, err
 		}
```
maleck13, line 87: do we not also need to requeue for validation here to ensure the record is still gone before removing the finalizer?

makslion: yes we do and it will be part of #38 (and the WIP PR is here #72 - I'm getting it ready now)

maleck13: ok `deleteRecord` calls applyChanges. But we need to requeue after a delete is done

makslion: the apply changes will return a timeout and in #72  I'm checking if that is a "validation" timeout I'm requeueing (does this word even exist? :laughing: ) until `applyChanges` won't say that there is nothing to be deleted

internal/controller/dnsrecord_controller.go
```diff
@@ -89,56 +97,61 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
-	message = "Provider ensured the dns record"
 	err = dnsRecord.Validate()
 	if err != nil {
-		status = metav1.ConditionFalse
 		reason = "ValidationError"
 		message = fmt.Sprintf("validation of DNSRecord failed: %v", err)
-		setDNSRecordCondition(dnsRecord, string(conditions.ConditionTypeReady), status, reason, message)
-		return r.updateStatus(ctx, previous, dnsRecord)
+		setDNSRecordCondition(dnsRecord, string(v1alpha1.ConditionTypeReady), metav1.ConditionFalse, reason, message)
+		return r.updateStatus(ctx, previous, dnsRecord, "0s")
```
maleck13, line 116: not a big fan of this being a string an being parsed later on. If we have to do it this way, perhaps we could also pull it out to a constant? `requeueImmediate`  or something?

internal/controller/dnsrecord_controller.go
```diff
@@ -89,56 +97,61 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
 	}
+
 	// Publish the record
-	err = r.publishRecord(ctx, dnsRecord)
+	requeueAfter, err := r.publishRecord(ctx, dnsRecord)
```
maleck13, line 120: ok so this makes sense 👍 so publish->applyChanges->returns requeue if changes applied

internal/controller/dnsrecord_controller.go
makslion: it seemed like that as well for me. [Here](https://github.com/Kuadrant/dns-operator/pull/72/commits/4f4f387d7c369fb04baca3c2ce8e097a88c3fc4d#diff-adacb6a4589c65c2d853c0f59ca02d6b8fb50bb388b41de1d13e99b0d128cfe1R51) is the way it looks good to me. Could have those changes sin this PR if you prefer it better. But all of them should land soon so i decided to keep this one as is :thinking:

internal/controller/dnsrecord_controller.go
makslion: yes, `publish` would make a decision on calling the `applyChanges` or going into the "not yet!" loop (got the record before we queued it). The `applyChanges` would either do changes and return validation requeue time or will report that no changes are needed and return a long timeout.

internal/controller/dnsrecord_controller.go
makslion: and along with it there would be only one `parse` call and it is only being set from `time.Duration.String()` string so we could be sure it will succeed

internal/controller/dnsrecord_controller.go
```diff
@@ -166,7 +179,7 @@ func (r *DNSRecordReconciler) deleteRecord(ctx context.Context, dnsRecord *v1alp
 		return fmt.Errorf("the managed zone is not in a ready state : %s", managedZone.Name)
 	}
 
-	err = r.applyChanges(ctx, dnsRecord, managedZone, true)
+	_, err = r.applyChanges(ctx, dnsRecord, managedZone, true)
```
maleck13, line 182: ok so we will handle a requeue here in the next pr

internal/controller/dnsrecord_controller.go
```diff
@@ -194,26 +207,39 @@ func (r *DNSRecordReconciler) publishRecord(ctx context.Context, dnsRecord *v1al
 	}
 	err := r.Get(ctx, client.ObjectKeyFromObject(managedZone), managedZone, &client.GetOptions{})
 	if err != nil {
-		return err
+		return "0s", err
```
maleck13, line 210: so kinda wondering again why we went for a string here?

internal/controller/dnsrecord_controller.go
```diff
@@ -194,26 +207,39 @@ func (r *DNSRecordReconciler) publishRecord(ctx context.Context, dnsRecord *v1al
...
+		return "0s", fmt.Errorf("the managed zone is not in a ready state : %s", managedZone.Name)
 	}
 
-	if dnsRecord.Generation == dnsRecord.Status.ObservedGeneration {
-		logger.V(3).Info("Skipping managed zone to which the DNS dnsRecord is already published", "dnsRecord", dnsRecord.Name, "managedZone", managedZone.Name)
-		return nil
+	// cut off here for the short reconcile loop
+	requeueIn := validFor
+	if dnsRecord.Status.ValidFor != "" {
+		requeueIn, _ = time.ParseDuration(dnsRecord.Status.ValidFor)
```
maleck13, line 221: shoud we be ignoring the error here? Is it possible we have a bad value?

maleck13 (commented):
Changes look pretty good. Main question is around use of strings for time durations.

internal/controller/dnsrecord_controller.go
maleck13: Ok in that case I am ok to approve with the improvements that are coming

maleck13 (approved):

makslion:
Added log on wrote counter bump and now there is only one parseDuration call (string that is being parsed from the CR and set only from the `time.Duration.String()` calls

internal/controller/dnsrecord_controller.go
makslion: updated

internal/controller/dnsrecord_controller.go
makslion: possible only if someone goes into the CR and changes values manually. This value is being set by `.String()` call on the duration struct. In case someone messes up with it will be `0` Duration. Then the logic later on will look like this:
```
expiryTime := previousReconcileStart.Add(0)
       // new reconcile will always be After expiryTime, so this clause always ends up as false 
	if !generationChanged(dnsRecord) && reconcileStart.Before(&expiryTime) {
               // will not requeue 
		return 0, nil
	}
```
and we won't use the `requeueIn` anywhere else so the zero duration doesn't matter. 

In other words, we really don't care about an error there - user would need to manually play with it and of he does and sets an invalid value we will end up ignoring it

internal/controller/dnsrecord_controller.go
makslion: well, there is no harm in having them here I guess.

internal/controller/dnsrecord_controller.go
```diff
@@ -302,15 +324,27 @@ func (r *DNSRecordReconciler) applyChanges(ctx context.Context, dnsRecord *v1alp
...
+			return validationRequeueTime, err
 		}
 	} else {
 		logger.Info("All records are already up to date")
+		dnsRecord.Status.WriteCounter = 0
+		setDNSRecordCondition(dnsRecord, string(v1alpha1.ConditionTypeReady), metav1.ConditionTrue, "ProviderSuccess", "Provider ensured the dns record")
 	}
 
-	return nil
+	return defaultRequeueTime, nil
```
mikenairn, line 349: nit: To me the logic around updating the conditions and deciding what this requeue time value should be is not the concern of `applyChanges`. This should just return an error and a bool to say if it had changes, and wherever this is being called use that info to calculate the status and requeue validation time.

It also looks like you are setting stuff in the status before you even know if it successfully updated or not.

PR #72, opened 2024-04-10
GH-38 validatie record on deletion

makslion:
Before removing the finalizer ensure that the DNS provider removed records

### DIFF
--- a/internal/controller/dnsrecord_controller.go
+++ b/internal/controller/dnsrecord_controller.go
@@ -81,9 +81,16 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	dnsRecord := previous.DeepCopy()
 
 	if dnsRecord.DeletionTimestamp != nil && !dnsRecord.DeletionTimestamp.IsZero() {
-		if err = r.deleteRecord(ctx, dnsRecord); err != nil {
+		requeueTime, err := r.deleteRecord(ctx, dnsRecord)
+		if err != nil {
 			logger.Error(err, "Failed to delete DNSRecord")
-			return ctrl.Result{}, err
+			return ctrl.Result{RequeueAfter: requeueTime}, err
+		}
+		// if requeueTime returned  is the same as validationRequeueTime - the deleteRecord has successfully applied changes
+		// in this case we need to queue for validation to ensure DNS Provider retained changes
+		// before removing finalizer and deleting the DNS Record CR
+		if requeueTime == validationRequeueTime {
+			return ctrl.Result{RequeueAfter: requeueTime}, nil
 		}
 		logger.Info("Removing Finalizer", "name", DNSRecordFinalizer)
 		controllerutil.RemoveFinalizer(dnsRecord, DNSRecordFinalizer)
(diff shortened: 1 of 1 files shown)


internal/controller/dnsrecord_controller.go
```diff
@@ -81,9 +81,14 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
 	if dnsRecord.DeletionTimestamp != nil && !dnsRecord.DeletionTimestamp.IsZero() {
-		if err = r.deleteRecord(ctx, dnsRecord); err != nil {
+		requeueTime, err := r.deleteRecord(ctx, dnsRecord)
+		if err != nil {
 			logger.Error(err, "Failed to delete DNSRecord")
-			return ctrl.Result{}, err
+			return ctrl.Result{RequeueAfter: requeueTime}, err
+		}
+		// requeue for validation
+		if requeueTime == validationRequeueTime {
```
maleck13, line 90: Ok so delete was successful but we need to revalidate. Hmm this took me a while to figure out. Could we add a comment explaining that is what is happening here and why we do this comparison?

maleck13 (approved):

internal/controller/dnsrecord_controller.go
```diff
@@ -81,9 +81,16 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
 
 	if dnsRecord.DeletionTimestamp != nil && !dnsRecord.DeletionTimestamp.IsZero() {
-		if err = r.deleteRecord(ctx, dnsRecord); err != nil {
+		requeueTime, err := r.deleteRecord(ctx, dnsRecord)
+		if err != nil {
 			logger.Error(err, "Failed to delete DNSRecord")
-			return ctrl.Result{}, err
+			return ctrl.Result{RequeueAfter: requeueTime}, err
+		}
+		// if ve got validationRequeue time - the deleteRecord has applied changes
```
maleck13, line 89: t
```suggestion
		// if requeueTime returned  is the same as validationRequeueTime - the deleteRecord has successfully applied changes
```

makslion: :facepalm:

internal/controller/dnsrecord_controller.go
```diff
@@ -167,28 +174,28 @@ func (r *DNSRecordReconciler) deleteRecord(ctx context.Context, dnsRecord *v1alp
...
 			logger.Info("Record not found in managed zone, continuing", "dnsRecord", dnsRecord.Name, "managedZone", managedZone.Name)
-			return nil
+			return noRequeueDuration, nil
 		} else if strings.Contains(err.Error(), "no endpoints") {
 			logger.Info("DNS record had no endpoint, continuing", "dnsRecord", dnsRecord.Name, "managedZone", managedZone.Name)
-			return nil
+			return noRequeueDuration, nil
 		}
-		return err
+		return noRequeueDuration, err
```
maleck13, line 194: does `noRequeueDuration` have the effect as returning `ctrl.Result{}, updateError`?

makslion: yes. The way it is interpreted: if `err != nil` it will ignore `ctrl.Result{}` entirely and will requeue immediately. If `err == nil` then it will look at instructions for requeue. 
This makes me think that I might made it more confusing than it should be. There was a suggestion to just roam with the boolean flag saying if there were changes published or not and we decide on how to requeue in the very last step. I'm planning on having that implemented after the code freeze (but if that will make more sense could have it ready by the end of a day)

maleck13: no leave as is for now and can follow up with improvements

maleck13 (approved):

Issue #35, opened 2024-03-01
Re-queue DNS Record whenever a write to the Cloud Provider occurs


philbrookes:
Whenever the DNS Operator calculates that a change to the zone is required (update, create or delete records) the DNS Operator should also schedule a follow-up reconcile a few thousand milliseconds later (including a random shudder (+/-10% of delay time) ) to ensure that the changes persisted.

If the changes are not persisted then it should be re-written and another reschedule should occur.

Update status with the next queued time and the time it was queued.

Requeue time should have a sane default (e.g. 4000ms) and be possible to override as a flag to the runtime.

- [ ] Delay added with default value in main.go
- [ ] Status updated with next queued reconcile
- [ ] Flag to override this value to a custom value
- [ ] Randomness added on each requeue
- [ ] Tests added / updated

PR #73, opened 2024-04-10
GH-35 randomize validation time

makslion:
Randomize validation reconcilation delays

### DIFF
--- a/internal/common/helper_test.go
+++ b/internal/common/helper_test.go
@@ -1,0 +1,41 @@
+package common
+
+import (
+	"testing"
+	"time"
+)
+
+func TestRandomizeDuration(t *testing.T) {
+	testIterations := 100
+
+	tests := []struct {
+		name     string
+		variance float64
+		duration time.Duration
+	}{
+		{
+			name:     "returns valid duration in range",
+			variance: 0.5,
+			duration: time.Second * 5,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			i := 0
+			for i < testIterations {
+				if got := RandomizeDuration(tt.variance, tt.duration); !isValidVariance(tt.duration, got, tt.variance) {
+					t.Errorf("RandomizeDuration() invalid randomization; got = %v", got.String())
+				}
+				i++
+			}
+		})
+	}
+}
+
+func isValidVariance(duration, randomizedDuration time.Duration, variance float64) bool {
+	upperLimit := float64(duration.Milliseconds()) + float64(duration.Milliseconds())*variance
+	lowerLimmit := float64(duration.Milliseconds()) - float64(duration.Milliseconds())*variance
+
+	return float64(randomizedDuration.Milliseconds()) > lowerLimmit &&
+		float64(randomizedDuration.Milliseconds()) < upperLimit
+}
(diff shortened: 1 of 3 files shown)


internal/controller/dnsrecord_controller.go
```diff
@@ -69,6 +70,13 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	logger := log.FromContext(ctx)
 
 	reconcileStart = metav1.Now()
+
+	// randomize validation reconcile delay
+	seconds := float64(validationRequeueTime.Milliseconds())
```
maleck13, line 75: nit: is this var name a bit misleading? should it be milliseconds?

internal/controller/dnsrecord_controller.go
```diff
@@ -69,6 +70,13 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+	validationRequeueTime = time.Millisecond * time.Duration(rand.Int63nRange(
+		int64(seconds*0.9),
+		int64(seconds*1.1)))
```
maleck13, line 78: ok so the interval will be between 90 and 110 % of the `validationRequeueTime`? wondering if we should increase the possible range a bit. Like 50-150% . Anyway just a thought

maleck13: possibly worth isolating into a function and adding a simple unit test?

maleck13 (approved):
looks ok to merge. just a couple small things to possibly improve

internal/controller/dnsrecord_controller.go
makslion: renamed

maleck13 (approved):

internal/controller/dnsrecord_controller.go
makslion: increased variance to 50% and isolated into a separate function. 

Not sure though how to reliably test the randomization function: we could have a test just ensuring that we get a valid value from this (but given how the code works we always will have a valid `time.Duration` value out of this) and we already testing for "usable" duration in the integration suite (if it to randomize seconds into minutes we will timeout out of the integration suite). The suggested way to test randomness is through a lot of iterations to build up a statistically significant amount of data - that doesn't sound like something worth doing. 
In any case - I put together a tiny test that should be what you had in mind 

Went through the package and there are a few functions that could use unit tests as well. If we do this - maybe better to move our integration suite into a separate package and only keep `filename_test.go` for the unit suite? I can create a issue for "better days" to reorganize tests in the DNS Operator if that sounds cool :thinking:

maleck13: Yeah not huge value here, I guess we could do a number of iterations and ensure the value is within the expected range?

internal/common/helper_test.go
```diff
@@ -0,0 +1,36 @@
...
+
+func TestRandomizeDuration(t *testing.T) {
+
+	tests := []struct {
+		name     string
+		variance float64
+		duration time.Duration
+	}{
+		{
+			name:     "returns valid duration in range",
```
maleck13, line 16: yep this is fine perhaps forcing it to do a loop and  a set number of iterations testing each is within the range?

maleck13 (approved):

internal/common/helper_test.go
makslion: :facepalm: yep, that is smart. Done

Issue #34, opened 2024-03-01
Record write attempts in status for current generation


philbrookes:
**What**

Whenever a write to the Cloud Provider occurs, we need to emit a metric noting the occurrence which includes the generation.

Add a condition on the status whenever we write to Cloud Provider, e.g.:
```
writeCount: <N>
firstWrite: <timestamp>
Reason: <human message>
```

Reset the count if either:
- Generation of DNS Record is different to observedGeneration
- Plan is empty

We will need to set an upper limit to how many sequential writes are permitted before the controller stops trying so frequently and exposes an error in the status.

When a status with writeCount zero is present, the DNS Record should also have a accepted status of true.

philbrookes:
@david-martin do you have any thoughts on the metric we are considering emitting here? Is this already going to be covered by kube state metrics as we are writing the details to the status.

david-martin:
> Is this already going to be covered by kube state metrics as we are writing the details to the status.

The state metrics config for DNSPolicy is https://github.com/Kuadrant/gateway-api-state-metrics/blob/f37b923fb6789d875992b71a1a88be3899995728/config/default/custom-resource-state.yaml#L514-L566

It can be extended to have different state information, however it might be better to have the metrics you think are valuable be emitted from the controller itself rather than depend on a separate library.
That would allow for easier consumption, alerting and visualisation of metrics.

The write count sounds like a good candidate for a counter metric, with some label for the DNS record name and namespace.
The firstWrite sounds like a good candidate for a gauge that's set to the timestamp, again having the DNS record name and namespace labels.

I would avoid having the reason message be in any metric name or label as it sounds like it could vary a lot.

PR #74, opened 2024-04-12
GH-34 emmit write counter as a metric

makslion:
Emmiting write counter as a metric and setting an upper limit for the counter after witch we give up

### DIFF
--- a/internal/controller/dnsrecord_controller.go
+++ b/internal/controller/dnsrecord_controller.go
@@ -18,6 +18,7 @@ package controller
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 	"time"
@@ -44,6 +45,7 @@ import (
 
 const (
 	DNSRecordFinalizer        = "kuadrant.io/dns-record"
+	WriteCounterLimit         = 20
 	validationRequeueVariance = 0.5
 )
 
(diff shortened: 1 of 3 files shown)


internal/controller/dnsrecord_controller.go
```diff
@@ -336,8 +339,16 @@ func (r *DNSRecordReconciler) applyChanges(ctx context.Context, dnsRecord *v1alp
 		// generation has not changed but there are changes.
 		// implies that they were overridden - bump write counter
 		if !generationChanged(dnsRecord) {
-			dnsRecord.Status.WriteCounter++
-			logger.V(3).Info("Changes needed on the same generation of record")
+			if dnsRecord.Status.WriteCounter < WriteCounterLimit {
```
maleck13, line 342: do we set any dimension on this metric? Seems like we should set the the dnsRecord name and namespace as labels?

maleck13: As I read it we will increment this for all dns records so only ever 20 writes or am I mistaken

maleck13 (changes requested):
I think the metric is not unique per dnsrecord?

makslion:
image
This is how it will behave now. That should be much better

internal/controller/dnsrecord_controller.go
makslion: yes, only 20 writes per generation of DNS record (frankly, if we have more then 2 writes per gen something is wrong with the DNS provider or, more accurately, something keeps overriding changes)

maleck13 (approved):

Issue #75, opened 2024-04-12
Record DNS Record lifecycle changes


makslion:
Write down changes related to the DNS Record lifecycle to ease future docs creation

PR #76, opened 2024-04-12
GH-75 dns record lifecycle notes

makslion:
A simple .md for people way better at docs to reference when documenting DNS Record lifecycle

### DIFF
--- a/docs/RFC/distributed-dns/dnsrecord-lifecycle.md
+++ b/docs/RFC/distributed-dns/dnsrecord-lifecycle.md
@@ -1,0 +1,54 @@
+# List of issues
+* [Re-queue validation intermittently GH-36](https://github.com/Kuadrant/dns-operator/issues/36)
+* [Re-queue DNS Record whenever a write to the Cloud Provider occurs GH-35](https://github.com/Kuadrant/dns-operator/issues/35)
+* [Schedule removal of finalizer from DNS Records GH-38](https://github.com/Kuadrant/dns-operator/issues/38)
+* [Record write attempts in status for current generation GH-34](https://github.com/Kuadrant/dns-operator/issues/34)
+
+# The idea
+We now will constantly reconcile DNS records. The reasoning is that other controllers may override/change records in the DNS provider so there is a need to requeue the DNS Record from time to time even when no local changes are introduced.
+
+
+# Details
+There are a few new fields on the DNS Record status:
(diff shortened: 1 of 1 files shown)


docs/RFC/distributed-dns/dnsrecord-lifecycle.md
```diff
@@ -0,0 +1,54 @@
...
+
+
+The `validationRequeueTime` duration is randomized +/- 50%.
+
+
+## When things go south
+If the record is received prematurely - the `ValidFor` + `QueuedAt` is more than the current time - we requeue it again for the `ValidFor` duration.
+
+
+When we encounter an error during the reconciliation we will not requeue the record and will put in an appropriate error message in the log and on the record. In order for it to reconcile again there must be a change to the DNS Record CR.
```
maleck13, line 45: This feels a bit final. Is that any error or an error from the dns provider?

maleck13 (approved):
looks good. Just wanted to clarify on the error handling

docs/RFC/distributed-dns/dnsrecord-lifecycle.md
makslion: Any error: from provider failures to MZ not being in a ready state. We treat all the errors in the same way. To be honest, in reality, there is one case - if it tries to update the record and fails to write the status to the CR - when the error would make it reconcile immediately (and that is exactly what we want). In any other case, the error will be put in the `Ready - false` condition and we would obey the `noRequeueDuration`. The idea is that there is something wrong in your configuration and you should first fix that and it will trigger reconciliation again. 
We have the power, however, to manipulate this behaviour through the returned duration right now: we could decide that for "managed zone not ready" we want to requeue the record in ~15 minutes but for an unknown policy type we might want to stop trying.

maleck13: Yeah I am just not sure we should always assume misconfiguration. There could be a network problem for example that means you can't reach the provider. So in my mind setting ready to false and requeuing but for `RequeueDuration` in these cases might not be a bad thing. IE allowing for some external state to have been fixed (perhaps there was something wrong with the zone or the credential for example) and then trying again. I realise we are documenting how it works right now which is fine, but I think we should capture whether we want it to behave that way permanently

makslion: I'll create an issue for this, or rather append it to #80

Issue #44, opened 2024-03-06
Add AWS Provider Health Checks


mikenairn:
**What**

Add implementation of health checks for AWS provider.

**How**

We had AWS provider health checks in MGC but we removed them when we thought we weren't going forward with provider health checks https://github.com/Kuadrant/multicluster-gateway-controller/commit/89e1893b8316ad265be036050cb923eaa83ad117. This can be used for reference.

Areas to consider in this work:

- New health check is required if an IP / CNAME in spec doesn't have a matching status health check
- When a new health check is required
  - List health checks in Provider
  - Add all relevant health checks to DNS Record status
  - Create any absent health checks
  - Delete any obsolete health checks

## Overview
- Health check defined in DNS Policy.
- Health check config is propagated to the DNS Operator via the DNS Record spec (in same format as DNS Policy).
- When values are present in DNS Record and health checks are supported in the provider: health checks are created in provider for all A record targets or external CNAME targets.
- Health checks created with name including the owner ID (if present)
- Store health check IDs in status of the DNS Record
  - Hostname 
  - IP / External CNAME
  - Health check ID
- Remove health checks for IPs / external CNAMEs in the status that are no longer in the endpoints.
  - If the health check delete fails because it's already gone, then remove it from the status.

PR #78, opened 2024-04-12
aws health checks

philbrookes:
First pass at AWS health checks in DNS Operator.

Some uses-cases not fully covered, and no smart backoff, or e2e tests yet.

Related PR for kuadrant-operator:
https://github.com/Kuadrant/kuadrant-operator/pull/543


internal/controller/dnsrecord_controller.go
```diff
@@ -123,17 +123,26 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+	if err := r.ReconcileHealthChecks(ctx, dnsRecord); err != nil {
+		return ctrl.Result{}, err
+	}
+
 	return r.updateStatus(ctx, previous, dnsRecord)
 }
 
 func (r *DNSRecordReconciler) updateStatus(ctx context.Context, previous, current *v1alpha1.DNSRecord) (reconcile.Result, error) {
 	if !equality.Semantic.DeepEqual(previous.Status, current.Status) {
+		log.Log.Info("status changed", "previous", previous.Status, "current", current.Status)
```
maleck13, line 136: might be a bit noisey?

internal/controller/dnsrecord_controller.go
```diff
@@ -123,17 +123,26 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
 		updateError := r.Status().Update(ctx, current)
 		if apierrors.IsConflict(updateError) {
+			log.Log.Info("requeueing due to conflict", "namespace/name", current.Namespace+"/"+current.Name)
```
maleck13, line 139: how valuable are these logs? Seems like they would also get quite noisy?

internal/controller/dnsrecord_controller.go
```diff
@@ -228,21 +237,46 @@ func setDNSRecordCondition(dnsRecord *v1alpha1.DNSRecord, conditionType string,
...
-	rootDomainFilter := externaldnsendpoint.NewDomainFilter([]string{filterDomain})
 
 	providerConfig := provider.Config{
 		DomainFilter:   externaldnsendpoint.NewDomainFilter([]string{managedZone.Spec.DomainName}),
 		ZoneTypeFilter: externaldnsprovider.NewZoneTypeFilter(""),
 		ZoneIDFilter:   externaldnsprovider.NewZoneIDFilter([]string{managedZone.Status.ID}),
 	}
-	logger.V(3).Info("applyChanges", "zone", managedZone.Spec.DomainName, "rootDomainFilter", rootDomainFilter, "providerConfig", providerConfig)
-	dnsProvider, err := r.ProviderFactory.ProviderFor(ctx, managedZone, providerConfig)
+	log.Log.V(3).Info("applyChanges", "zone", managedZone.Spec.DomainName, "providerConfig", providerConfig)
```
maleck13, line 257: this log looks inaccurate ?

internal/controller/dnsrecord_healthchecks.go
```diff
@@ -0,0 +1,242 @@
...
+	var err error
+
+	// Get the configuration for the health checks. If no configuration is
+	// set, ensure that the health checks are deleted
+	config := getHealthChecksConfig(dnsRecord)
+
+	log.Log.Info("DNS Record healthcheck status", "status", dnsRecord.Status.HealthCheck)
+	if config == nil {
+		log.Log.V(3).Info("No health check for DNSPolicy, ensuring deletion")
+		results, err = r.reconcileHealthCheckDeletion(ctx, dnsRecord)
```
maleck13, line 37: having a reconcile for deletion and a reconcile for adding / updating feels a bit odd? Why can't it just be done in the one reconcile?

internal/controller/dnsrecord_healthchecks.go
```diff
@@ -0,0 +1,242 @@
...
+	healthCheckReconciler := dnsProvider.HealthCheckReconciler()
+
+	var results []provider.HealthCheckResult
+	for _, dnsEndpoint := range dnsRecord.Spec.Endpoints {
+		log.Log.Info("reconciling DNS Record health checks", "endpoint", dnsEndpoint)
+		address, ok := provider.GetEndpointAddress(dnsEndpoint)
+		if !ok {
+			log.Log.Info("address not ok", "address", address)
+			continue
+		} else if !net.IsIPv4String(address) {
```
maleck13, line 68: intresting so will we not have any health checks if the using an AWS LB as it uses hostnames?

internal/controller/dnsrecord_controller.go
philbrookes: these were primarily for debugging, I will remove a bunch of them now.

internal/controller/dnsrecord_healthchecks.go
philbrookes: ah good point, I need to also capture when it's an external CNAME.

internal/controller/dnsrecord_healthchecks.go
philbrookes: this was brought in from the old MGC implementation, I can change it

maleck13: well doesn't block merging, but just seems a bit cumbersome

internal/controller/dnsrecord_healthchecks.go
```diff
@@ -0,0 +1,243 @@
...
+func (r *DNSRecordReconciler) getProbeStatus(address string, dnsRecord *v1alpha1.DNSRecord) *v1alpha1.HealthCheckStatusProbe {
+	if dnsRecord.Status.HealthCheck == nil || dnsRecord.Status.HealthCheck.Probes == nil {
+		log.Log.Info("status or probes are nil", "status", dnsRecord.Status.HealthCheck)
+		return nil
+	}
+	for _, probeStatus := range dnsRecord.Status.HealthCheck.Probes {
+		if probeStatus.IPAddress == address {
+			return &probeStatus
+		}
+		log.Log.Info("not a match", "ip", probeStatus.IPAddress, "address", address)
```
maleck13, line 114: also noisy?

internal/controller/dnsrecord_healthchecks.go
```diff
@@ -0,0 +1,243 @@
...
+	return nil
+}
+
+func (r *DNSRecordReconciler) reconcileHealthCheckStatus(results []provider.HealthCheckResult, dnsRecord *v1alpha1.DNSRecord) error {
+	probesCondition := metav1.Condition{
+		ObservedGeneration: dnsRecord.Generation,
+		Status:             metav1.ConditionTrue,
+		Reason:             "AllProbesSynced",
+		LastTransitionTime: metav1.Now(),
+		Message:            "all probes synced successfully",
```
maleck13, line 126: what if this was fmt.Sprintf("all %v probes synced successfully",len(results))

then lower down you prob don't need the if (len(results) > 0 ) ?

internal/provider/endpointUtils.go
```diff
@@ -0,0 +1,13 @@
...
+import (
+	externaldns "sigs.k8s.io/external-dns/endpoint"
+)
+
+func GetEndpointAddress(endpoint *externaldns.Endpoint) (string, bool) {
+	if len(endpoint.Targets) == 0 {
+		return "", false
+	}
+
+	return endpoint.Targets[0], true
```
maleck13, line 12: is there ever more than 1 target?

internal/controller/dnsrecord_healthchecks.go
```diff
@@ -0,0 +1,243 @@
...
+	log.Log.Info("deleting health checks")
+	dnsProvider, err := r.getDNSProvider(ctx, dnsRecord)
+	if err != nil {
+		log.Log.Info("error getting dns provider", "err", err)
+		return nil, err
+	}
+	reconciler := dnsProvider.HealthCheckReconciler()
+	var results []provider.HealthCheckResult
+	for _, endpoint := range dnsRecord.Spec.Endpoints {
+		address, _ := provider.GetEndpointAddress(endpoint)
```
maleck13, line 190: should we be ignoring the error here?

internal/provider/aws/health.go
```diff
@@ -0,0 +1,254 @@
...
+func NewRoute53HealthCheckReconciler(client route53iface.Route53API) *Route53HealthCheckReconciler {
+	return &Route53HealthCheckReconciler{
+		client: client,
+	}
+}
+
+func (r *Route53HealthCheckReconciler) Reconcile(ctx context.Context, spec provider.HealthCheckSpec, endpoint *externaldns.Endpoint, probeStatus *v1alpha1.HealthCheckStatusProbe) (provider.HealthCheckResult, error) {
+	healthCheck, exists, err := r.findHealthCheck(ctx, probeStatus)
+	if err != nil {
+		log.Log.Info("error finding health check", "err", err)
```
maleck13, line 46: should we log this here or just add it as context to the returned error? I imagine it will get logged here and then back at the dnsrecord reconciler?

internal/provider/aws/health.go
```diff
@@ -0,0 +1,254 @@
...
+		return provider.NewHealthCheckResult(status, *healthCheck.Id, address, *spec.Host, ""), nil
+	}
+
+	log.Log.Info("health check doesn't exist, creating")
+	healthCheck, err = r.createHealthCheck(ctx, spec, endpoint)
+	if err != nil {
+		return provider.HealthCheckResult{}, err
+	}
+
+	return provider.NewHealthCheckResult(provider.HealthCheckCreated, *healthCheck.Id, address, *spec.Host, fmt.Sprintf("Created health check with ID %s", *healthCheck.Id)), nil
```
maleck13, line 68: what is a `HealthCheckResult` in this context? Is the result of the update/create or is it whether the health check is failing / passing?

internal/provider/aws/health.go
```diff
@@ -0,0 +1,254 @@
...
+func (r *Route53HealthCheckReconciler) Reconcile(ctx context.Context, spec provider.HealthCheckSpec, endpoint *externaldns.Endpoint, probeStatus *v1alpha1.HealthCheckStatusProbe) (provider.HealthCheckResult, error) {
+	healthCheck, exists, err := r.findHealthCheck(ctx, probeStatus)
+	if err != nil {
+		log.Log.Info("error finding health check", "err", err)
+		return provider.HealthCheckResult{}, err
+	}
+
+	address, _ := provider.GetEndpointAddress(endpoint)
+
+	if exists {
```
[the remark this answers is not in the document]
maleck13: I guess the client might be the place to do that?

internal/provider/health.go
```diff
@@ -0,0 +1,129 @@
...
+	HealthCheckNoop    HealthCheckReconciliationResult = "Noop"
+	HealthCheckFailed  HealthCheckReconciliationResult = "Failed"
+)
+
+type HealthCheckProtocol string
+
+const HealthCheckProtocolHTTP HealthCheckProtocol = "HTTP"
+const HealthCheckProtocolHTTPS HealthCheckProtocol = "HTTPS"
+
+type FakeHealthCheckReconciler struct{}
```
maleck13, line 62: These fake and cached reconcilers, maybe would make sense in a different file

maleck13 (commented):
few questions and comments. I realise there is some pressure to get this merged. But lets see what tests we can get in over the next day or so WDYT?

internal/provider/aws/health.go
philbrookes: yeah, it's the result of the API call.

internal/controller/dnsrecord_controller.go
```diff
@@ -250,6 +255,27 @@ func setDNSRecordCondition(dnsRecord *v1alpha1.DNSRecord, conditionType string,
...
+	if err != nil {
+		return nil, err
+	}
+
+	providerConfig := provider.Config{
+		DomainFilter:   externaldnsendpoint.NewDomainFilter([]string{managedZone.Spec.DomainName}),
+		ZoneTypeFilter: externaldnsprovider.NewZoneTypeFilter(""),
+		ZoneIDFilter:   externaldnsprovider.NewZoneIDFilter([]string{managedZone.Status.ID}),
+	}
+	log.Log.V(3).Info("applyChanges", "zone", managedZone.Spec.DomainName, "providerConfig", providerConfig)
```
maleck13, line 275:
```suggestion
```

internal/controller/dnsrecord_healthchecks.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+
+	// Get the configuration for the health checks. If no configuration is
+	// set, ensure that the health checks are deleted
+	config := getHealthChecksConfig(dnsRecord)
+
+	if config == nil {
+		// deleting
+		reconciler := dnsProvider.HealthCheckReconciler()
+		var results []provider.HealthCheckResult
```
maleck13, line 40: looks like redefining the var here?

internal/controller/dnsrecord_healthchecks.go
```diff
@@ -0,0 +1,208 @@
...
+				result, err := reconciler.Delete(ctx, endpoint, probeStatus)
+				if err != nil {
+					return err
+				}
+
+				results = append(results, result)
+			}
+		}
+
+		if err != nil {
```
maleck13, line 57: whats this for?

internal/controller/dnsrecord_healthchecks.go
```diff
@@ -0,0 +1,208 @@
...
+				if err != nil {
+					return err
+				}
+				results = append(results, result)
+			}
+		}
+		if err != nil {
+			return err
+		}
+	}
```
maleck13, line 94: I think this whole code block need a bit of re-working

maleck13: So this is better but WDYT
```
for _, endpoint := range dnsRecord.Spec.Endpoints {
			addresses := provider.GetExternalAddresses(endpoint, dnsRecord)
			for _, address := range addresses {
				endpointId, err := idForEndpoint(dnsRecord, dnsEndpoint, address)
				if err != nil {
					return err
				}
				probeStatus := r.getProbeStatus(address, dnsRecord)
				// trigger delete
				if config == nil{
					log.Log.Info("deleting health checks for","endpoint ", endpoint)
					if probeStatus == nil {
						continue
					}
					result, err := healthCheckReconciler.Delete(ctx, endpoint, probeStatus)
					if err != nil {
						return err
					}

					results = append(results, result)
					continue
				}
				spec := provider.HealthCheckSpec{
					Id:               endpointId,
					Name:             fmt.Sprintf("%s-%s-%s", *dnsRecord.Spec.RootHost, dnsEndpoint.DNSName, address),
					Host:             dnsRecord.Spec.RootHost,
					Path:             config.Endpoint,
					Port:             config.Port,
					Protocol:         config.Protocol,
					FailureThreshold: config.FailureThreshold,
				}
				result, err := healthCheckReconciler.Reconcile(ctx, spec, dnsEndpoint, probeStatus, address)
				if err != nil {
					return err
				}
				results = append(results, result)
			}
		}
	

	result := r.reconcileHealthCheckStatus(results, dnsRecord)
	return result
```

maleck13: basically drop the repeated loop in the code.

internal/controller/dnsrecord_healthchecks.go
```diff
@@ -0,0 +1,191 @@
...
+	var allSynced = metav1.ConditionTrue
+
+	if dnsRecord.Status.HealthCheck == nil {
+		dnsRecord.Status.HealthCheck = &v1alpha1.HealthCheckStatus{
+			Conditions: []metav1.Condition{},
+			Probes:     []v1alpha1.HealthCheckStatusProbe{},
+		}
+	}
+
+	for _, c := range dnsRecord.Status.HealthCheck.Conditions {
```
maleck13, line 113: worth using ```meta.FindStatusCondition``` ?

internal/controller/dnsrecord_healthchecks.go
```diff
@@ -0,0 +1,191 @@
...
+
+	return &healthChecksConfig{
+		Endpoint:         dnsRecord.Spec.HealthCheck.Endpoint,
+		Port:             valueAs(toInt64, dnsRecord.Spec.HealthCheck.Port),
+		FailureThreshold: valueAs(toInt64, dnsRecord.Spec.HealthCheck.FailureThreshold),
+		Protocol:         (*provider.HealthCheckProtocol)(dnsRecord.Spec.HealthCheck.Protocol),
+	}
+}
+
+func valueAs[T, R any](f func(T) R, original *T) *R {
```
maleck13, line 171: seems a bit overkill ?

internal/controller/dnsrecord_healthchecks.go
philbrookes: oh cool, I didn't know this existed.

internal/controller/dnsrecord_healthchecks.go
philbrookes: I think this is just making it a little more legible and compact to convert stuff to pointers, does it need changing here?

maleck13 (commented):
There being not tests here is really not ideal. We need to make sure we get these in place @philbrookes can you  link to an issue 

@mikenairn In order to unblock moving kuadarant operator back to main of dns operator, I think we can make an exception for this knowing the focus will be on adding tests before we do the next release of DNS operator?

philbrookes:
> There being not tests here is really not ideal. We need to make sure we get these in place @philbrookes can you link to an issue
> 
> @mikenairn In order to unblock moving kuadarant operator back to main of dns operator, I think we can make an exception for this knowing the focus will be on adding tests before we do the next release of DNS operator?

My intention was to keep this issue open until the tests are added and the logic is firmed up more. https://github.com/Kuadrant/dns-operator/issues/44

maleck13 (approved):

PR #79, opened 2024-04-15
Gh49 distributed dns plan

### LINKED ISSUE #49: Implement multi cluster plan and registry
**What**

Adapt external-dns plan and registry APIs to allow multi cluster support.

[Plan](https://github.com/kubernetes-sigs/external-dns/blob/master/plan/plan.go) - Takes inputs (currently current and desired endpoints) and calculates a set of changes(Create/Update/Delete) that need to be made to get to the desired state.
[Registry](https://github.com/kubernetes-sigs/external-dns/blob/master/registry/txt.go) - Mechanism which allows setting ownership of records, we will be using the TXT registry initially. More info [here](https://github.com/kubernetes-sigs/external-dns/blob/master/docs/registry/txt.md).

Changes were made as part of a [POC](https://github.com/Kuadrant/external-dns/tree/multicluster_external_dns_poc) to both of these to prove some of this out, this can be used for reference/inspiration. Diff view [here](https://github.com/kubernetes-sigs/external-dns/compare/v0.14.0...Kuadrant:external-dns:multicluster_external_dns_poc?expand=1).
 
## How

related [ticket](https://github.com/Kuadrant/dns-operator/issues/40)

Use cases to ensure (Create test to ensure these are true). 
- [x] Cannot change record type for an existing record
- [x] Cannot update a record when no owner TXT record exists
- [x] Must only delete a record when current owner is the only owner
- [x] Must append targets and preserve existing targets when a matching record exists
- [x] Must be able to identify and update old invalid values
- [x] Must only remove targets whose record has no other owners

As the plan work is the core of the multi-cluster DNS solution, we need to ensure we document the code well so it is clearly understood how the plan works

- [ ] Update the RFC under the multi-cluster plan section to link to the plan code. https://github.com/Kuadrant/architecture/pull/70/files

Follow up tasks:

- [ ] https://github.com/Kuadrant/kuadrant-operator/issues/546
- [ ] https://github.com/Kuadrant/dns-operator/issues/85
- [ ] https://github.com/Kuadrant/kuadrant-operator/issues/545

mikenairn:
closes #49 #85 

Initial work to modify the external-dns plan to work in the proposed way for distubuted dns.

Updates the plan to treat all endpoints passed to it (current/desired) as part of a single shared set of records. The calculation logic now takes into account other endpoints added to the plan when calculating the desired changes for any other endpoint also added.

The plan calculation logic is broken into two steps:

1. Calculate what endpoints will be created/updated and deleted based on the current and desired records passed to the plan. For each endpoint that will still exist if these changes were to be applied(create/update), a map of dnsNames and their owners is calculated.

2. Create the external-dns Changes for the endpoints calculated in step one. For all endpoints calculated to be an update the target values are re-calculated using the information gathered in step one about ownership of dnsNames. If we detect that a CNAME has a "managed" target value (it's in our owner map) we use the information known about who owns it to manipulate the desired targets values.

### DIFF
--- a/internal/controller/dnsrecord_controller.go
+++ b/internal/controller/dnsrecord_controller.go
@@ -366,6 +366,10 @@ func (r *DNSRecordReconciler) applyChanges(ctx context.Context, dnsRecord *v1alp
 
 	plan = plan.Calculate()
 
+	if err = plan.ConflictError(); err != nil {
+		return noRequeueDuration, err
+	}
+
 	dnsRecord.Status.ValidFor = defaultRequeueTime.String()
 	dnsRecord.Status.QueuedAt = reconcileStart
 	if plan.Changes.HasChanges() {
@@ -381,7 +385,6 @@ func (r *DNSRecordReconciler) applyChanges(ctx context.Context, dnsRecord *v1alp
 				logger.Error(err, "Giving up on trying to maintain desired state of the DNS record - changes are being overridden")
 				return noRequeueDuration, err
 			}
-
 		}
 		dnsRecord.Status.ValidFor = validationRequeueTime.String()
 		setDNSRecordCondition(dnsRecord, string(v1alpha1.ConditionTypeReady), metav1.ConditionFalse, "AwaitingValidation", "Awaiting validation")
--- a/internal/controller/dnsrecord_controller.go
+++ b/internal/controller/dnsrecord_controller.go
@@ -356,6 +356,7 @@ func (r *DNSRecordReconciler) applyChanges(ctx context.Context, dnsRecord *v1alp
 		Policies: []externaldnsplan.Policy{policy},
 		Current:  zoneEndpoints,
 		Desired:  specEndpoints,
+		Previous: statusEndpoints,
 		//Note: We can't just filter domains by `managedZone.Spec.DomainName` it needs to be the exact root domain for this particular record
 		DomainFilter:   externaldnsendpoint.MatchAllDomainFilters{&rootDomainFilter},
 		ManagedRecords: managedDNSRecordTypes,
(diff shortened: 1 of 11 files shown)


maleck13 (commented):
Ok made a first pass. I have some understanding of what this is doing now. But can't say I know enough at this point to say it right or wrong I guess. I wil give it another pass later today

internal/controller/dnsrecord_controller.go
```diff
@@ -365,6 +366,10 @@ func (r *DNSRecordReconciler) applyChanges(ctx context.Context, dnsRecord *v1alp
 
 	plan = plan.Calculate()
 
+	if err = plan.ConflictError(); err != nil {
```
maleck13, line 369: nice so conflicts in this pr now also

internal/controller/dnsrecord_controller_test.go
```diff
@@ -159,4 +165,75 @@ var _ = Describe("DNSRecordReconciler", func() {
 		}, TestTimeoutLong, time.Second).Should(Succeed())
 	})
 
+	It("should not allow second record to change the type", func() {
```
maleck13, line 168: 👍

maleck13 (approved):

internal/controller/dnsrecord_controller.go
mikenairn: Yes, when i was writing the issue for it (https://github.com/Kuadrant/dns-operator/issues/85), i was looking into the best way to do it and pretty much just did it, so carried on. Kept it fairly simple anyway, just collect errors and expose them from the plan to query later.

Was thinking this morning i might make it more generic so we can just add all types of possible errors to it. We can use it for the missing managed record errors we want for the likes of default geo etc..

PR #81, opened 2024-04-17
Randomize duration bug

makslion:
prevent usage of `randomizeDuration()` for less than a second and address an issue when it would randomize itself into zero duration

### DIFF
--- a/internal/common/helper.go
+++ b/internal/common/helper.go
@@ -9,6 +9,10 @@ import (
 // RandomizeDuration randomizes duration for a given variance.
 // variance is expected to be of a format 0.1 for 10%, 0.5 for 50% and so on
 func RandomizeDuration(variance float64, duration time.Duration) time.Duration {
+	// do not allow less than a second requeue
+	if duration.Milliseconds() < 1000 {
+		duration = time.Second * 1
+	}
 	// we won't go smaller than a second - using milliseconds to have a relatively big number to randomize
 	millisecond := float64(duration.Milliseconds())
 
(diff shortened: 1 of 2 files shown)


internal/controller/dnsrecord_controller.go
```diff
@@ -47,11 +47,12 @@ const (
 	DNSRecordFinalizer        = "kuadrant.io/dns-record"
 	WriteCounterLimit         = 20
 	validationRequeueVariance = 0.5
+	DefaultValidationDuration = time.Second * 5
```
philbrookes, line 50: does this mean the delay range is from 2.5 to 7.5?

internal/common/helper.go
```diff
@@ -9,6 +9,10 @@ import (
 // RandomizeDuration randomizes duration for a given variance.
 // variance is expected to be of a format 0.1 for 10%, 0.5 for 50% and so on
 func RandomizeDuration(variance float64, duration time.Duration) time.Duration {
+	// do not allow less than a second requeue
+	if duration.Milliseconds() < 1000 {
```
philbrookes, line 13: In the old code, how was the duration ever anything other than `time.Second * 5`? I'm not understanding how the original problem was occurring.

philbrookes: discussed in person.

internal/controller/dnsrecord_controller.go
philbrookes: discussed in person.

PR #99, opened 2024-04-29
fix: DNSRecord status updates on errors

### LINKED ISSUE #96: bug: errors in the dns record controller do no backoff and continue to retry
**What**

When an error occurs in the dns record controller(provider error for example), the controller appears to no longer return an error from the reconcile and so never goes into the exponential backoff of retries that would normally happen in this case. It currently instead goes into a never ending loop which will cause it to reach API request limits very quickly. 

If there is an error [here](https://github.com/Kuadrant/dns-operator/blob/main/internal/controller/dnsrecord_controller.go#L140), or anywhere really, it should be returning `return ctrl.Result{}, err` from Reconcile. 

**Steps to reproduce**

- set up a DNSPolicy with an invalid GEO in a single cluster

mikenairn:
fixes #96 

Fixes an issue where the DNSRecord status was incorrectly being updated on every reconciliation even when errors were occuring causing the reconciler to go into a never ending loop and never exponentially backoff as would be expected in an error case.

### DIFF
--- a/internal/controller/dnsrecord_controller.go
+++ b/internal/controller/dnsrecord_controller.go
@@ -133,7 +133,7 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		reason = "ValidationError"
 		message = fmt.Sprintf("validation of DNSRecord failed: %v", err)
 		setDNSRecordCondition(dnsRecord, string(v1alpha1.ConditionTypeReady), metav1.ConditionFalse, reason, message)
-		return r.updateStatus(ctx, previous, dnsRecord, noRequeueDuration)
+		return r.updateStatus(ctx, previous, dnsRecord, noRequeueDuration, err)
 	}
 
 	// Publish the record
@@ -142,7 +142,7 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		reason = "ProviderError"
 		message = fmt.Sprintf("The DNS provider failed to ensure the record: %v", provider.SanitizeError(err))
 		setDNSRecordCondition(dnsRecord, string(v1alpha1.ConditionTypeReady), metav1.ConditionFalse, reason, message)
-		return r.updateStatus(ctx, previous, dnsRecord, noRequeueDuration)
+		return r.updateStatus(ctx, previous, dnsRecord, noRequeueDuration, err)
 	}
 	// success
 	dnsRecord.Status.ObservedGeneration = dnsRecord.Generation
@@ -152,19 +152,30 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		return ctrl.Result{}, err
 	}
 
-	return r.updateStatus(ctx, previous, dnsRecord, requeueAfter)
+	return r.updateStatus(ctx, previous, dnsRecord, requeueAfter, nil)
 }
 
-func (r *DNSRecordReconciler) updateStatus(ctx context.Context, previous, current *v1alpha1.DNSRecord, requeueAfter time.Duration) (reconcile.Result, error) {
-	current.Status.QueuedFor = metav1.NewTime(reconcileStart.Add(requeueAfter))
+func (r *DNSRecordReconciler) updateStatus(ctx context.Context, previous, current *v1alpha1.DNSRecord, requeueAfter time.Duration, specErr error) (reconcile.Result, error) {
+	if specErr != nil {
+		current.Status.WriteCounter = previous.Status.WriteCounter
+		current.Status.ValidFor = previous.Status.ValidFor
+		current.Status.QueuedAt = previous.Status.QueuedAt
+		current.Status.QueuedFor = previous.Status.QueuedFor
+	}
 
 	if !equality.Semantic.DeepEqual(previous.Status, current.Status) {
-		updateError := r.Status().Update(ctx, current)
-		if apierrors.IsConflict(updateError) {
-			return ctrl.Result{Requeue: true}, nil
+		if updateError := r.Status().Update(ctx, current); updateError != nil {
+			if apierrors.IsConflict(updateError) {
+				return ctrl.Result{Requeue: true}, nil
+			}
+			return ctrl.Result{}, updateError
 		}
-		return ctrl.Result{RequeueAfter: requeueAfter}, updateError
 	}
+
+	if specErr != nil {
+		return ctrl.Result{Requeue: true}, nil
+	}
+
 	return ctrl.Result{RequeueAfter: requeueAfter}, nil
 }
 
--- a/internal/external-dns/plan/plan.go
+++ b/internal/external-dns/plan/plan.go
@@ -437,13 +437,14 @@ func (e *managedRecordSetChanges) calculateDesired(update *endpointUpdate) {
 	}
 
 	currentCopy := update.current.DeepCopy()
+	desiredCopy := update.desired.DeepCopy()
 
 	// A Records can be merged, but we remove the known previous target values first in order to ensure potentially stale values are removed
 	if update.current.RecordType == endpoint.RecordTypeA {
 		if update.previous != nil {
 			removeEndpointTargets(update.previous.Targets, currentCopy)
 		}
-		mergeEndpointTargets(update.desired, currentCopy)
+		mergeEndpointTargets(desiredCopy, currentCopy)
 	}
 
 	// CNAME records can be merged, it's expected that the provider implementation understands that a CNAME might have
@@ -452,15 +453,13 @@ func (e *managedRecordSetChanges) calculateDesired(update *endpointUpdate) {
 		if update.previous != nil {
 			removeEndpointTargets(update.previous.Targets, currentCopy)
 		}
-		mergeEndpointTargets(update.desired, currentCopy)
+		mergeEndpointTargets(desiredCopy, currentCopy)
 
 		//ToDo manirn Check this is actually needed, and if it is add a test that requires it to be here
-		if len(update.desired.Targets) <= 1 {
+		if len(desiredCopy.Targets) <= 1 {
 			log.Debugf("skipping check for managed dnsNames for CNAME with single target value")
 			return
 		}
-
-		desiredCopy := update.desired.DeepCopy()
 
 		// Calculate if any of the new desired targets are also managed dnsNames within this record set.
 		// If a target is not managed, do nothing and continue with the current targets.
@@ -476,7 +475,7 @@ func (e *managedRecordSetChanges) calculateDesired(update *endpointUpdate) {
 
 				// If the target has no owners we can just remove it
 				if len(tOwners) == 0 {
-					removeEndpointTarget(t, update.desired)
+					removeEndpointTarget(t, desiredCopy)
 					break
 				}
 
@@ -490,13 +489,14 @@ func (e *managedRecordSetChanges) calculateDesired(update *endpointUpdate) {
 						}
 					}
 					if !hasMutualOwner {
-						removeEndpointTarget(t, update.desired)
+						removeEndpointTarget(t, desiredCopy)
 					}
 				}
 
 			}
 		}
 	}
+	update.desired = desiredCopy
 }
 
 func removeEndpointTarget(target string, endpoint *endpoint.Endpoint) {
--- a/test/e2e/multi_record_test.go
+++ b/test/e2e/multi_record_test.go
@@ -119,7 +119,7 @@ var _ = Describe("Multi Record Test", func() {
 				err := k8sClient.Create(ctx, dnsRecord1)
 				Expect(err).ToNot(HaveOccurred())
 
-				By("creating dnsrecord " + dnsRecord1.Name)
+				By("creating dnsrecord " + dnsRecord2.Name)
 				err = k8sClient.Create(ctx, dnsRecord2)
 				Expect(err).ToNot(HaveOccurred())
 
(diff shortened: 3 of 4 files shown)


internal/controller/dnsrecord_controller.go
```diff
@@ -371,9 +381,13 @@ func (r *DNSRecordReconciler) applyChanges(ctx context.Context, dnsRecord *v1alp
...
 
-	dnsRecord.Status.ValidFor = defaultRequeueTime.String()
-	dnsRecord.Status.QueuedAt = reconcileStart
 	if plan.Changes.HasChanges() {
+		logger.Info("Applying changes")
+		err = registry.ApplyChanges(ctx, plan.Changes)
+		if err != nil {
+			return noRequeueDuration, err
+		}
+
```
makslion, line 390: If there is a situation when generation has changed and we failed to `.ApplyChanges()` we will not reset the counter and on the next run (if we succeed to apply the changes) the generation will be the same so the counter will not be reset even through the generation change. Should we move this block in the beginning on the logic then? :thinking: I might be missing something

internal/controller/dnsrecord_controller.go
```diff
@@ -387,19 +401,16 @@ func (r *DNSRecordReconciler) applyChanges(ctx context.Context, dnsRecord *v1alp
...
 	} else {
 		logger.Info("All records are already up to date")
 		dnsRecord.Status.WriteCounter = 0
 		wrtiteCounter.WithLabelValues(dnsRecord.Name, dnsRecord.Namespace).Set(0)
 		setDNSRecordCondition(dnsRecord, string(v1alpha1.ConditionTypeReady), metav1.ConditionTrue, "ProviderSuccess", "Provider ensured the dns record")
 	}
 
+	dnsRecord.Status.ValidFor = defaultRequeueTime.String()
+	dnsRecord.Status.QueuedAt = reconcileStart
+
```
makslion, line 414: I think I found a bug in the way it was before :laughing:  Ok, so we are never really returning the `validationRequeueTime` from the `aaplyChanges` function. The reason it was passing tests - integration suite was able to get validation on the `defaultValidation`. Basically the `.Status.ValidFor` is a return value for the validation time. In the changes you are proposing we are setting that status always to the `defaultRequeuTime` while we should set it to either `ValidationRequeueTime` or `defaultRequeueTime` based on if we had changes or not and we should use the appropriate return value.

internal/controller/dnsrecord_controller.go
```diff
@@ -156,15 +156,25 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 }
 
 func (r *DNSRecordReconciler) updateStatus(ctx context.Context, previous, current *v1alpha1.DNSRecord, requeueAfter time.Duration) (reconcile.Result, error) {
-	current.Status.QueuedFor = metav1.NewTime(reconcileStart.Add(requeueAfter))
+	if requeueAfter != noRequeueDuration {
+		current.Status.QueuedFor = metav1.NewTime(reconcileStart.Add(requeueAfter))
+	}
```
makslion, line 161: This is one of two problems this PR is solving - constant updates to this field were causing the reconciler to go nuts. If we aren't setting/maintaining this field always (in other words it won't always be a reliable source of information) we better reset it to zero time. With something like this:
```
if requeueAfter != noRequeueDuration {
		current.Status.QueuedFor = metav1.NewTime(reconcileStart.Add(requeueAfter))
	} else {
		current.Status.QueuedFor = metav1.NewTime(time.Time{})
	}
```
wdyt?

internal/controller/dnsrecord_controller.go
```diff
@@ -156,15 +156,25 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+			return ctrl.Result{}, updateError
 		}
-		return ctrl.Result{RequeueAfter: requeueAfter}, updateError
 	}
+
+	if requeueAfter == noRequeueDuration {
+		// `ctrl.Result{RequeueAfter: 0, Requeue: true}, nil` results in the same backoff behaviour as `ctrl.Result{}, err`
+		// but with no error reported in the logs
+		return ctrl.Result{RequeueAfter: 0, Requeue: true}, nil
+	}
```
makslion, line 176: So the `noRequeueDuration` isn't really a no requeue anymore? Should we call it an `errorRequeueDuration` instead? And the `return ctrl.Result{RequeueAfter: 0, Requeue: true}, nil` is confusing a bit. It will still requeue immediately as the boolean flag takes precedence but `RequeueAfter: 0` is:
1. Ignored if `Reqeue: true` 
2. Means "do not requeue" if `Requeue: false` 
I'd vote for `return ctrl.Result{Requeue: true}, nil` to be less confusing

internal/controller/dnsrecord_controller.go
```diff
@@ -156,15 +156,25 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
 
 	if !equality.Semantic.DeepEqual(previous.Status, current.Status) {
-		updateError := r.Status().Update(ctx, current)
-		if apierrors.IsConflict(updateError) {
-			return ctrl.Result{Requeue: true}, nil
+		if updateError := r.Status().Update(ctx, current); updateError != nil {
+			if apierrors.IsConflict(updateError) {
+				return ctrl.Result{Requeue: true}, nil
+			}
+			return ctrl.Result{}, updateError
```
makslion, line 168: in here we want to requeue if we have a conflict with no error returned. And if there is another error we want a default backoff with errors in the logs. If we make the last line to be `return ctrl.Result{RequeueAfter: requeueAfter}, updateError` it should do the same but we will have less of `if` statements around.

internal/controller/dnsrecord_controller.go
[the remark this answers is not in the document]
mikenairn: As discussed, no.

internal/controller/dnsrecord_controller.go
```diff
@@ -152,22 +152,43 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+
 	return ctrl.Result{RequeueAfter: requeueAfter}, nil
 }
 
+// calculateStatus calculate and apply the status of the dnsRecord based on what has happened during the reconcile.
+// ToDo This should be the only thing updating the status of the dnsRecord.
+func (r *DNSRecordReconciler) calculateStatus(dnsRecord *v1alpha1.DNSRecord, requeueAfter time.Duration, specErr error) {
+	if specErr != nil {
+		//Set these values to whatever they should be in an error state, they should not change again until we leave the error state.
+		dnsRecord.Status.WriteCounter = -1
```
philbrookes, line 182: I'm not  sure we should reset the writeCounter, I'd think we leave that unchanged.

mikenairn: What's the value in leaving it? Is it not meaningless in this scenario? We can possible leave it, but we need to ensure wherever it is getting changed doesn't change it on error or it potentially will keep updating which this what this is trying to avoid.

internal/controller/dnsrecord_controller.go
```diff
@@ -387,19 +408,18 @@ func (r *DNSRecordReconciler) applyChanges(ctx context.Context, dnsRecord *v1alp
 				return noRequeueDuration, err
 			}
 		}
-		dnsRecord.Status.ValidFor = validationRequeueTime.String()
```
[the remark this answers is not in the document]
mikenairn: I'm tempted to revert all these changes in `applyChanges` since now that the updateStatus is making sure these values are set to a consistent value on error the changes here are not really required to fix the bug.

That all said, this does also fix an issue where it didn't return `validationRequeueTime`, and also makes more sense to leave these updates to after we actually apply changes and know what is happening.

Why would you set `ValidFor` if you are entering a validation loop? Does that not suggest it's valid, even though we haven't done the validation loop yet?

internal/controller/dnsrecord_controller.go
philbrookes: yes, it should be unchanged in an error. 

The scenario I'm imagining is:
1) A controller is having to write it's changes a lot, incrementing this counter to make that visible
2) The controller gets ratelimited by the DNS provider (or network outage or whatever) temporarily.
3) With the above change, this error resets the counter to zero, but in reality the problem causing the multiple writes isn't resolved.

internal/controller/dnsrecord_controller.go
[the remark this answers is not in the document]
philbrookes: I would think: If there has been an error, validFor should be 0 (or nil, or similar) because it is not valid.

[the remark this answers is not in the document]
mikenairn: This isn't related to errors, ensuring it's set to 0 on error is dealt with in `calculateStatus`. 

What should `validFor` be set to between the time it applies changes and comes back round and validates that there are no new changes.

internal/controller/dnsrecord_controller.go
mikenairn: OK, i can see that. We will need the changes i made in `applyChanges` in that case as currently it bumps this before even doing anything.

internal/controller/dnsrecord_controller.go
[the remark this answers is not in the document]
makslion: that, actually, makes more sense than what we have now :thinking: and gives that bit of information for the future reconciliation [here](https://github.com/Kuadrant/dns-operator/pull/99/files#diff-adacb6a4589c65c2d853c0f59ca02d6b8fb50bb388b41de1d13e99b0d128cfe1R189) but in this case I'd prefer it to be 0 just like Phil suggested  even if we have no errors and just doing validation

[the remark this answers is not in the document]
philbrookes: I think it should still be the couple of seconds. If we think of this as... how long I'm willing to leave this before rechecking that it is still like I expect it to be. Then this is the period of time until it goes back to the API right? Initially 5s or whatever, then 10, 20, 40. It's never really saying "it's valid" but instead more like "I believe it's valid".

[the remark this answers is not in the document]
mikenairn: OK `validFor` as a name does not suggest what you are saying, but changing it here is not needed to resolve the bug so will add it back but inside the code where we expect it to validating not at the start before we do anything.

internal/controller/dnsrecord_controller.go
mikenairn: updated

internal/controller/dnsrecord_controller.go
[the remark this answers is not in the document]
mikenairn: added back, sets all the values to what is expected deepening on whether there are changes or not

internal/controller/dnsrecord_controller.go
```diff
@@ -371,9 +387,14 @@ func (r *DNSRecordReconciler) applyChanges(ctx context.Context, dnsRecord *v1alp
 		return noRequeueDuration, err
 	}
 
-	dnsRecord.Status.ValidFor = defaultRequeueTime.String()
-	dnsRecord.Status.QueuedAt = reconcileStart
 	if plan.Changes.HasChanges() {
+		logger.Info("Applying changes")
```
philbrookes, line 391: should this be removed?

internal/controller/dnsrecord_controller.go
```diff
@@ -387,19 +408,22 @@ func (r *DNSRecordReconciler) applyChanges(ctx context.Context, dnsRecord *v1alp
...
-		dnsRecord.Status.WriteCounter = 0
-		wrtiteCounter.WithLabelValues(dnsRecord.Name, dnsRecord.Namespace).Set(0)
-		setDNSRecordCondition(dnsRecord, string(v1alpha1.ConditionTypeReady), metav1.ConditionTrue, "ProviderSuccess", "Provider ensured the dns record")
+
+		logger.Info(fmt.Sprintf("endpoints were updated, requing for validation in %s", validationRequeueTime.String()))
+		return validationRequeueTime, nil
 	}
 
+	//Set WriteCounter, ValidFor, QueuedAt, Ready condition true and return defaultRequeueTime when we have no changes!
+	dnsRecord.Status.WriteCounter = 0
```
philbrookes, line 421: WriteCounter shouldn't be getting reset here either, right? It should only reset to 0 when the generation changes

internal/controller/dnsrecord_controller.go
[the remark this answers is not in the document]
makslion: the `validFor` will be exponential once we validated the record, but before the validation occurs that should be 0 :thinking:  the "I'll poke API at this time" is the `queuedFor` field. We aren't using it at the moment but that could be changed. 

Right now it looks like this: 

- got a fresh record, applied changes.
-  Set `queuedAt` as the current reconcile time, `validFor` is set to the duration in which i will poke the API and `queuedFor` = `queuedAt`+`validFor` 
- got record for validation. if  `queuedAt`+`validFor` is before `time.Now()` 
- repeat step 2

We could ask the logic to check for premature arrival of the record based on the `queuedFor` instead of doing that math. So the `queuedAt` and `validFor` fields will be just for humans to read. This should let us to use the `validFor` in a way it is called - those values are validated and will be valid for that much since the `queuedAt`

internal/controller/dnsrecord_controller.go
mikenairn: This is the same logic as main

internal/controller/dnsrecord_controller.go
```diff
@@ -152,19 +152,30 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
 
-func (r *DNSRecordReconciler) updateStatus(ctx context.Context, previous, current *v1alpha1.DNSRecord, requeueAfter time.Duration) (reconcile.Result, error) {
-	current.Status.QueuedFor = metav1.NewTime(reconcileStart.Add(requeueAfter))
+func (r *DNSRecordReconciler) updateStatus(ctx context.Context, previous, current *v1alpha1.DNSRecord, requeueAfter time.Duration, specErr error) (reconcile.Result, error) {
+	if specErr != nil {
+		current.Status.WriteCounter = previous.Status.WriteCounter
+		current.Status.ValidFor = previous.Status.ValidFor
+		current.Status.QueuedAt = previous.Status.QueuedAt
+		current.Status.QueuedFor = previous.Status.QueuedFor
+	}
```
mikenairn, line 164: If we get an error ensure that these status fileds are not updated from there previous values

philbrookes (approved):

Issue #84, opened 2024-04-18
Enhance DNS Record conflict detection


makslion:
## What 
Curently the DNS Record lifecycle looks like this: 
1. Writre changes and set `Ready - false` with `Awaiting Validation`
2. Submit for a validation requeue (5 seconds +/- 50%)
3. Once received if no changes needed mark as `Ready - true` 
4. Submit for a defaultRequeue time (15 minutes) 

If there is a conflict between two controllers the scenario looks like this: 
1. Record received after 15 minutes and there are changes needed 
2. Writre changes, increase the write counter (because generation is the same) and set `Ready - false` with `Awaiting Validation`
3. Submit for a validation requeue (5 seconds +/- 50%)
4. Once received if no changes are needed mark it as `Ready - true` and reset the counter
5. Submit for a defaultRequeue time (15 minutes) 

This will lead to us never reporting a problem with the configuration unless two controllers manage to do their reconciliation within the validation requeue time. In this case, the write counter will reach the limit and we will stop reconciliation.  

## How 
We would need to change: 
- `defaultRequeueTime` should be changed from the constant 15 minutes to an exponential on each successful reconcile loop. Starting with the validation time (5 seconds) and reaching 15 minutes in the end. In other words, after successful validation the requeue times should look like this: 5 sec, 10 sec, 20 sec, 40 sec, 80 sec ... `defaultRequeueTime`.
- We need to reset the Write counter ***only*** on a generation change 
- We need to eliminate the upper limit of the counter and associated `Ready - false` condition. Meaning that there would be only two `Ready` conditions: `Ready - true` and `Ready - false` with `Awaiting Validation` 
- We need to introduce an integration test for this scenario: create two conflicting records with `defaultRequeueTime`/2 duration between and watch the controller converge into the `validationRequeueTime` and marking booth as `AwaitingValidation` with the counter going up. After this resolve the conflict and watch the records become ready (The counter should stay the same on the one that was not changed and should be reset to 0 on the changed record)

PR #100, opened 2024-04-29
GH-84 refactor DNSRecord conflict detection

maksymvavilov:
Currently contains only the logic to change behaviour on conflict detection and conflict resolution. 
Also moves all the logic for updating the status of the DNS Record to the `updateStatus()` func. 
Finally, allows you to override default validation timeout using a flag (defaults to the 5 sec)

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -48,8 +48,9 @@ var (
 )
 
 const (
-	RequeueDuration  = time.Minute * 15
-	ValidityDuration = time.Minute * 14
+	RequeueDuration           = time.Minute * 15
+	ValidityDuration          = time.Minute * 14
+	DefaultValidationDuration = time.Second * 5
 )
 
 func init() {
@@ -63,18 +64,22 @@ func main() {
 	var metricsAddr string
 	var enableLeaderElection bool
 	var probeAddr string
-	var requeueTime time.Duration
+	var minRequeueTime time.Duration
 	var validFor time.Duration
+	var maxRequeueTime time.Duration
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
-	flag.DurationVar(&requeueTime, "requeue-time", RequeueDuration,
-		"Duration for the validation reconciliation loop. "+
+	flag.DurationVar(&maxRequeueTime, "max-requeue-time", RequeueDuration,
+		"The maximum times it takes between reconciliations of DNS Record "+
 			"Controls how ofter record is reconciled")
 	flag.DurationVar(&validFor, "valid-for", ValidityDuration,
 		"Duration when the record is considered to hold valid information"+
+			"Controls if we commit to the full reconcile loop")
+	flag.DurationVar(&minRequeueTime, "min-requeue-time", DefaultValidationDuration,
+		"The minimal timeout between calls to the DNS Provider"+
 			"Controls if we commit to the full reconcile loop")
 	opts := zap.Options{
 		Development: true,
@@ -127,7 +132,7 @@ func main() {
 		Client:          mgr.GetClient(),
 		Scheme:          mgr.GetScheme(),
 		ProviderFactory: providerFactory,
-	}).SetupWithManager(mgr, requeueTime, validFor); err != nil {
+	}).SetupWithManager(mgr, maxRequeueTime, validFor, minRequeueTime); err != nil {
 		setupLog.Error(err, "unable to create controller", "controller", "DNSRecord")
 		os.Exit(1)
 	}
--- a/internal/controller/dnsrecord_controller.go
+++ b/internal/controller/dnsrecord_controller.go
@@ -18,7 +18,6 @@ package controller
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"strings"
 	"time"
@@ -28,7 +27,6 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/utils/clock"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
@@ -46,18 +44,15 @@ import (
 
 const (
 	DNSRecordFinalizer        = "kuadrant.io/dns-record"
-	WriteCounterLimit         = 20
 	validationRequeueVariance = 0.5
-	DefaultValidationDuration = time.Second * 5
 )
 
 var (
-	defaultRequeueTime    time.Duration
-	validationRequeueTime time.Duration
-	noRequeueDuration     = time.Duration(0)
-	validFor              time.Duration
-	reconcileStart                    = metav1.Time{}
-	Clock                 clock.Clock = clock.RealClock{}
+	defaultRequeueTime          time.Duration
+	defaultValidationRequeue    time.Duration
+	randomizedValidationRequeue time.Duration
+	validFor                    time.Duration
+	reconcileStart              metav1.Time
 )
 
 // DNSRecordReconciler reconciles a DNSRecord object
@@ -77,7 +72,7 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	reconcileStart = metav1.Now()
 
 	// randomize validation reconcile delay
-	validationRequeueTime = common.RandomizeDuration(validationRequeueVariance, DefaultValidationDuration)
+	randomizedValidationRequeue = common.RandomizeDuration(validationRequeueVariance, defaultValidationRequeue)
 
 	previous := &v1alpha1.DNSRecord{}
 	err := r.Client.Get(ctx, client.ObjectKey{Namespace: req.Namespace, Name: req.Name}, previous)
@@ -91,19 +86,19 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	dnsRecord := previous.DeepCopy()
 
 	if dnsRecord.DeletionTimestamp != nil && !dnsRecord.DeletionTimestamp.IsZero() {
-		if err := r.ReconcileHealthChecks(ctx, dnsRecord); client.IgnoreNotFound(err) != nil {
+		if err = r.ReconcileHealthChecks(ctx, dnsRecord); client.IgnoreNotFound(err) != nil {
 			return ctrl.Result{}, err
 		}
-		requeueTime, err := r.deleteRecord(ctx, dnsRecord)
+		hadChanges, err := r.deleteRecord(ctx, dnsRecord)
 		if err != nil {
 			logger.Error(err, "Failed to delete DNSRecord")
-			return ctrl.Result{RequeueAfter: requeueTime}, err
+			return ctrl.Result{}, err
 		}
-		// if requeueTime returned  is the same as validationRequeueTime - the deleteRecord has successfully applied changes
+		// if hadChanges - the deleteRecord has successfully applied changes
 		// in this case we need to queue for validation to ensure DNS Provider retained changes
 		// before removing finalizer and deleting the DNS Record CR
-		if requeueTime == validationRequeueTime {
-			return ctrl.Result{RequeueAfter: requeueTime}, nil
+		if hadChanges {
+			return ctrl.Result{RequeueAfter: randomizedValidationRequeue}, nil
 		}
 
 		logger.Info("Removing Finalizer", "name", DNSRecordFinalizer)
@@ -118,14 +113,14 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	}
 
 	if !controllerutil.ContainsFinalizer(dnsRecord, DNSRecordFinalizer) {
-		dnsRecord.Status.QueuedFor = metav1.NewTime(reconcileStart.Add(validationRequeueTime))
+		dnsRecord.Status.QueuedFor = metav1.NewTime(reconcileStart.Add(randomizedValidationRequeue))
 		logger.Info("Adding Finalizer", "name", DNSRecordFinalizer)
 		controllerutil.AddFinalizer(dnsRecord, DNSRecordFinalizer)
 		err = r.Update(ctx, dnsRecord)
 		if err != nil {
 			return ctrl.Result{}, err
 		}
-		return ctrl.Result{RequeueAfter: validationRequeueTime}, nil
+		return ctrl.Result{RequeueAfter: randomizedValidationRequeue}, nil
 	}
 
 	var reason, message string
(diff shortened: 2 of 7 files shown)


internal/controller/dnsrecord_controller.go
```diff
@@ -264,6 +263,22 @@ func generationChanged(record *v1alpha1.DNSRecord) bool {
...
+// exponentialRequeueTime consumes the current time and doubles it until it reaches defaultRequeueTime
+func exponentialRequeueTime(lastRequeueTime string) time.Duration {
+	lastRequeue, err := time.ParseDuration(lastRequeueTime)
+	// corrupted DNSRecord. This value naturally set only via time.Duration.String() call
+	if err != nil {
+		// default to the least confidence timeout
+		return validationRequeueTime
+	}
+	// double the duration. Return the max timeout if overshoot
+	newReqeueue := lastRequeue * 2
```
philbrookes, line 275: we don't want to just take the last requeue time and double it, as the initial value is randomized. So we need to work off of the initial values, e.g.

requeueTime := min(max_requeue, math.Pow(requeueDuration, math.Pow(2, requeueCount)))

makslion: I'd have it double instead of raising to the power - makes it grow too fast IMO. 
The double sequence will be (seconds): 5, 10, 20, 40, 80, 160, 320, 640, 900
While the power 2 sequence: is 5, 25, 625, 900. 
And we don't keep the `requeueCount`. I'll reset the timeout on the successful validation to ensure we are not doubling the randomized value.

makslion:
It has as a base Mike's PR #103 I guess it is a merge race! (that is there will be changes on that PR) If that looks good we can have one PR to rule 'em all

internal/controller/dnsrecord_controller.go
```diff
@@ -407,10 +418,19 @@ func (r *DNSRecordReconciler) applyChanges(ctx context.Context, dnsRecord *v1alp
...
+	// reset the counter on the gen change regardless of having changes in the plan
+	if generationChanged(dnsRecord) {
 		dnsRecord.Status.WriteCounter = 0
 		wrtiteCounter.WithLabelValues(dnsRecord.Name, dnsRecord.Namespace).Set(0)
-		setDNSRecordCondition(dnsRecord, string(v1alpha1.ConditionTypeReady), metav1.ConditionTrue, "ProviderSuccess", "Provider ensured the dns record")
 	}
 
-	return defaultRequeueTime, nil
+	return requeueTime, nil
 }
```
mikenairn, line 436: Do we have an issue to remove all this logic from applyChanges? It's really hard to read and the status is still being updated in places it shouldn't i.e. before it's even done anything

makslion: There was one but it seems like I can't find it. Changed anyway. Now it should be easier to read\maintain

makslion:
Also added an option to override validation timeout for the DNS controller. With the "default" 5 seconds some tests could be flaky and would timeout after 15 secs.  In this repo the integration suite doesn't take ages to run, but I'd rather make tests faster if I can rather than increasing timeouts

internal/controller/dnsrecord_controller.go
philbrookes: ah, I see, if we aren't keeping the requeueCount then yes, we cannot do it this way.

cmd/main.go
```diff
@@ -69,11 +70,14 @@ func main() {
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
 	flag.DurationVar(&requeueTime, "requeue-time", RequeueDuration,
-		"Duration for the validation reconciliation loop. "+
+		"The maximum times it takes between reconciliations of DNS Record "+
```
philbrookes, line 73: requeue-time feels slightly off here, this should be called something like "maximum-requeue-time" for more clarity.

cmd/main.go
```diff
@@ -69,11 +70,14 @@ func main() {
...
 			"Controls how ofter record is reconciled")
 	flag.DurationVar(&validFor, "valid-for", ValidityDuration,
 		"Duration when the record is considered to hold valid information"+
 			"Controls if we commit to the full reconcile loop")
+	flag.DurationVar(&validFor, "validation-duration", DefaultValidationDuration,
```
philbrookes, line 78: this looks like a dupe of the line above?

internal/controller/dnsrecord_controller.go
```diff
@@ -133,36 +128,83 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+		requeueTime = validationRequeueTime
+		setDNSRecordCondition(current, string(v1alpha1.ConditionTypeReady), metav1.ConditionFalse, "AwaitingValidation", "Awaiting validation")
+	} else {
+		logger.Info("All records are already up to date")
+		// reset the valid for from randomized value to a fixed value once validation succeeds
+		if !meta.IsStatusConditionTrue(current.Status.Conditions, string(v1alpha1.ConditionTypeReady)) {
+			requeueTime = exponentialRequeueTime(validationRequeueTime.String())
+		} else {
+			// uses current.Status.ValidFor as the last requeue duration. Double it.
+			requeueTime = exponentialRequeueTime(current.Status.ValidFor)
```
philbrookes, line 188: this still reads to me like:
```
if it wasn't previously ready {
   requeue for randomized time
} else {
   requeue for double the last requeue time
}
```

I'm not seeing how this avoids doubling the randomized number.

makslion: used the wrong var while moving/renaming things.
Now it should be fine

cmd/main.go
makslion: wrong war was holding the value. Fixed

cmd/main.go
makslion: renamed

internal/controller/dnsrecord_controller.go
```diff
@@ -133,36 +128,83 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+		setDNSRecordCondition(current, string(v1alpha1.ConditionTypeReady), metav1.ConditionTrue, "ProviderSuccess", "Provider ensured the dns record")
+	}
+
+	// valid for is always a requeue time
+	current.Status.ValidFor = requeueTime.String()
+
+	// reset the counter on the gen change regardless of having changes in the plan
+	if generationChanged(current) {
+		current.Status.WriteCounter = 0
+		wrtiteCounter.WithLabelValues(current.Name, current.Namespace).Set(0)
```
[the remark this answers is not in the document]
makslion: fixed

internal/provider/aws/health.go
```diff
@@ -142,8 +142,13 @@ func (r *Route53HealthCheckReconciler) Reconcile(ctx context.Context, spec provi
...
 	healthCheck, found, err := r.findHealthCheck(ctx, probeStatus)
 	if err != nil {
-		return provider.HealthCheckResult{}, err
+		var awserror awserr.Error
+		_ = errors.As(err, &awserror)
+		if awserror.Code() != route53.ErrCodeNoSuchHealthCheck {
+			return provider.HealthCheckResult{}, err
+		}
 	}
+
```
mikenairn, line 151: What is this change for?

internal/controller/suite_test.go
```diff
@@ -112,7 +112,7 @@ var _ = BeforeSuite(func() {
 		Client:          mgr.GetClient(),
 		Scheme:          mgr.GetScheme(),
 		ProviderFactory: providerFactory,
-	}).SetupWithManager(mgr, RequeueDuration, ValidityDuration)
+	}).SetupWithManager(mgr, RequeueDuration, ValidityDuration, DefaultValidationDuration)
```
mikenairn, line 115: What is the relationship between these? Can the controller be started with any combination of values of these things and behave in a an appropriate way?

internal/controller/dnsrecord_controller.go
```diff
@@ -257,42 +296,57 @@ func (r *DNSRecordReconciler) publishRecord(ctx context.Context, dnsRecord *v1al
...
+}
+
+func recordReceivedPrematurely(record *v1alpha1.DNSRecord) (bool, time.Duration) {
+	requeueIn := validFor
+	if record.Status.ValidFor != "" {
+		requeueIn, _ = time.ParseDuration(record.Status.ValidFor)
+	}
+	expiryTime := metav1.NewTime(record.Status.QueuedAt.Add(requeueIn))
+	return !generationChanged(record) && reconcileStart.Before(&expiryTime), requeueIn
 }
```
mikenairn, line 328: What does `ReceivedPrematurely` mean? Can you add a description to the function?

internal/controller/dnsrecord_controller.go
```diff
@@ -393,38 +447,12 @@ func (r *DNSRecordReconciler) applyChanges(ctx context.Context, dnsRecord *v1alp
...
-	} else {
-		logger.Info("All records are already up to date")
-		dnsRecord.Status.WriteCounter = 0
-		wrtiteCounter.WithLabelValues(dnsRecord.Name, dnsRecord.Namespace).Set(0)
-		setDNSRecordCondition(dnsRecord, string(v1alpha1.ConditionTypeReady), metav1.ConditionTrue, "ProviderSuccess", "Provider ensured the dns record")
+		return true, err
 	}
-
-	return defaultRequeueTime, nil
+	return false, nil
```
mikenairn, line 457: The boolean returned here will mean there were changes and they were successfully, iF the plan has changes but they aren't applied it will return false and we rely on the error wherever we are needing to update status right?

Can you add a description to make it clear what the purpose of this boolean value is and what the value is expected to be when there are calculated changes but they can't be applied for whatever reason.

internal/controller/dnsrecord_controller.go
```diff
@@ -91,19 +86,19 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	dnsRecord := previous.DeepCopy()
 
 	if dnsRecord.DeletionTimestamp != nil && !dnsRecord.DeletionTimestamp.IsZero() {
-		if err := r.ReconcileHealthChecks(ctx, dnsRecord); client.IgnoreNotFound(err) != nil {
+		if err = r.ReconcileHealthChecks(ctx, dnsRecord); client.IgnoreNotFound(err) != nil {
 			return ctrl.Result{}, err
 		}
-		requeueTime, err := r.deleteRecord(ctx, dnsRecord)
+		hadChanges, err := r.deleteRecord(ctx, dnsRecord)
```
mikenairn, line 92:
```suggestion
		appliedChanges, err := r.deleteRecord(ctx, dnsRecord)
```

Looks to me like it tells you if changes were applied (calculated changes and successfully applied them) rather than just if there were any. Maybe the name of this variable should reflect that?

internal/controller/dnsrecord_controller_test.go
```diff
@@ -360,7 +335,40 @@ var _ = Describe("DNSRecordReconciler", func() {
...
+		}, TestTimeoutLong, time.Second).Should(Succeed())
+
+		By("fixing conflict with dnsrecord " + dnsRecord2.Name + " with endpoint dnsName: `foo.example.com` and target: `127.0.0.1`")
+		Eventually(func(g Gomega) {
+			// refresh the second record CR
+			err := k8sClient.Get(ctx, client.ObjectKeyFromObject(dnsRecord2), dnsRecord2)
+			g.Expect(err).NotTo(HaveOccurred())
+		}, TestTimeoutShort, time.Second).Should(Succeed())
+		dnsRecord2.Spec.Endpoints = getDefaultTestEndpoints()
+		Expect(k8sClient.Update(ctx, dnsRecord2)).To(Succeed())
```
mikenairn, line 348: Move these inside the eventually. If it fails for some random conflict it will bail out here.

Personal opinion maybe, but i would also much rather see in the spec itself what is being applied to make this work. Just calling `getDefaultTestEndpoints()` really isn't obvious and requires anyone reading this to understand what that is doing in order to know what is being applied. We seem to keep trying to re-factor tests unnecessarily IMO.

internal/controller/helper_test.go
```diff
@@ -47,31 +49,18 @@ func testBuildInMemoryCredentialsSecret(name, ns string) *v1.Secret {
...
 		{
-			DNSName: "nope.example.com",
+			DNSName: dnsName,
 			Targets: []string{
-				"127.0.0.1",
+				ip,
 			},
 			RecordType:       "A",
-			SetIdentifier:    "",
+			SetIdentifier:    "foo",
```
mikenairn, line 63: Why add the SetIdentifier?

Personally i think these "helper" functions just lead to confusion in the tests. There is no real one size fits all endpoint list and we should probably be explicitly defining what we expect it to be for each spec.

philbrookes: I think it'd be best to convert this to a fixture, like this: https://github.com/Kuadrant/dns-operator/blob/main/test/e2e/healthcheck_test.go#L62 and this: https://github.com/Kuadrant/dns-operator/blob/main/test/e2e/fixtures/healthcheck_test/geo-dnsrecord-healthchecks.yaml

mikenairn: > I think it'd be best to convert this to a fixture, like this: https://github.com/Kuadrant/dns-operator/blob/main/test/e2e/healthcheck_test.go#L62 and this: https://github.com/Kuadrant/dns-operator/blob/main/test/e2e/fixtures/healthcheck_test/geo-dnsrecord-healthchecks.yaml

These are the integration tests where a much larger amount of variations in what we are testing would be expected. I'm not sure using fixtures is appropriate here, or would help with the readability of tests.

internal/provider/aws/health.go
makslion: fix for the bug when aws returns `NoSuchHealthCheck` error code and it is not treated well [here](https://github.com/Kuadrant/dns-operator/blob/0986c428ba16f04893e238e6016e881076b288a8/internal/controller/dnsrecord_controller.go#L89) by `client.IgnoreNotFound(err)`. It was not a problem before as we were not "validating" deletion so haven't really encountered such a scenario.

internal/controller/dnsrecord_controller.go
makslion: Added a comment. Hopefully, it will make it more obvious. What it does can be summarized as `return reconcileStart.Before(lastReconcile.Add(validFor))`

internal/controller/dnsrecord_controller.go
makslion: Not exactly. If the plan had changes we will return `true`. The success or failure in applying will affect only the `err` returned. I've added more comments around it to make it less confusing.

internal/controller/dnsrecord_controller.go
makslion: If I'm not mistaken it tells if there were any changes in the plan regardless of how successful it was in applying them

internal/controller/dnsrecord_controller_test.go
makslion: Moved inside `Eventually()`. 

We are using the same endpoint 4 times and with slight modification 3 more times. It just screams at me "I will break at the very first change to the tests". I made the change with `getDefaultTestEndpoints()` exactly because it took me a while to understand what names/IPs are healthy and which are not. Was intending it as simplification:thinking: but looks like it failed. Would you prefer it being "rolled back"?

internal/controller/helper_test.go
makslion: If `SetIdentifier` is an empty string it does not conflict records ([here](https://github.com/Kuadrant/dns-operator/blob/GH-84/internal/external-dns/plan/plan.go#L434-L437)). There is probably a smarter way of doing it

makslion: The way I read these tests is "we want to use valid/broken endpoints". As a person who isn't yet used to stare into enpoints it made it easier for me to see a func call with parameters that will differ compared to the whole struct each time. 
On the other hand if "helper" functions aren't helping the majority we should definitely remove those.

internal/controller/dnsrecord_controller.go
mikenairn: OK

internal/controller/dnsrecord_controller.go
```diff
@@ -134,36 +129,83 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+			}
+		}
+		return ctrl.Result{Requeue: true}, updateError
+	}
+
+	// success
+	if hadChanges {
+		// generation has not changed but there are changes.
+		// implies that they were overridden - bump write counter
+		if !generationChanged(current) {
```
philbrookes, line 175: this needs an else:
```
} else {
    current.Status.WriteCounter = 1
}
```

internal/controller/dnsrecord_controller.go
```diff
@@ -46,18 +44,15 @@ import (
...
 var (
-	defaultRequeueTime    time.Duration
-	validationRequeueTime time.Duration
-	noRequeueDuration     = time.Duration(0)
-	validFor              time.Duration
-	reconcileStart                    = metav1.Time{}
-	Clock                 clock.Clock = clock.RealClock{}
+	defaultRequeueTime       time.Duration
+	defaultValidationRequeue time.Duration
+	validationRequeueTime    time.Duration
```
philbrookes, line 53: This var is always the randomized requeue, right? I think the name makes this confusing, it would be better with something like: `randomizedRequeueTime`. Though I'm not convinced that it's required at this scope... why not just calculate it at the point it is required?

internal/controller/dnsrecord_controller.go
philbrookes: I see now this is handled further down.

internal/controller/dnsrecord_controller.go
```diff
@@ -134,36 +129,83 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+		return ctrl.Result{Requeue: true}, updateError
+	}
+
+	// success
+	if hadChanges {
+		// generation has not changed but there are changes.
+		// implies that they were overridden - bump write counter
+		if !generationChanged(current) {
+			current.Status.WriteCounter++
+			writeCounter.WithLabelValues(current.Name, current.Namespace).Inc()
```
philbrookes, line 177: should this involve the generation as well?

internal/controller/helper_test.go
mikenairn: > If `SetIdentifier` is an empty string it does not conflict records ([here](https://github.com/Kuadrant/dns-operator/blob/GH-84/internal/external-dns/plan/plan.go#L434-L437)). There is probably a smarter way of doing it

It won't conflict because it's not supposed to, it will merge the targets, which i kind of my point here. There are lots of permutations of how the endpoints could be configured that trying to write a single function that works for all tests is probably impossible and will just make tests harder to understand.

We can leave this for now, but we should look at doing it differently going forward since the endpoint and endpoint list on the record is the main thing we will be changing in tests to force different scenarios and you want to be able to see what they look like easier than this.

philbrookes:
This looks basically fine, other than a few small comments

philbrookes (approved):

internal/controller/dnsrecord_controller.go
makslion: It is used a few times across the code and if we will calculate it every time we need that we will end up with different values in one instance (right now will not cause any troubles, but might in future) and it seems cleaner to me to randomize once and then just use the var. But I do have a problem with naming :laughing:

internal/controller/dnsrecord_controller.go
makslion: We could :thinking:  but not sure that we should. Considering that we should have a record per listener there will be a lot happening in the metrics. I'm concerned about the dimensions of this metric. Would generation be useful considering that we reset it anyway on a generation change? I reason that whatever tool will be consuming this metric (Grafana most likely) will maintain the history to see write counts of the past gens.

PR #113, opened 2024-05-09
[WIP] adding health check e2e tests

philbrookes:


test/e2e/healthcheck_test.go
```diff
@@ -0,0 +1,189 @@
...
+			Expect(client.IgnoreNotFound(err)).ToNot(HaveOccurred())
+		}
+	})
+
+	Context("DNS Provider health checks", func() {
+		It("creates health checks for a health check spec", func(ctx SpecContext) {
+			probeHost := "app." + testZoneDomainName
+			if !slices.Contains(supportedHealthCheckProviders, strings.ToLower(testDNSProvider)) {
+				Skip(fmt.Sprintf("Health checks only supported with DNS Providers: %v", supportedHealthCheckProviders))
+			}
```
mikenairn, line 48: Should we have some test for providers that do not support health checks so we can test what will happen if you did add a health check?

test/e2e/healthcheck_test.go
```diff
@@ -0,0 +1,189 @@
...
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/kuadrant/dns-operator/api/v1alpha1"
+	"github.com/kuadrant/dns-operator/test/e2e/helpers"
+)
+
+const TestTimeoutMedium = 10 * time.Second
+const TestTimeoutLong = 60 * time.Second
```
mikenairn, line 23: Should probably be somewhere else, suite_test maybe?

test/e2e/healthcheck_test.go
```diff
@@ -0,0 +1,192 @@
...
+			healthChecksSupported := false
+			if slices.Contains(supportedHealthCheckProviders, strings.ToLower(testDNSProvider)) {
+				healthChecksSupported = true
+			}
+
+			geoCode := "EU"
+			if testDNSProvider == "gcp" {
+				geoCode = "europe-west1"
+			}
+			Expect(os.Setenv("TEST_DNS_GEO_CODE", geoCode)).To(BeNil())
```
mikenairn, line 37: Hmm, we need to think about what the consequences of using  `os.Setenv` are. Generally, in tests, we have found it to not be a great idea since it's global to all test specs (so can't run in parallel), and you are setting these values on the system running it so you need to make sure you unset it, or put back any old value. The golang testing framework had some helper for this from what i remember.

philbrookes: the best I can find is [this](https://www.yellowduck.be/posts/using-environment-variables-in-go-tests), but it seems that ginkgo doesn't expose the `testing.T`. So I will do this via a `BeforeEach/AfterEach` set /unset, if that seems suitable?

test/e2e/healthcheck_test.go
```diff
@@ -0,0 +1,196 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/strings/slices"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/kuadrant/dns-operator/api/v1alpha1"
+	"github.com/kuadrant/dns-operator/test/e2e/helpers"
+)
+
+// Test Cases covering multiple creation and deletion of health checks
+var _ = Describe("Health Check Test", func() {
```
mikenairn, line 24:
```suggestion
var _ = Describe("Health Check Test", Serial, func() {
```

If intend to use ENV vars we can never run this in parallel see https://onsi.github.io/ginkgo/#mental-model-spec-decorators

test/e2e/healthcheck_test.go
```diff
@@ -0,0 +1,196 @@
...
+			geoCode := "EU"
+			if testDNSProvider == "gcp" {
+				geoCode = "europe-west1"
+			}
+			Expect(os.Setenv("TEST_DNS_GEO_CODE", geoCode)).To(BeNil())
+		})
+		AfterEach(func(ctx SpecContext) {
+			Expect(os.Unsetenv("TEST_DNS_GEO_CODE")).To(BeNil())
+		})
+		FIt("creates health checks for a health check spec", func(ctx SpecContext) {
```
mikenairn, line 36: focus needs removed. fyi, you can ensure this by running:
```
./bin/ginkgo unfocus
```

test/e2e/healthcheck_test.go
```diff
@@ -0,0 +1,185 @@
...
+				healthChecksSupported = true
+			}
+
+			provider, err := providerForManagedZone(ctx, testManagedZone)
+			Expect(err).To(BeNil())
+
+			By("creating a DNS Record")
+			dnsRecord := &v1alpha1.DNSRecord{}
+			err = helpers.ResourceFromFile("./fixtures/healthcheck_test/geo-dnsrecord-healthchecks.yaml", dnsRecord, helpers.GetTestEnv)
+			Expect(err).ToNot(HaveOccurred())
```
mikenairn, line 38: You are creating this record but not ensuring it gets removed on failure. see https://github.com/Kuadrant/dns-operator/blob/main/test/e2e/single_record_test.go#L48-L52

mikenairn: Another issue here is that we aren't creating a suite or test specific host to use see https://github.com/Kuadrant/dns-operator/blob/main/test/e2e/single_record_test.go#L36-L38

This is important as it allows the e2e tests to be executed with a set of unique hosts that won't conflict with each other. As this is we wouldn't be able to have the e2e test workflow execute on more than one PR at a time. Also if it fails to cleanup it will mess up the next run.

philbrookes: I was wondering if I needed to use those name generators or not, cheers. I've updated it now.

mikenairn:
@philbrookes This doesn't appear to be creating any health checks, tests pass even when there are errors:

```.json
{
	"lastTransitionTime": "2024-05-14T10:44:45Z",
	"message": "probe (id: , address: 172.32.200.1, host: t-single-proud-glitter.dns-op-e2e-dawn-sun.mn.hcpapps.net) error from DNS Provider: A different health check has already been created with the specified caller reference. If that health check has been deleted, Route 53 will still maintain the caller reference for a number of days.",
	"reason": "DNSProviderError",
	"status": "False",
	"type": "ProbeSynced"
}
```

test/e2e/healthcheck_test.go
```diff
@@ -0,0 +1,210 @@
...
+	var testID string
+	// testDomainName generated domain for this test e.g. t-e2e-12345.e2e.hcpapps.net
+	var testDomainName string
+	// testHostname generated hostname for this test e.g. t-gw-mgc-12345.t-e2e-12345.e2e.hcpapps.net
+	var testHostname string
+
+	var dnsRecord *v1alpha1.DNSRecord
+
+	BeforeEach(func() {
+		testID = "t-single-" + GenerateName()
```
mikenairn, line 34:
```suggestion
		testID = "t-health-" + GenerateName()
```

test/e2e/suite_test.go
```diff
@@ -89,6 +93,16 @@ var _ = BeforeSuite(func(ctx SpecContext) {
...
+
+	geoCode := "EU"
+	if testDNSProvider == "gcp" {
+		geoCode = "europe-west1"
+	}
+	helpers.SetTestEnv("testGeoCode", geoCode)
+})
+
+var _ = AfterSuite(func() {
+	helpers.ResetTestEnvVars()
```
mikenairn, line 105: Not sure there's much point in this? Would only be needed if you were actually setting environment variables that needed cleaned up.

test/e2e/healthcheck_test.go
```diff
@@ -0,0 +1,215 @@
...
+				g.Expect(client.IgnoreNotFound(err)).To(BeNil())
+
+				err = k8sClient.Get(ctx, client.ObjectKeyFromObject(dnsRecord), dnsRecord)
+				g.Expect(errors.IsNotFound(err)).Should(BeTrue())
+			}, TestTimeoutMedium, time.Second).Should(Succeed())
+
+			By("confirming the health checks were removed in the provider")
+			Eventually(func(g Gomega) {
+				for _, healthCheck := range oldHealthCheckStatus.Probes {
+					exists, err := provider.HealthCheckReconciler().HealthCheckExists(ctx, &healthCheck)
```
mikenairn, line 207: What was wrong with that it was passing when the health didn't exist?

mikenairn (approved):
Tried it locally, it does create health checks now.

Question about the `HealthCheckExists` function, does it even work? but if it is, changes look fine to me.

test/e2e/healthcheck_test.go
philbrookes: it jumped straight over to the status check, and I hadn't added the following:
```
g.Expect(len(dnsRecord.Status.HealthCheck.Probes)).ToNot(BeZero())
```

So it looped over all 0 probes and thought it was OK.

philbrookes: the problem wasn't in the `healthcheckExists though, it was [here](https://github.com/Kuadrant/dns-operator/pull/113/files#diff-8e289e1d78e0d1d3a21094f8373fad5460cae17ca8f70d2ce7de1721e2d7cc00R76) and [here](https://github.com/Kuadrant/dns-operator/pull/113/files#diff-8e289e1d78e0d1d3a21094f8373fad5460cae17ca8f70d2ce7de1721e2d7cc00R155)

mikenairn: ok, so it never executed this before?

mikenairn: yeah ok, since you squashed it was hard to tell what you changed to make it work. Get it now.

PR #114, opened 2024-05-10
fix for status setting and ensuring we requeue on errors

### LINKED ISSUE #111: Bug: When AWS secret added after the zone the zone does not get into a ready state
**Steps**

Create a ManagedZone with no AWS secret present in the namespace
Create a secret afterwards
observe ManagedZone stays not ready with provider error

**Options**

- we should re-queue on this error

maleck13:
Fixes #111 

**Validation**

- create a managed zone with a provider secret ref that doesn't exist
- check status of zone
- create secret wait for reconcile to happen (depends on how long it has been failing as will back off)
- check status of zone

### DIFF
--- a/internal/controller/managedzone_controller.go
+++ b/internal/controller/managedzone_controller.go
@@ -18,6 +18,7 @@ package controller
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 
@@ -37,6 +38,10 @@ import (
 
 const (
 	ManagedZoneFinalizer = "kuadrant.io/managed-zone"
+)
+
+var (
+	ErrProvider = errors.New("ProviderError")
 )
 
 // ManagedZoneReconciler reconciles a ManagedZone object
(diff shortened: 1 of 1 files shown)


internal/controller/managedzone_controller.go
```diff
@@ -106,40 +106,44 @@ func (r *ManagedZoneReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	// Publish the managed zone
 	err = r.publishManagedZone(ctx, managedZone)
 	if err != nil {
-		status = metav1.ConditionFalse
-		reason = "ProviderError"
-		message = fmt.Sprintf("The DNS provider failed to ensure the managed zone: %v", err)
-
-		err = r.Status().Update(ctx, managedZone)
```
maleck13, line 113: this wasn't doing anything before

internal/controller/managedzones.go
```diff
@@ -0,0 +1,50 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/kuadrant/dns-operator/api/v1alpha1"
+	"github.com/kuadrant/dns-operator/internal/provider"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func (r *ManagedZoneReconciler) publishManagedZone(ctx context.Context, managedZone *v1alpha1.ManagedZone) error {
```
maleck13, line 12: I kinda want to put these on their own type so they have no access to the client

internal/controller/suite_test.go
```diff
@@ -65,6 +64,10 @@ var dnsProviderFactory = &providerFake.Factory{
 	},
 }
 
+func SetDNSProvider(p provider.Provider) {
+	dnsProvider = p
+}
```
mikenairn, line 69: I think we should avoid changing suite variables in tests. You won't be able to run tests in parallel if you do this as you will be modifying the provider that all specs will get back through the factory.

internal/controller/managedzone_controller_test.go
```diff
@@ -100,5 +109,56 @@ var _ = Describe("ManagedZoneReconciler", func() {
...
+				if meta.IsStatusConditionTrue(mz.Status.Conditions, string(v1alpha1.ConditionTypeReady)) {
+					return fmt.Errorf("expected the ready condition to be false")
+				}
+				if mz.Status.NameServers != nil {
+					return fmt.Errorf("expected nameservers not to be set on provider error")
+				}
+				if mz.Status.RecordCount != 0 {
+					return fmt.Errorf("expected record count to be 0")
+				}
+				return nil
```
mikenairn, line 145: You might want to consider using gomega assertions here to be more consistent with all the other tests:

```.go
Eventually(func(g Gomega) {
	err = k8sClient.Get(ctx, client.ObjectKey{Namespace: mz.Namespace, Name: mz.Name}, mz)
	g.Expect(err).NotTo(HaveOccurred())
	g.Expect(mz.Status.Conditions).To(
		ContainElement(MatchFields(IgnoreExtras, Fields{
			"Type":               Equal(string(v1alpha1.ConditionTypeReady)),
			"Status":             Equal(metav1.ConditionTrue),
			"Reason":             Equal("Whatever this is meant to be"),
			"Message":            Equal("Whatever this is meant to be"),
			"ObservedGeneration": Equal(mz.Generation),
		})),
	)
	g.Expect(mz.Finalizers).To(ContainElement(ManagedZoneFinalizer))
	g.Expect(mz.Status.NameServers).To(BeNil())
	g.Expect(mz.Status.RecordCount).To(BeZero())
}, TestTimeoutMedium, time.Second).Should(Succeed())
```

internal/controller/suite_test.go
maleck13: Yeah just not really sure what other option we have

internal/controller/managedzone_controller.go
```diff
@@ -186,21 +204,21 @@ func (r *ManagedZoneReconciler) deleteManagedZone(ctx context.Context, managedZo
...
 	if err != nil {
 		if strings.Contains(err.Error(), "was not found") || strings.Contains(err.Error(), "notFound") {
 			log.Log.Info("ManagedZone was not found, continuing", "managedZone", managedZone.Name)
 			return nil
 		}
-		return err
+		return ZoneErr{
+			code:    ZoneErrProvider,
+			message: fmt.Sprintf("Failed to delete from provider. Provider Error: %v", provider.SanitizeError(err)),
+		}
```
mikenairn, line 221: Is this preferable to just wrapping an error with a known type?

```.go
var (
   ErrProvider = errors.New("ProviderError")
)
...
if err != nil {
   return fmt.Errorf("%w, failed to get provider for the zone: %v", ErrProvider, provider.SanitizeError(err))
}
```

and using errors.Is where you want to check the type:

```go
if errors.Is(ErrProvider, err) {
   reason = ErrProvider.Error()
   message = err.Error()
}
```

internal/controller/suite_test.go
maleck13: We need a way to simulate errors from the provider IMO. Perhaps we create a follow up to add some form of error injection into the inmem provider?
Like if a test could set a response keyed to itself that would work

internal/controller/managedzone_controller_test.go
```diff
@@ -100,5 +110,59 @@ var _ = Describe("ManagedZoneReconciler", func() {
 			Expect(err).To(HaveOccurred())
 			Expect(err.Error()).To(ContainSubstring("spec.domainName in body should match"))
 		})
+
+		It("status should move from false to true when provider error resolved", func() {
```
maleck13, line 114: @mikenairn I am not sure how great these tests are. If we would prefer I can remove them.

maleck13:
@mikenairn if no objections will I go ahead and merge?

internal/controller/suite_test.go
```diff
@@ -59,6 +59,10 @@ var testEnv *envtest.Environment
 var ctx context.Context
 var cancel context.CancelFunc
 
+func SetDNSProvider(p provider.Provider) {
+	dnsProvider = p
+}
+
```
mikenairn, line 65: Think this can be removed

mikenairn (approved):

mikenairn (approved):

PR #115, opened 2024-05-10
update controller-gen

philbrookes:


maleck13:
does this need to match the version in kuadrant-operator? If not /lgtm

philbrookes:
> does this need to match the version in kuadrant-operator? If not /lgtm

no, no as far as I know.

philbrookes:
> > does this need to match the version in kuadrant-operator? If not /lgtm
> 
> no, no as far as I know.

actually, this is bringing it in line with [kuadrant-operator](https://github.com/Kuadrant/kuadrant-operator/blob/main/Makefile#L196) anyway.

PR #116, opened 2024-05-13
Update inmemory provider

mikenairn:
Update the inmemory provider to work more like aws and google providers when creating a managed zone. If an ID is specified in the spec it is expected that the zone should exist and return an error otherwise. Previously created managed zones should be retrieved using the ID in the status of the managed zone.

Update integration tests to use updated inmemory provider, ensure managedzone being created has no ID in the spec. Also updates the tests to load the provider via the factory instead of overriding the factory in the suite test.

Add a test that checks we get the expected provider error when trying to use a zone id that does not exist.

### DIFF
--- a/internal/controller/managedzone_controller_test.go
+++ b/internal/controller/managedzone_controller_test.go
@@ -19,11 +19,13 @@ limitations under the License.
 package controller
 
 import (
-	"context"
+	"time"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	. "github.com/onsi/gomega/gstruct"
 
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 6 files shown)


internal/controller/managedzone_controller_test.go
```diff
@@ -100,5 +105,68 @@ var _ = Describe("ManagedZoneReconciler", func() {
 			Expect(err).To(HaveOccurred())
 			Expect(err.Error()).To(ContainSubstring("spec.domainName in body should match"))
 		})
+
+		It("managed zone should not become ready with a spec.ID that does not exist", func() {
+			//Create a managed zone with no spec.ID
+			Expect(k8sClient.Create(ctx, managedZone)).To(Succeed())
+			Eventually(func(g Gomega) {
```
maleck13, line 112: ok so this is covering if a zone has no id it will get created 👍

maleck13 (approved):

PR #126, opened 2024-05-15
Add verify generate and gomod

mikenairn:
[Add verify generate and go.mod make targets](https://github.com/Kuadrant/dns-operator/pull/126/commits/9877b7de4626d747ff1b085cd250ded82ef8d0c7)

Add new make targets to verify that `make generate` and `go mod tidy` have been executed against the current code and the results committed.
Add verify-generate and verify-go-mod jobs to CI github workflow.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -98,6 +98,38 @@ jobs:
       - name: Run make verify-imports
         run: |
           make verify-imports
+  verify-generate:
+    name: Verify generate
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up Go 1.21.x
+        uses: actions/setup-go@v5
+        with:
+          go-version: 1.21.x
+          cache: false
+        id: go
+      - name: Check out code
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -98,6 +98,38 @@ jobs:
       - name: Run make verify-imports
         run: |
           make verify-imports
+  verify-generate:
```
pehala, line 101: I see you have a `verify-all` target, could all of these steps be replaced with just running that?

mikenairn: Meh, maybe, maybe not, tbh that only got added and will exit on the first error so wouldn't be ideal. Also i think it it is preferable to see all these as separate checks on the PR:

image

pehala: You could use `make -k verify-all` to mitigate that. I see a few small advantages of doing it that way (quicker run, no need to update actions if check is added) but it is just a preference thing

philbrookes (approved):

PR #135, opened 2024-05-20
add watch and ns config

### LINKED ISSUE #82: managed-zone events not triggering DNS Record reconciliation
When a managed zone is updated, created or deleted any DNS Records that reference it, should reconcile.

maleck13:
fixes #82 

- Add a watch on ManagedZones in the DNSRecord controller
- Add watch namespace env var to allow specifying which namespace to watch

**Verification**

- set ```WATCH_NAMESPACES=dnstest```
- start controller
- create zone in default ns (note it is not reconciled)
- create zone in dnstest (it is reconciled)
- create dnsrecord in dnstest (it is reconciled)
- add label to managedzone see dnsrecord is reconciled when MZ is updated

### DIFF
--- a/internal/controller/dnsrecord_controller.go
+++ b/internal/controller/dnsrecord_controller.go
@@ -32,6 +32,7 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	externaldnsendpoint "sigs.k8s.io/external-dns/endpoint"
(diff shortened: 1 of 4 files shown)


philbrookes:
> Does this mean that DNSOperator will only reconcile ManagedZones in one namespace, i.e. you can no longer have DNSPolicies in multiple namespace?

this change is only intended for use in tests. As I understand it, this will not be used in production.

maleck13:
@pehala by default it will still watch all namespaces. This is so we can install many dnsoperators in a single cluster and have a large scale test without needing hundreds of clusters

philbrookes (approved):

maleck13:
I have come up with a test for proving dns records are re-reconciled when the zone is changed. So will add that before merging

maleck13:
@philbrookes @mikenairn added a test to that. Starts with broken mz and good dnsrecord. Fix mz and assert dnsrecord also reconciled into correct state

internal/controller/dnsrecord_controller.go
```diff
@@ -186,6 +187,23 @@ func (r *DNSRecordReconciler) SetupWithManager(mgr ctrl.Manager, requeueIn, vali
...
+				return toReconcile
+			}
+			for _, record := range records.Items {
+				if record.Spec.ManagedZoneRef.Name == o.GetName() {
+					logger.Info("managed zone updated", "managedzone", o.GetNamespace()+"/"+o.GetName(), "enqueuing dnsrecord ", record.GetName())
+					toReconcile = append(toReconcile, reconcile.Request{NamespacedName: client.ObjectKeyFromObject(&record)})
+				}
+			}
+			return toReconcile
+		})).
```
mikenairn, line 206: Would setting an ownerRef on the DNSRecord and using `Owns` achieve the same?

maleck13: Sounds reasonable 👍

PR #140, opened 2024-05-27
prepare-release: release-0.3

philbrookes:


bundle/manifests/dns-operator.clusterserviceversion.yaml
```diff
@@ -55,14 +55,14 @@ metadata:
...
-    createdAt: "2024-05-10T10:23:31Z"
+    containerImage: quay.io/kuadrant/dns-operator:release-0.3
+    createdAt: "2024-05-27T10:26:10Z"
     description: A Kubernetes Operator to manage the lifecycle of DNS resources
     operators.operatorframework.io/builder: operator-sdk-v1.33.0
     operators.operatorframework.io/project_layout: go.kubebuilder.io/v4
     repository: https://github.com/kuadrant/dns-operator
     support: kuadrant
-  name: dns-operator.v0.0.0
+  name: dns-operator.v0.3.0-dev
```
pehala, line 65: why the -dev suffix? I think we do not use it anywhere else

mikenairn: The dev suffix is intended to reflect the fact it's not a final release yet as per https://github.com/Kuadrant/dns-operator/blob/main/docs/RELEASE.md#new-majorminor-version

If you have no intention of doing any bug fixes or testing prior to publishing the final release you can skip to step 6 of that doc and create the final version.

mikenairn:
We don't want to merge this back to main, it should be getting pushed to a new release branch

bundle/manifests/dns-operator.clusterserviceversion.yaml
pehala: Ahh ok, so basically `-dev` equals `-rc`. Thanks for explaining

PR #142, opened 2024-05-28
feat: Add providers startup flag

### LINKED ISSUE #105: Add option for supported dns providers
**What**

Add an option to the dns operator startup flags that allows the configuration of what dns providers are supported (e.g. --supported-providers=aws,gcp). This should restrict what providers will be allowed to be loaded by the [provider factory](https://github.com/Kuadrant/dns-operator/blob/main/internal/provider/factory.go#L79).

Default to not include testing providers.

**Why**

Some providers probably don't want to be enabled in all environments, [inmemory](https://github.com/Kuadrant/dns-operator/pull/103) for example should probably be disabled by default and only enabled in the dev/test environments. It would also allow us to gradually introduce new providers such as Azure.

mikenairn:
closes #105 

Add a new flag to the controller startup that determines what providers are enabled, if none are specified the list of registered default providers is used.

```
$ go run ./cmd/main.go --provider google,inmemory --provider aws
2024-05-28T17:29:15+01:00       INFO    setup   init provider factory   {"providers": ["google","inmemory","aws"]}
2024-05-28T17:29:15+01:00       INFO    setup   starting manager
....

$ go run ./cmd/main.go --provider aws
2024-05-28T17:32:28+01:00       INFO    setup   init provider factory   {"providers": ["aws"]}
2024-05-28T17:32:28+01:00       INFO    setup   starting manager
....

$ go run ./cmd/main.go
2024-05-28T17:32:48+01:00       INFO    setup   init provider factory   {"providers": ["google","aws"]}
2024-05-28T17:32:48+01:00       INFO    setup   starting manager
....
```

The provider factory is updated to take a list of providers that are enabled and ensures that only providers in this list can be loaded via the factory methods. Providers themselves optionally register themselves as "default" providers, in the case where no providers are specified these providers are enabled in the factory.


internal/provider/inmemory/inmemory.go
```diff
@@ -91,5 +91,7 @@ func (i InMemoryDNSProvider) ProviderSpecific() provider.ProviderSpecificLabels
 // Register this Provider with the provider factory
 func init() {
 	client = inmemory.NewInMemoryClient()
-	provider.RegisterProvider("inmemory", NewProviderFromSecret)
+	//ToDo mnairn: This is temporarily being registered as a default to allow time for compatibility changes to be made
+	// to the kuadrant operator dev/test deployment.
+	provider.RegisterProvider("inmemory", NewProviderFromSecret, true)
```
philbrookes, line 96: so once this merges, we can set the startup value in kuadrant to include `inmemory`, and then change this to false?

philbrookes (approved):
looks good, just want to make sure I've understood why inmemory is set to a default provider.

internal/provider/inmemory/inmemory.go
mikenairn: Exactly. If i set this to false now, which is what we really want it to be, it will potentially break the inmemory provider in the[ kuadrant operator tests ](https://github.com/Kuadrant/kuadrant-operator/pull/658) until it's updated. That said, that's not merged yet :thinking: , i could just make those changes as part of that PR after this is merged.

mikenairn: Change of plan, will merge this first as we want it and update the kuadrant operator PR accordingly. Set this to false now, default is only "aws" and "google" now.

Issue #118, opened 2024-05-14
Add a Grafana Dashboard focused around engineering concerns


maleck13:
**What**

Add a specific dashboard that shows important information from an engineering (platform and us) perspective around what the DNS operator is doing. 

- Requests to external DNS providers (success, ratelimited etc)
- State of a given dns record (IE is it ready/ still trying to be validated/ is it in a conflicted state etc)
- Number of reconciles for a given DNSRecord or Managed zone 
- Total number of DNSRecords and Managed Zones 

Other important things

Issue #117, opened 2024-05-14
Add audit logs for actions taken against a given zone


maleck13:
**What**

With multiple DNS controllers and even with just a single one, in order to debug and be able to understand what happened with a given zone we should ensure we some structured logging in place that uses the 
zoneid, 
DNSName
the cluster / owner id, 
the DNSPolicy (name and namespace that triggered the change (poss has to be done in kuadrant operator or could be covered by ownerid?). There could be optional meta data added by the DNSPolicy controller that if present should make it into logs?
The event that triggered it, the action taken and the resulting response. 

Ideally the outcome would be a set of logs that could be sent to a centralised logging system and then interrogated to understand what happened to a given zone by any controller instance that acted on it 

**use case**

I want to be able to diagnose and understand the set of actions taken in my DNS provider by the Kuadrant dns operator for a given zone or dns name. I want to be able to filter down in the logs to just actions taken against the dns provider that would have impacted the records in a given zone.

**Done**

- [ ] Log metadata defined and documented
- [ ] Logging reviewed in the controllers and providers updated 
- [ ] Logs tested in an aggregator for ability to filter down and find useful information
- [ ] Have a think about how to ensure logs have correct meta data (key values) (https://grafana.com/docs/loki/latest/query/logcli/)

maleck13:
@makslion I believe this is in review now?

makslion:
> @makslion I believe this is in review now?

I still consider it to be WIP. I'll move it to the "in review" once I'll be confident in changes. Right now I'm just looking for a feedback

philbrookes:
https://github.com/Kuadrant/dns-operator/pull/197

PR #152, opened 2024-06-04
chore: Consistent/Improved logging

### LINKED ISSUE #124: Update logging of external-dns code
**What**

We currently have code copied and modified from external-dns which uses the logrus logger. We don't really want to change this since we hope to eventually remove it , but we do need all the logs it produces to go through the same logger as everything else in the dns operator.

**How**

Use a hook and catch all logrus logs and pass them to the dns operator logs?

mikenairn:
closes #124 

[external-dns: Remove logrus logger](https://github.com/Kuadrant/dns-operator/commit/55bf01a274f7e927afba9d03a9fbada88d8f8050)

Removes the use of logrus in all external-dns code in favour of setting a `logr.Logger` on all objects.

Constructors are used throughout to create new instances of components (plan, registry and providers), with each requiring the current context,  retrieving a logger instance from it with an appropriate logger name added where required.

[chore: Consistent/Improved logging](https://github.com/Kuadrant/dns-operator/commit/c4d27a924e6885a9d376239ec1bcdd759d781713)

Construct logger in `Reconcile` functions (DNSRecord and ManagedZone), naming each appropriately, and ensure it is used by all logging statements throughout the call stack for the request.

Initialise external-dns components (registry, plan and providers) using correct context to ensure the appropriate logger is used internally. Removes all references to logrus logger.


internal/controller/managedzone_controller.go
```diff
@@ -156,7 +157,7 @@ func (r *ManagedZoneReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	if err != nil {
 		return ctrl.Result{}, err
 	}
-	log.Log.Info("Reconciled ManagedZone", "managedZone", managedZone.Name)
+	logger.Info("Reconciled ManagedZone")
```
philbrookes, line 160: why don't we need the managedzone name here, but we do need it in some of the others?

philbrookes: like here: https://github.com/Kuadrant/dns-operator/pull/152/files#diff-5882be9025c1bbbef34525f6ffba174154eaac4c288cd24fb8ebb968a287bf84R220

mikenairn: We don't will update those to remove the duplicate key/values

internal/controller/dnsrecord_controller.go
```diff
@@ -271,15 +282,15 @@ func (r *DNSRecordReconciler) deleteRecord(ctx context.Context, dnsRecord *v1alp
...
 			return false, nil
 		} else if strings.Contains(err.Error(), "no endpoints") {
-			logger.Info("DNS record had no endpoint, continuing", "dnsRecord", dnsRecord.Name, "managedZone", managedZone.Name)
+			logger.Info("DNS record had no endpoint, continuing", "managedZone", managedZone.Name)
 			return false, nil
 		}
 		return false, err
 	}
-	logger.Info("Deleted DNSRecord in manage zone", "dnsRecord", dnsRecord.Name, "managedZone", managedZone.Name)
+	logger.Info("Deleted DNSRecord in manage zone", "managedZone", managedZone.Name)
```
mikenairn, line 293: Example:

```.json
{
  "level": "info",
  "ts": "2024-06-04T16:32:37+01:00",
  "logger": "dnsrecord_controller",
  "msg": "Deleted DNSRecord in manage zone",
  "controller": "dnsrecord",
  "controllerGroup": "kuadrant.io",
  "controllerKind": "DNSRecord",
  "DNSRecord": {
    "name": "t-single-lively-waterfall",
    "namespace": "dnstest"
  },
  "namespace": "dnstest",
  "name": "t-single-lively-waterfall",
  "reconcileID": "81a1838b-c782-4323-8baf-58486cadbcec",
  "managedZone": "dev-mz-aws"
}
```

Left `managedZone` here for now although i expect that will be moved and set on the context higher up the stack as part of https://github.com/Kuadrant/dns-operator/issues/117

internal/controller/managedzone_controller.go
mikenairn: Should all be updated now to remove adding the details of the current resource at individual log statements. That is already being done as part of the logger setup in `Reconcile`. Further improvements to the logging and what metadata is included and where will be done as part of  https://github.com/Kuadrant/dns-operator/issues/118

Issue #106, opened 2024-05-02
harden DNSRecord clean up when namespaces deleted


maleck13:
**What**

Sometimes it seems possible for the `ManagedZone` to get deleted before the DNSRecords are removed. Seems we need to make sure the DNSRecords are removed before removing any finalizers on other objects (secrets, zones etc).

review event trigger relationships between all objects, (e.g. if mz secret is changed reconcile mz and related dns records).

Objects to consider:
- secret for managed zone
- managed zone
- dns policies
- dns records

If DNS Record or DNS Policy is deleting and the secret is absent, the DNS Record or DNS Policy should not have it's finalizer removed. 

DNS Record finalizers should only be removed if the controller has confirmed removal of those records from the provider.

If a DNS Record or Policy reconciles and the secret is absent, take actions:
- A clear log entry
- Emit a metric
- Investigate - Can we update status on namespace in this scenario.

philbrookes:
PR: https://github.com/Kuadrant/dns-operator/pull/156

PR #156, opened 2024-06-05
NS Deletion hardening

philbrookes:
closes: https://github.com/Kuadrant/dns-operator/issues/106

Moved some logic around as we were pulling the managed zone from the API in a few places, instead I've just pulled it in the main reconcile and passed it around.  This also meant that I could verify that the MZ and it's secret existed before moving on to setting finalizers and owner references.

In other words, if a DNS Record is created and it's managed zone (or that managed zones secret) has never existed, it will not get the cleanup finalizer, this makes more sense to me, as this record will never have sent anything to a DNS Provider.

I had to modify a test to reflect this more accurately.

New watches:
- Managed zone events triggers DNS Record reconcile
- Secret triggers relevant (if any) managed zone reconcile
- Secret triggers relevant (if any) DNS Record reconcile. 

New metric:
- dns_provider_secret_absent: 
  - 1 when absent
  - 0 when present.


internal/controller/dnsrecord_controller.go
```diff
@@ -305,10 +305,22 @@ func (r *DNSRecordReconciler) publishRecord(ctx context.Context, dnsRecord *v1al
...
 		},
 	}
+
 	err := r.Get(ctx, client.ObjectKeyFromObject(managedZone), managedZone, &client.GetOptions{})
 	if err != nil {
 		return false, err
 	}
+
+	patchFrom := client.MergeFrom(dnsRecord.DeepCopy())
+	err = controllerutil.SetOwnerReference(managedZone, dnsRecord, r.Scheme)
```
mikenairn, line 315: :+1: , you changed it before i got to comment. Is there no helper available  for checking owns? Also, this doesn't add the block on delete flag, do we care?

internal/controller/dnsrecord_controller.go
```diff
@@ -305,10 +305,22 @@ func (r *DNSRecordReconciler) publishRecord(ctx context.Context, dnsRecord *v1al
...
+	if err != nil {
+		return false, err
+	}
+	err = r.Client.Patch(ctx, dnsRecord, patchFrom)
+	if err != nil {
+		return false, err
+	}
```
mikenairn, line 322: Why patch instead of update?

internal/controller/dnsrecord_controller.go
```diff
@@ -270,8 +271,7 @@ func (r *DNSRecordReconciler) deleteRecord(ctx context.Context, dnsRecord *v1alp
 	}
```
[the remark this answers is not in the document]
mikenairn: If you do this you probably don't need that owns helper since it's not used anywhere else that i can see.

internal/controller/dnsrecord_controller.go
philbrookes: yeah, thanks for the hint, haha! I found this: https://github.com/redhat-cop/operator-utils/blob/26a9d4efe6c40ed22f302342196ea6afc8b63170/pkg/util/owner.go#L24-L35 shall I add it as an import?

internal/controller/dnsrecord_controller.go
philbrookes: I assumed this would help avoid a conflict when the controller tried to update the status later, is that not true?

mikenairn: I just realised this is happening way down in `publishRecord`. If we are setting up a relationship between the DNSRecord and the ManagedZone, and this reconciler is the thing ensuring it, i would have expected it be done done at the very start of the Reconcile, with a more explicit step much like when we add the [finalizer](https://github.com/Kuadrant/dns-operator/blob/main/internal/controller/dnsrecord_controller.go#L126).

Issue we have at the moment is that we don't load the MZ until we get down here, but i think we need to change that anyway re. logging [see](https://github.com/Kuadrant/dns-operator/pull/152#discussion_r1626235920)

Generally i think the `Reconcile` function should probably first ensure all bits that require an update to the resource are done and then move on to the actual reconciliation of the spec once that is complete.

1. * Add resource data this reconciler is ensuring (OwnerRefs, finalizers etc..)
2. * Update and return for an immediate re-queue
3. * Process the spec of the resource
4. * Update the resource status and return with validation re-queue

We can leave it like this if we don't want to change that here, but i would expect that to be changed fairly soon after as part of follow on work.

internal/controller/dnsrecord_controller_test.go
```diff
@@ -199,13 +199,52 @@ var _ = Describe("DNSRecordReconciler", func() {
...
 		Eventually(func(g Gomega, ctx context.Context) {
 			err := k8sClient.Get(ctx, client.ObjectKeyFromObject(dnsRecord), dnsRecord)
 			g.Expect(err).To(HaveOccurred())
 			g.Expect(err).To(MatchError(ContainSubstring("not found")))
 		}, 5*time.Second, time.Second, ctx).Should(Succeed())
 	})
-
 	It("should have ready condition with status true", func() {
 		Expect(k8sClient.Create(ctx, dnsRecord)).To(Succeed())
 		Eventually(func(g Gomega) {
```
mikenairn, line 250: Tests should probably be updated to add an additional assertion on the owner ref on the records being added  correctly since this reconciler is the thing ensuring it.

mikenairn:
Changes here look fine, but does this cover everything that was being asked for from https://github.com/Kuadrant/dns-operator/issues/106?

The change to setting an `ownerRef` instead of using the watch seems like a good one, but doesn't really change the behaviour in any way, the existing Watch was doing the same thing to ensure the record reconcile was triggered on MZ changes. What about the MZ secrets if they are modified or deleted? Do they not need to be taken into account somewhere?

philbrookes:
> Changes here look fine, but does this cover everything that was being asked for from #106?
> 
> The change to setting an `ownerRef` instead of using the watch seems like a good one, but doesn't really change the behaviour in any way, the existing Watch was doing the same thing to ensure the record reconcile was triggered on MZ changes. What about the MZ secrets if they are modified or deleted? Do they not need to be taken into account somewhere?

hmm, yeah, that seems out of scope for the title of that issue, but it is mentioned in the body alright and I'd overlooked it. I'll make some changes to cover the extra areas.

internal/controller/dnsrecord_controller.go
```diff
@@ -237,18 +271,34 @@ func (r *DNSRecordReconciler) SetupWithManager(mgr ctrl.Manager, maxRequeue, val
...
+				}
+				// if secret is related to zone for this record, add this record to requeue items
+				if managedZone.Spec.SecretRef.Name == o.GetName() {
+					logger.Info(
+						"managed zone Secret updated",
+						"managedZone", managedZone.Namespace+"/"+managedZone.Name,
+						"secret", o.GetNamespace()+"/"+o.GetName(),
+						"enqueuing record ", record.GetName())
 					toReconcile = append(toReconcile, reconcile.Request{NamespacedName: client.ObjectKeyFromObject(&record)})
 				}
```
mikenairn, line 303: Why is the watch for a secret on the dnsrecord and not on the managed zone, the thing that needs it?

MZ watches secrets, changes to secret trigger MZ reconcile, changes to the state of the MZ due to the secret will trigger a DNSRecord reconcile since it owns them.

internal/controller/dnsrecord_controller.go
```diff
@@ -96,11 +97,36 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+	// Ensure provider secret exists before continuing
+	err = r.Get(ctx, client.ObjectKey{Namespace: managedZone.Namespace, Name: managedZone.Spec.SecretRef.Name}, &v1.Secret{}, &client.GetOptions{})
+	if err != nil {
+		secretMissing.WithLabelValues(managedZone.Name, managedZone.Namespace, managedZone.Spec.SecretRef.Name).Set(1)
+		reason := "ProviderError"
+		message := fmt.Sprintf("The DNS provider failed to ensure the record: %v", provider.SanitizeError(err))
+		setDNSRecordCondition(dnsRecord, string(v1alpha1.ConditionTypeReady), metav1.ConditionFalse, reason, message)
+		return r.updateStatus(ctx, previous, dnsRecord, false, err)
+	}
+	secretMissing.WithLabelValues(managedZone.Name, managedZone.Namespace, managedZone.Spec.SecretRef.Name).Set(0)
```
mikenairn, line 123: Doesn't feel like this is where this should happen. We might want to check the status of the ManagedZone somewhere to make sure its ready, but i don't think dnsrecord should to be aware of secrets.

internal/controller/dnsrecord_controller.go
mikenairn: I see you have the watch on the ManagedZone https://github.com/Kuadrant/dns-operator/pull/156/files#diff-5882be9025c1bbbef34525f6ffba174154eaac4c288cd24fb8ebb968a287bf84R173-R189, so not sure why we need this one.

philbrookes: Why would the managedzone be updated due to a value in the secret changing?

internal/controller/dnsrecord_controller.go
philbrookes: yes ok, maybe it makes more sense to look for this in the managedzone controller? Seeing as that triggers on secret changes now it should see it immediately anyway.

internal/controller/dnsrecord_controller.go
philbrookes: discussed on a call, will remove this.

internal/controller/managedzone_controller.go
```diff
@@ -105,6 +109,22 @@ func (r *ManagedZoneReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		}
 	}
 
+	// Ensure provider secret exists before continuing
+	err = r.Get(ctx, client.ObjectKey{Namespace: managedZone.Namespace, Name: managedZone.Spec.SecretRef.Name}, &v1.Secret{}, &client.GetOptions{})
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			secretMissing.WithLabelValues(managedZone.Name, managedZone.Namespace, managedZone.Spec.SecretRef.Name).Set(1)
```
mikenairn, line 116: I know metrics was already in here, but it would probably be better in it's own "metrics" package as it really wasn't obvious what this was. One for later maybe as we add more metrics.

internal/controller/managedzone_controller.go
```diff
@@ -105,6 +109,22 @@ func (r *ManagedZoneReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+		}
+		message := fmt.Sprintf("Could not find secret: %v/%v for managedzone: %v/%v", managedZone.Namespace, managedZone.Spec.SecretRef.Name, managedZone.Namespace, managedZone.Name)
+		setManagedZoneCondition(managedZone, string(v1alpha1.ConditionTypeReady), metav1.ConditionFalse, "DNSProviderSecretNotFound", message)
+		statusUpdateErr := r.Status().Update(ctx, managedZone)
+		if statusUpdateErr != nil {
+			return ctrl.Result{}, fmt.Errorf("zone error %v : status update failed error %v", err, statusUpdateErr)
+		}
+		return ctrl.Result{}, err
+	}
+	secretMissing.WithLabelValues(managedZone.Name, managedZone.Namespace, managedZone.Spec.SecretRef.Name).Set(0)
```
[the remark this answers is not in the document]
philbrookes: That link isn't working for me.

internal/controller/dnsrecord_controller.go
```diff
@@ -96,11 +97,25 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
 	dnsRecord := previous.DeepCopy()
 
+	managedZone := &v1alpha1.ManagedZone{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      dnsRecord.Spec.ManagedZoneRef.Name,
+			Namespace: dnsRecord.Namespace,
+		},
+	}
+	err = r.Get(ctx, client.ObjectKeyFromObject(managedZone), managedZone, &client.GetOptions{})
+	if err != nil {
```
eguzki, line 107: what if the managed zone is not found when the dns record is being deleted? The finalizer on the dnsrecord will never be removed

philbrookes: Yes, this is the intention, that means that DNS Records have been created in the DNS Provider and have not been cleaned up.

eguzki: Well, you might be right. I know I do not know. But that is causing an issue anyway. Some tests are failing because the dns record is never gone after having deleted it.

KevFan: @philbrookes Is this the intention if a user deletes a `ManagedZone` first before deleting the `DNSRecords` ? 🤔 This is how the integration tests are in Kuadrant Operator. In the tests, the `ManagedZone` has completed their deletion, but `DNSRecords` remained, and prevented the test NS deletion. 

Seems like this introduces a required order of deletion for the user. So maybe the `ManagedZone` deletion should wait until the assoicated `DNSRecords` are deleted ? 🤔 

If this is the intended behaviour anyway, I've a fix to update the Kuadrant Operator `DNSPolicy` tests as part of a PR https://github.com/Kuadrant/kuadrant-operator/pull/689/commits/c70b5e1fc1e8372988a60aaa327452083a89b764. Just want to check this is the expected flow being merging

philbrookes: Yes, this is the intention, if the DNS Record is deleted after the managed zone, the DNS Operator cannot remove the records from the DNS Provider, in order to highlight this to the user/admin the DNS Record never has it's finalizer removed.

See the original issue here: https://github.com/Kuadrant/dns-operator/issues/106

Issue #110, opened 2024-05-08
Improve validations on DNSRecord spec fields


mikenairn:
**What**

Improve the validations we currently have for fields on the DNSRecord spec. 

Review the DNSRecord spec and check that each field exposed via the public API has appropriate validations added.  
We may want to add a limit to the number of endpoints that can be added to the spec, we need to look into CEL docs to see what sort of upper limits are permitted.

Investigate more advanced CEL validations, can this whole [validate method](https://github.com/Kuadrant/dns-operator/blob/main/api/v1alpha1/dnsrecord_types.go#L175) be replaced with [CEL](https://kubernetes.io/docs/reference/using-api/cel/) validations?

makslion:
Closing this issue. We can't conveniently use CEL validations unless we reimplement the Endpoint struct from external-dns. Reimplementing will allow us to reduce complexity of the CEL validation, otherwise openAPIV3Schema validation fails. Refer to the #158 for more details.

mikenairn:
> Closing this issue. We can't conveniently use CEL validations unless we reimplement the Endpoint struct from external-dns. Reimplementing will allow us to reduce complexity of the CEL validation, otherwise openAPIV3Schema validation fails. Refer to the #158 for more details.

Using CEL to replace the validate function was only part of what we wanted to get out of this issue, and was really a question to answer as part of looking into the validation rather than being the only thing to try and do. 

The main thing was to "Improve the validations we currently have for fields on the DNSRecord spec". There is still a lot for the fields in the spec that have no validations where there probably should be something. HealthCheck spec in particular is quite lacking https://github.com/Kuadrant/dns-operator/blob/main/api/v1alpha1/dnsrecord_types.go#L27-L53

PR #160, opened 2024-06-13
[WIP] GH-110 use CRD validation on DNSRecord

makslion:


api/v1alpha1/dnsrecord_types.go
```diff
@@ -33,10 +33,19 @@ const HttpsProtocol HealthProtocol = "HTTPS"
...
-	FailureThreshold *int            `json:"failureThreshold,omitempty"`
+	// Should start with "?" or "|", contain only allowed for URLs characters and end with alphanumeric char or "/"
+	// +kubebuilder:validation:Pattern=`^(?:\?|\/)[\w\-.~:\/?#\[\]@!$&'()*+,;=]+(?:[a-zA-Z0-9]|\/){1}$`
+	Endpoint string `json:"endpoint,omitempty"`
+
+	// Only following ports allowed: 80, 443, 1024-49151
+	// +kubebuilder:validation:XValidation:rule="self in [80, 443] || (self >= 1024 && self <= 49151)",message="Allow ports 80, 443, 1024-49151 are allowed"
+	Port *int `json:"port,omitempty"`
+
+	// +kubebuilder:validation:XValidation:rule="self in ['HTTP','HTTPS']",message="Only HTTP or HTTPS protocols are allowed"
```
philbrookes, line 44: is this case sensitive? Ideally it wouldn't be.

api/v1alpha1/dnsrecord_types.go
```diff
@@ -62,7 +71,10 @@ type DNSRecordSpec struct {
 
 	// rootHost is the single root for all endpoints in a DNSRecord.
 	// it is expected all defined endpoints are children of or equal to this rootHost
+	// Should contain two groups of allowed for URLs characters separated by a "."
```
philbrookes, line 74: // Should contain two groups of valid URL characters separated by a "."

api/v1alpha1/dnsrecord_types.go
```diff
@@ -33,10 +33,19 @@ const HttpsProtocol HealthProtocol = "HTTPS"
...
+	// Should start with "?" or "|", contain only allowed for URLs characters and end with alphanumeric char or "/"
+	// +kubebuilder:validation:Pattern=`^(?:\?|\/)[\w\-.~:\/?#\[\]@!$&'()*+,;=]+(?:[a-zA-Z0-9]|\/){1}$`
+	Endpoint string `json:"endpoint,omitempty"`
+
+	// Only following ports allowed: 80, 443, 1024-49151
+	// +kubebuilder:validation:XValidation:rule="self in [80, 443] || (self >= 1024 && self <= 49151)",message="Allow ports 80, 443, 1024-49151 are allowed"
+	Port *int `json:"port,omitempty"`
+
+	// +kubebuilder:validation:XValidation:rule="self in ['HTTP','HTTPS']",message="Only HTTP or HTTPS protocols are allowed"
+	Protocol *HealthProtocol `json:"protocol,omitempty"`
```
mikenairn, line 45: Using Enum would probably make more sense here, and for port see https://github.com/Kuadrant/kuadrant-operator/blob/main/api/v1alpha1/dnspolicy_types.go#L255

Also, all the validations should be the same as https://github.com/Kuadrant/kuadrant-operator/blob/main/api/v1alpha1/dnspolicy_types.go#L246-L261

@philbrookes Why is that duplicated? Would it not make more sense for the DNSPolicy to reference this HealthCheckSpec type?

api/v1alpha1/dnsrecord_types.go
```diff
@@ -33,10 +33,19 @@ const HttpsProtocol HealthProtocol = "HTTPS"
 // By default this health check will be applied to each unique DNS A Record for
 // the listeners assigned to the target gateway
 type HealthCheckSpec struct {
-	Endpoint         string          `json:"endpoint,omitempty"`
-	Port             *int            `json:"port,omitempty"`
-	Protocol         *HealthProtocol `json:"protocol,omitempty"`
-	FailureThreshold *int            `json:"failureThreshold,omitempty"`
+	// Should start with "?" or "|", contain only allowed for URLs characters and end with alphanumeric char or "/"
```
philbrookes, line 36: // Must start with "?" or "/", contain only valid URL characters and end with alphanumeric char or "/"

api/v1alpha1/dnsrecord_types.go
```diff
@@ -33,10 +33,19 @@ const HttpsProtocol HealthProtocol = "HTTPS"
...
+	// +kubebuilder:validation:Pattern=`^(?:\?|\/)[\w\-.~:\/?#\[\]@!$&'()*+,;=]+(?:[a-zA-Z0-9]|\/){1}$`
+	Endpoint string `json:"endpoint,omitempty"`
+
+	// Only following ports allowed: 80, 443, 1024-49151
```
philbrookes, line 40: // Port must be either 80, 443 or 1024 and above

philbrookes (commented):
Some small changes to the language used, the regexes look fine.

api/v1alpha1/dnsrecord_types.go
philbrookes: I cannot remember clearly, I suspect the Specs were different from each other at some point, but are aligned now.

philbrookes: as the validations in this PR are superior to the validations in Kuadrant-operator, it might be worth a PR to update kuadrant to use this.

api/v1alpha1/dnsrecord_types.go
makslion: Right now if we will use lower case protocols the dns-operator will break. Well, not break but will try to create healtcheck with `nil` protocol. I can change that to allow for `http` and `https` ports. but that would mean changing the logic a bit :thinking:

api/v1alpha1/dnsrecord_types.go
makslion: I'd prefer the CEL validations - way nicer error message if things aren't right. 

Hmm, would it be a good idea to remove duplication as part of this issue?

philbrookes:
I see @mikenairn has added some [tests](https://github.com/Kuadrant/dns-operator/pull/170/files#diff-1a6ccb5268cdf90ca2f3a2c51e27f975774acc00e8e4ccb565086d1c4a81bf49R379) to cover his validations, which is pretty neat, is that something we could also do here?

api/v1alpha1/dnsrecord_types.go
makslion: I kept the upper limit since it is possible to set a port higher but we won't allow it (the range of unregistered ports is not allowed)

makslion:
Added test @philbrookes. It is a strange one, but, hopefully, is useful. At least it is cheep to check it.

api/v1alpha1/dnsrecord_types.go
makslion: Opened a [PR](https://github.com/Kuadrant/kuadrant-operator/pull/723) to the kuadrant-operator as well

philbrookes (approved):

PR #162, opened 2024-06-14
Owned DNS Records will block Managed Zone deletion

philbrookes:


internal/controller/managedzone_controller.go
```diff
@@ -84,6 +85,15 @@ func (r *ManagedZoneReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
 			return ctrl.Result{}, err
 		}
+
+		if recordsExist, err := r.hasOwnedRecords(ctx, managedZone); err != nil {
+			logger.Error(err, "Failed to check owned records")
+			return ctrl.Result{}, err
+		} else if recordsExist {
+			logger.Info("ManagedZone deletion awaiting removal of owned DNS records")
+			return ctrl.Result{Requeue: true}, nil
+		}
```
mikenairn, line 95: Should this re-queue? Do the owner records not cause this to reconcile on change anyway?

internal/controller/dnsrecord_controller_test.go
```diff
@@ -99,15 +99,15 @@ var _ = Describe("DNSRecordReconciler", func() {
...
 			Expect(client.IgnoreNotFound(err)).ToNot(HaveOccurred())
 		}
 		if managedZone != nil {
-			err := k8sClient.Delete(ctx, managedZone)
+			err := k8sClient.Delete(ctx, managedZone, client.PropagationPolicy(metav1.DeletePropagationForeground))
 			Expect(client.IgnoreNotFound(err)).ToNot(HaveOccurred())
 		}
 		if brokenZone != nil {
-			err := k8sClient.Delete(ctx, brokenZone)
+			err := k8sClient.Delete(ctx, brokenZone, client.PropagationPolicy(metav1.DeletePropagationForeground))
```
mikenairn, line 110: Not sure these make any difference unless you set the owner ref up with block owner deletion https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/#ownership-and-finalizers

mikenairn (approved):
As discussed offline I'm not really sure if there is a lot of point adding this change considering we can't prevent the secrets being removed, but mostly because we want to [remove the ManagedZone API](https://github.com/Kuadrant/dns-operator/issues/166) anyway. That said, it looks fine, so can be merged if desired.

PR #163, opened 2024-06-18
tests: Add multi instance test suite and namespace scoped deployments

### LINKED ISSUE #121: Add basic multi-instance e2e scale test for DNS Operator
**What**

Currently we only test the DNSOperator either in a single cluster or at most in 2 clusters. We need to increase on this and find where there are issues beyond this scale:

Define and implement a test that will  test the DNSOperators capabilities at a larger scale. A goal here would be that you could have something like 

``` make deploy-multi INSTANCES=N ```

That would deploy N instances namespace scoped to the target cluster.

Ideally we would use kustomize but I am not sure it is possible to generate lots of instances with this method

We would then want something like 

```make test-e2e-multi  INSTANCES=N ```

That would use deploy-multi and execute a set of tests against these deployed instances.

**Proposal**

**Note by correct state here I am referring to the DNSRecord and possibly DNSPolicy status and also the provider zone state**

- install multiple instances of the DNS operator into a single cluster that are each looking only at their own namespaces via the WATCH_NAMSPACES env var. 
- Define a set of DNSRecord (based off a DNSPolicy or just install kuadrant and create a DNSPolicy) that sets up common shared dns records for a common dns name and create that DNSRecord in each of the DNS operator namespaces
- Validate that the DNSRecords get into the correct state within a defined time
- Tear down Gateways and DNSPolicies ensure correct state 
- Clean up entirely and ensure correct state 

Beyond a simple smoke test where the we expect the records not to conflict we also should look to cover areas such as:

- Understand where and when we begin to hit rate limits and how this presents itself and if it is clearly recognisable 
- Define conflicting DNSRecords (IE a simple and loadbalanced) define and test the expected result in this case 

Beyond the basic happy path defined we should also add in additional tests for common misconfiguration

- missing credentials
- bad credentials 
- Bad zone definition

### PR DESCRIPTION
**Pull request by mikenairn**
closes #121 

[Update local-setup for namespaced deployments](https://github.com/Kuadrant/dns-operator/pull/163/commits/eb5cb60cc9855d1cce53f3821e074829f8d18018)

Allow local-setup to create x number of namespaced deployments of the dns operator in a single cluster. 

Adds new options to control the behaviour of the deployment:

DEPLOYMENT_SCOPE=cluster|namespace (default is cluster)
DEPLOYMENT_COUNT=2 - Number of instances to deploy (namespace deployment
scope only)

```
make local-setup DEPLOY=true DEPLOYMENT_SCOPE=namespace DEPLOYMENT_COUNT=2
...
kubectl get deployments -l app.kubernetes.io/part-of=dns-operator -A
dns-operator-1   dns-operator-controller-manager-1   1/1     1            1           14s
dns-operator-2   dns-operator-controller-manager-2   1/1     1            1           14s
```

When using the namespace deployment scope, kustomization overlays are generated and used to deploy the desired resources. These could eventually be used against any cluster and not limited to local kind clusters, and can be found in the default overlays dir:

```
tree tmp/overlays/kuadrant-dns-local/
tmp/overlays/kuadrant-dns-local/
├── kustomization.yaml
├── namespace-dns-operator-1
│   ├── dns-operator
│   │   └── kustomization.yaml
│   ├── kustomization.yaml
│   └── managedzones
│       └── kustomization.yaml
└── namespace-dns-operator-2
    ├── dns-operator
    │   └── kustomization.yaml
    ├── kustomization.yaml
    └── managedzones
        └── kustomization.yaml

7 directories, 7 files
```
 

[tests: Add multi instance e2e test suite](https://github.com/Kuadrant/dns-operator/pull/163/commits/b5f64eb05bd1dc3839f6738e75abc020b263d062)

Adds a multi instance test suite intended to test scenarios where multiple instances of the dns operator are running each reconciling DNSRecord resources contributing to the same shared dns zone.
The suite allows runtime configuration to alter the number of instances that are under test allowing stress testing scenarios to be executed using more extreme numbers of instances and records.

In a local development kind setup, the following local-setup and test commands can be used:

Create kind cluster with four dns operator deployments each watching their own namespace, and each containing the default managedzones (dev-mz-aws and dev-mz-gcp)
```
make local-setup DEPLOY=true DEPLOYMENT_SCOPE=namespace DEPLOYMENT_COUNT=4
```

Run the multi instance test suite using the aws zone:
```
make test-e2e-multi TEST_DNS_MANAGED_ZONE_NAME=dev-mz-aws TEST_DNS_NAMESPACES=dns-operator DEPLOYMENT_COUNT=4
```

Notes:

* Currently an issue if you use a large amount of instances, the endpoint don't always get deleted correctly. This is because of the ownership/registry and the way it currently works. We will need to look into changing this in follow on tasks.
* Additional assertions required in the loadbalanced test scenario to ensure the records are created correctly in the provider.
* New test suite is doing a similar test to the [current multi record test](https://github.com/Kuadrant/dns-operator/blob/main/test/e2e/multi_record_test.go) in the normal e2e test suite. When we get to a point where the multi test suite is dong the same assertions and invoked by the GH actions we can probably remove this.
* Follow on task required to improve the overlay generation tasks to allow them to be used more easily outside of local setup.

Follow on issues:

* https://github.com/Kuadrant/dns-operator/issues/176
* https://github.com/Kuadrant/dns-operator/issues/177

### CODE REVIEW

**Comment by maleck13** [comment]
@ficap you may also be interested in this from a QE persepective

File: make/kustomize_overlays.mk [hunk 1]
```diff
@@ -0,0 +1,63 @@
...
+	# Generate managedzones overlay
+	mkdir -p $(CLUSTER_OVERLAY_DIR)/$(CLUSTER_NAME)/namespace-$(DEPLOYMENT_NAMESPACE)/managedzones
+	cd $(CLUSTER_OVERLAY_DIR)/$(CLUSTER_NAME)/namespace-$(DEPLOYMENT_NAMESPACE)/managedzones && \
+	touch kustomization.yaml
+
+	@if [[ -f "config/local-setup/managedzone/gcp/managed-zone-config.env" && -f "config/local-setup/managedzone/gcp/gcp-credentials.env" ]]; then\
+		cd $(CLUSTER_OVERLAY_DIR)/$(CLUSTER_NAME)/namespace-$(DEPLOYMENT_NAMESPACE)/managedzones && \
+		$(KUSTOMIZE) edit add resource "../../../../../config/local-setup/managedzone/gcp" ;\
+	fi
+	@if [[ -f "config/local-setup/managedzone/aws/managed-zone-config.env" && -f "config/local-setup/managedzone/aws/aws-credentials.env" ]]; then\
```
maleck13, line 48: @philbrookes importnt to note this will need to be updated once azure is added

maleck13 (commented):
Changes and initial tests look good to me. @mikenairn do we have the follow ons created , so we don't loos track of the additional work?

maleck13 (approved):

PR #170, opened 2024-06-26
feat: Generate unique OwnerID based on DNSRecord UID

### LINKED ISSUE #109: Use DNSRecord UID as ownerID
**What**

Make `ownerID` optional and immutable from the DNSRecord spec and instead use a shortened version of the DNSRecord UID as the `ownerID`.

phases:
- make it optional and imutable (defaults to UID)
- update kuadrant to not set ownerID

**Why**

The ownership of records is all based on the DNSRecord resources, DNSRecord owns the records it is creating in the provider. Kubernetes already assigns a unique id to all resources, so there seems little need to create another one.

**Questions**

* How short can we make the uid without risking conflicts?
  * 12 chars means a limit of 20 clusters
* How likely are UIDs to conflict across clusters?
* Should `ownerID` be left as optional in the spec?
  * this could allow a workaround to a conflicting set of IDs.
 
** Tests **
Ensure ownerID value is used if present
Ensure UID is used when absent
Ensure ownerID cannot be set later when initially absent
ensure ownerID cannot be changed or removed when initially set

### LINKED ISSUE #165: Bug: allow domain filtering for wildcard DNS Records
## What
Consider the following:
Record 1 with `other.test.example-boop.com` endpoint that supplies `rootDomainFilter` of `other.test.example-boop.com`
Record 2 with `*.example-boop.com` endpoint that supplies `rootDomainFilter` of `example-boop.com`

There could be 6 scenarios:

1. Zone has booth endpoints. We are reconciling record 1. We ignore record 2 endpoint because it does not match  filters.  Remainder matches record 1 so no action. 
2. Zone has booth endpoints. We are reconciling record 2. We find an extra endpoint from record 1 and delete it. <--- this should be not found/filtered out
3. Zone has endpoints of record 1. We are reconciling record 1. No changes 
4. Zone has endpoints of record 1. We are reconciling record 2. Delete record 1 and create record 2 endpoints <--- record 1 eps should be filtered out
5. Zone has endpoints of record 2. We are reconciling record 1. Create record 1 endpoints - (ignore record 2 endpoints - they are filtered out)
6. Zone has endpoints of record 2. We are reconciling record 2. No changes

The way filtering works is we have one of two filters: domainFilter from mz or rootDomainFilter from the record with the preference for the rootDomainFilter.
The matching is done by [this](https://github.com/kubernetes-sigs/external-dns/blob/master/endpoint/domain_filter.go#L101-L123) code. And in order to ignore something we should not match record to the filter.

in scenario 1 where we ignore record 2:
filter:  `other.test.example-boop.com`
and we have records of `other.test.example-boop.com` and `*.example-boop.com` . Match is straight forward.

in scenario 2 where we don't ignore anything:
filter: `example-boop.com`
and we have records of `other.test.example-boop.com` and `*.example-boop.com`. They are booth matched so we delete one.

This causes records to constantly override each other and flip from `ready` to `awaitingValidation` state. 

## How
#109 should fix it 
Before closing this issue ensure that we have a test for this scenario. [Example](https://github.com/Kuadrant/dns-operator/commit/312dbe66cdec05c88b5ae9336c34c3f6497ff2a5) of the test

mikenairn:
closes #109
fixes #165 

Makes the ownerID in the DNSrecord spec optional. If no ownerID is set in the DNSRecord spec on initial creation, the dns record controller itself will generate a unique ownerID value based on the DNSRecord UID. The spec ownerID is still immutable, but now also has additional validations to ensure it can never be changed from a set to unset or unset to set state after initial creation.
Since the spec ownerID is now optional, the ownerID used throughout all operations is stored and retrieved from the status of the DNSRecord.

Adds additional validations to the ownerID field specifying a MinLength and MaxLength as per best practices for CEL validations. MaxLength is currently 36 which allows the full UID value to be used if desirable.


api/v1alpha1/dnsrecord_types.go
```diff
@@ -53,12 +55,16 @@ type HealthCheckStatusProbe struct {
 }
 
 // DNSRecordSpec defines the desired state of DNSRecord
+// +kubebuilder:validation:XValidation:rule="!has(oldSelf.ownerID) || has(self.ownerID)", message="OwnerID can't be unset if it was previously set"
```
philbrookes, line 58: Why isn't this covered by the rule on line 64? 
`// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="OwnerID is immutable"`

mikenairn: This is only checked once it's set see https://kubernetes.io/blog/2022/09/29/enforce-immutability-using-cel/#immutablility-after-first-modification

Also [here](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#transition-rules) and [here](https://opensource.googleblog.com/2023/11/kubernetes-crd-validation-using-cel.html)

Transition Rule | Purpose
-- | --
self == oldSelf | For a required field, make that field immutable once it is set. For an optional field, only allow transitioning from unset to set, or from set to unset.

mikenairn: Lot of messing around trying to get that to work, but if you see a better way... In reality though it doesn't really matter if it gets changed since the controller sets the `ownerID` once and never lets it change again.

philbrookes (approved):

PR #179, opened 2024-07-01
Helm charts

didierofrivia:
This PR introduces a way to manage a DNS Operator [Helm Chart](https://helm.sh/). This is not meant to replace the way we are building and delivering our manifests (Kustomize, OLM) but to provide an alternative (complementary) way of delivering the operator.

This early implementation uses Kustomize to create the chart template, instead of creating and maintaining new ones with Helm, to later customize the Helm only settings with its `values.yaml`

**NOTES**

* This chart (and every other operator chart we package in the future: Limitador, Authorino, etc) bundles only the operator in the same fashion OLM does.
* The operator version matches the chart version
* This first release doesn't use helm templating, it provides only the manifests previously built by Kustomize.
* The sole repository for all our operator charts will be hosted in [Kuadrant/helm-charts](https://github.com/Kuadrant/helm-charts/).
* The "source of truth" (source code and releasing logic) will be provided in each operator repo, not in the helm charts repository above.
* The helm charts repository is [Kuadrant/helm-charts](https://github.com/Kuadrant/helm-charts/)
* Documentation will be provided once the repository and other operator charts are tested.
* The steps to try out the repository will be provided in the kuadrant helm charts repo.
* For the release of the package, it will be tested once merged with alpha pre releases.

**Verification Steps**

1. Create a local kind cluster
```sh
kind create cluster --name kuadrant-local
```

2. Deploy (install) the default (`latest`) DNS Operator chart
```sh
make helm-install
```

3. Verify the installed operator image:
```sh
kubectl get deployments/dns-operator-controller-manager -n dns-operator-system -o yaml | grep image:
```
it should return:
```sh
image: quay.io/kuadrant/dns-operator:latest
```

4. Build new manifests with specific version of the operator
```sh
make helm-build VERSION=0.4.0-alpha1
```

5. Upgrade the current installed operator to the freshly built one
```sh
make helm-upgrade
```

6. Check the installed operator image:
```sh
kubectl get deployments/dns-operator-controller-manager -n dns-operator-system -o yaml | grep image:
```

it should return:
```sh
image: quay.io/kuadrant/dns-operator:v0.4.0-alpha1
```

7. For uninstalling the chart:
```sh
make helm-uninstall
```

### DIFF
--- a/.github/workflows/delete-release-helm-chart.yaml
+++ b/.github/workflows/delete-release-helm-chart.yaml
@@ -1,0 +1,23 @@
+# Description: This workflow is used to communicate Kuadrant helm charts repo that a release has been deleted.
+
+name: Delete Release Helm Chart
+on:
+  release:
+    types:
+      - deleted
+jobs:
+  delete_chart_release:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Parse Tag
+        run: |
+          tag=${{ github.event.release.tag_name || inputs.operatorTag }}
+          echo "OPERATOR_VERSION=${tag#v}" >> $GITHUB_ENV
+      - name: Sync deleted Helm Chart with repository
+        run: |
+          make helm-sync-package-deleted \
+            VERSION=${{env.OPERATOR_VERSION}} \
+            HELM_WORKFLOWS_TOKEN=${{ secrets.HELM_WORKFLOWS_TOKEN }}
--- a/.github/workflows/release-helm-chart.yaml
+++ b/.github/workflows/release-helm-chart.yaml
@@ -1,0 +1,52 @@
+# Description: This workflow is used to release the Helm chart to the GitHub repository. The chart manifests should be already
+# built with the target `helm-build` and the manifests changes already committed to the tag to be released.
+
+name: Release Helm Chart
+on:
+  release:
+    types:
+      - published
+  workflow_dispatch:
+    inputs:
+      operatorTag:
+          description: Operator bundle version tag
+          default: v0.0.0
+          type: string
+jobs:
+  chart_release:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v4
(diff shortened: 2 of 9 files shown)


didierofrivia:
It might require extra steps for validation, since I'm not familiar with the operator itself. Also with the construction of the manifests... maybe @maleck13, @mikenairn or @philbrookes can point what else is needed?

mikenairn (approved):
```
kind create cluster --name kuadrant-local
make helm-install
kubectl wait --timeout=60s --for=condition=Available deployments --all -l app.kubernetes.io/part-of=dns-operator -A
kubectl create namespace dnstest
make local-setup-managedzones TARGET_NAMESPACE=dnstest
make test-e2e TEST_DNS_MANAGED_ZONE_NAME=dev-mz-gcp TEST_DNS_ZONE_DOMAIN_NAME=mn.google.hcpapps.net TEST_DNS_NAMESPACE=dnstest TEST_DNS_PROVIDER=gcp
...
Ran 8 of 8 Specs in 137.959 seconds
SUCCESS! -- 8 Passed | 0 Failed | 0 Pending | 0 Skipped
PASS

Ginkgo ran 1 suite in 2m20.271935527s
Test Suite Passed
```

Personally I'm not convinced by the need for a helm chart, at least from a technical perspective. I can achieve the same install for testing purposes with `kubectl apply -k github.com/Kuadrant/dns-operator/config/default?ref=main` (Released versions can also be specified) and have full control over how i want to modify that with an overlay should i need it, and have my CD manage the deployment of it. The templating of the deployments and other resources also seems like it could be a maintenance headache fairly quickly.

Maybe there is some gain in terms of user adoption we are aiming for here, people that already use helm, and want the management of deployments that it offers?

Since that helm build copies the result of the kustomize build command you likely want to make sure we have suitable CI in place to make sure it doesn't get left behind, similar to `verify-bundle`.

didierofrivia:
Thanks for the review and feedback @mikenairn ! I agree that one can achieve the same, at this stage, applying the `default` ksutomization we already have. However, as you also guessed, this aims to widen the alternatives of adoption and management and in the near future give users flexible customization with Helm templating.

PR #191, opened 2024-07-15
Version commit hash

ehearneRedHat:
### What:
Added version commit hash implementation similar to [authorino](https://github.com/Kuadrant/authorino/issues/471) and [limitador-operator](https://github.com/Kuadrant/limitador-operator/pull/149)

### How: 

- Check Dirty script checks to see if branch has uncomitted changes and reports dirty if so . 
- Target level vars for specific targets that require ldflags in Makefile.
- Dockerfile and workflow now reference new vars

### Verify:

- `make build` (requires `kind create cluster` be ran first)
- `make docker-build` (load image into cluster)
- `make run` (requires `kind create cluster` be ran first)

Specified make targets should run fine with new information in their logs when checked such as the version, whether the branch had changes (i.e. dirty) and also the commit hash. Below is an example of what that may look like. 

```bash
2024-07-22T13:48:24+01:00       INFO    setup   dns-operator 0.4.0-dev (96475e401ed30c233419dc31632e2ed6729f708a)
```

Workflow runs successfully --> https://github.com/Kuadrant/dns-operator/actions/runs/10055959082

Image, bundle image and catalog image from workflow install fine and image/bundle image displays new changes.


Makefile
```diff
@@ -5,6 +5,8 @@
 # - use environment variables to overwrite this value (e.g export VERSION=0.0.2)
 VERSION ?= 0.0.0
 
+PROJECT_PATH := $(patsubst %/,%,$(dir $(MKFILE_PATH)))
+
```
mikenairn, line 9: Is this used anywhere?

cmd/main.go
```diff
@@ -47,6 +47,9 @@ import (
 var (
 	scheme   = runtime.NewScheme()
 	setupLog = ctrl.Log.WithName("setup")
+	version  = "0.4.0-dev" // Change as versions are released
```
mikenairn, line 50: This should be loaded from a version file.

mikenairn (commented):
Adding version info the the dns operator and exposing it in the runtime logs is a good idea and something we intended to do at some point, so thanks for starting that.

However, I'm looking through the other repos and issues you linked and nothing about the way we are doing it in each repo is aligned, each is doing it differently still from what i can gather. Looks like we are also in each PR changing it based on whatever feedback your are getting on each PR which won't really work if we are trying to go for a common approach across all repos. Do you have a repo that has been updated with the changes we intend to make in them all?

Would it be worth picking a repo, implementing version info changes in it(on a branch), and getting agreement that all four main golang based operator repos follow the same setup? I agree in most part with what is said [here](https://github.com/Kuadrant/kuadrantctl/pull/85#issuecomment-2147851247), but i would expect to see automation in place that ensures that version file is as expected on each branch.

The `dirty` flag also, while i can see the need for it, doesn't really make much sense in a semantic version string. I prefer how this operator has done it making it explicitly about the state of the git repo which it is https://github.com/hashicorp/vault-secrets-operator/blob/dcfee8e513e198aafad467ae71ec5cdd39269282/scripts/ldflags-version.sh#L51-L53

ehearneRedHat:
> Do you have a repo that has been updated with the changes we intend to make in them all?

No, there isn't a particular repo in which a universal implementation is made. As you mentioned, I have been implementing each one by one. 

> Would it be worth picking a repo, implementing version info changes in it(on a branch), and getting agreement that all four main golang based operator repos follow the same setup?

That makes sense. That way, the implementation can remain the same.

> The dirty flag also, while i can see the need for it, doesn't really make much sense in a semantic version string. I prefer how this operator has done it making it explicitly about the state of the git repo which it is

I can see where you're coming from... I anticipate some changes will be needed for this implementation to be universal.

Okay, so your suggestions are to determine a universal implementation strategy across all four operators, to be done through a PR so that anyone who make suggestions can so that we can come to an agreement, and then within that agreement, ensure the suggestion of how the aformentioned operator repo implemented "dirty" ?

Makefile
ehearneRedHat: No, it isn't. Removing now...

cmd/main.go
ehearneRedHat: I'll seek to add that to the suggestion below and act on it.

.github/workflows/build-images.yaml
```diff
@@ -46,6 +46,10 @@ jobs:
           image: ${{ env.OPERATOR_NAME }}
           tags: ${{ env.IMG_TAGS }}
           platforms: linux/amd64,linux/arm64
+          build-args: |
+            COMMIT=${{ github.sha }}
```
eguzki, line 50: like in limitador operator https://github.com/Kuadrant/limitador-operator/blob/main/.github/workflows/build-images-base.yaml#L78

let it be 
```
GIT_SHA=${{ github.sha }}
```

.github/workflows/build-images.yaml
```diff
@@ -92,6 +96,9 @@ jobs:
           image: ${{ env.OPERATOR_NAME }}-bundle
           tags: ${{ needs.build.outputs.build-tags }}
           platforms: linux/amd64,linux/arm64
+          build-args: |
+            COMMIT=${{ github.sha }}
```
eguzki, line 100: `GIT_SHA=${{ github.sha }}`

Dockerfile
```diff
@@ -21,7 +28,7 @@ COPY internal/ internal/
 # was called. For example, if we call make docker-build in a local env which has the Apple Silicon M1 SO
 # the docker BUILDPLATFORM arg will be linux/arm64 when for Apple x86 it will be linux/amd64. Therefore,
 # by leaving it empty we can ensure that the container and binary shipped on it will have the same platform.
-RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -o manager cmd/main.go
+RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -ldflags "-X main.commit=${COMMIT} -X main.dirty=${DIRTY}" -a -o manager cmd/main.go
```
eguzki, line 31: https://github.com/Kuadrant/limitador-operator/blob/main/Dockerfile#L25

` "-X main.gitSHA=${GIT_SHA} -X main.dirty=${DIRTY}"`

cmd/main.go
eguzki: maybe something like this https://github.com/Kuadrant/limitador-operator/blob/main/version/version.go @ehearneRedHat

cmd/main.go
```diff
@@ -129,6 +132,16 @@ func main() {
 		providers = defaultProviders
 	}
 
+	// Display version here.
+
+	dnsInfoDisplay := fmt.Sprintf("dns-operator %s (%s-dirty)", version, commit)
```
eguzki, line 137: after setting the log options in the logger, I would implement something like the limitador operator is doing. 

```go
func printControllerMetaInfo() {                                                        
    setupLog := log.Log                                                                 
                                                                                        
    setupLog.Info(fmt.Sprintf("go version: %s", runtime.Version()))                     
    setupLog.Info(fmt.Sprintf("go os/arch: %s/%s", runtime.GOOS, runtime.GOARCH))       
    setupLog.Info("base logger", "log level", logLevel, "log mode", logMode)            
    setupLog.Info("", "version", version.Version, "commit", gitSHA, "dirty", dirty)     
}                                                                                       

```

this function should be called first thing in `main` ideally. In the limitador operator the logger is set at `init` time, even before `main` is called because it is based on env vars. If this the DNS operator relies on command line args to configure the logger, it needs to be done in main. So, first thing to do is to  parse command line args, then set the logger and then, inmediately log operator version and other info. If there is any error afterwards, we can tell which version of the operator is being run.

eguzki (changes requested):
some changes pointed out

ehearneRedHat:
@mikenairn pinging for review since @eguzki is on PTO for 3 weeks. :smile:

version/version.go
```diff
@@ -0,0 +1,21 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package version
+
+var (
+	// This variable is dependent on what the current release is e.g. if it is v0.10.0 then this variable, outside of releases, will be v0.10.1-dev .
+	Version = "0.4.0-dev" // Change as versions are released
+)
```
mikenairn, line 21: This should be in `internal` i.e. `internal/version/version.go`

Dockerfile
```diff
@@ -15,13 +22,15 @@ RUN go mod download
 COPY cmd/main.go cmd/main.go
 COPY api/ api/
 COPY internal/ internal/
+COPY version/ version/
+COPY pkg/ pkg/
```
mikenairn, line 26: This repo was created using the newer layout(v4) that uses an "internal" package for all private application code, anything in here isn't exposed by the go module see https://github.com/golang-standards/project-layout/tree/master/internal and https://go.dev/doc/go1.4#internalpackages

Everything you are doing here should be in the internal package.

Makefile
```diff
@@ -214,19 +214,25 @@ local-deploy-namespaced: docker-build kind-load-image ## Deploy the dns operator
...
 build: manifests generate fmt vet ## Build manager binary.
-	go build -o bin/manager cmd/main.go
+	go build -ldflags "-X main.gitSHA=${GIT_SHA} -X main.dirty=${DIRTY}" -o bin/manager cmd/main.go
 
 .PHONY: run
+run: GIT_SHA=$(shell git rev-parse HEAD || echo "unknown") 
+run: DIRTY=$(shell hack/check-git-dirty.sh || echo "unknown")
 run: manifests generate fmt vet ## Run a controller from your host.
-	go run ./cmd/main.go --zap-devel --provider inmemory,aws,google,azure
+	go run -ldflags "-X main.gitSHA=${GIT_SHA} -X main.dirty=${DIRTY}" ./cmd/main.go --zap-devel --provider inmemory,aws,google,azure
```
mikenairn, line 226: nit: Personally I'd prefer it if these weren't required at runtime. Seems to me default values could be set in code for these.

Makefile
```diff
@@ -214,19 +214,25 @@ local-deploy-namespaced: docker-build kind-load-image ## Deploy the dns operator
...
 .PHONY: build
+build: GIT_SHA=$(shell git rev-parse HEAD || echo "unknown") 
+build: DIRTY=$(shell hack/check-git-dirty.sh || echo "unknown")
 build: manifests generate fmt vet ## Build manager binary.
-	go build -o bin/manager cmd/main.go
+	go build -ldflags "-X main.gitSHA=${GIT_SHA} -X main.dirty=${DIRTY}" -o bin/manager cmd/main.go
 
 .PHONY: run
+run: GIT_SHA=$(shell git rev-parse HEAD || echo "unknown") 
+run: DIRTY=$(shell hack/check-git-dirty.sh || echo "unknown")
```
mikenairn, line 224: nit: I see we are basing these on what we have done in other repos, so we can leave it. My personal preference would have been to make them more explicit in their naming though https://github.com/hashicorp/vault-secrets-operator/blob/dcfee8e513e198aafad467ae71ec5cdd39269282/scripts/ldflags-version.sh#L52-L53

version/version.go
mikenairn: Presumably this will also need updated as part of a release. Do any doc updates need to be made to call this out, [here](https://github.com/Kuadrant/dns-operator/blob/main/docs/RELEASE.md) for example? Also the chances of this not getting updated correctly are fairly high so we should probably consider trying to add some automation to ensure it is kept correct. i.e. does it match the version being set at build time. (Don't need to do that in this PR)

Makefile
ehearneRedHat: It makes it more difficult to have them hardcoded as it requires using Git in Go. However, no problem with doing this if that is your preference. :)

Makefile
ehearneRedHat: We could do that, but the issue is needing to re-do on all other operators for consistency. The second round of this implementation can address this ?

cmd/main.go
```diff
@@ -60,9 +68,24 @@ func init() {
 
 	utilruntime.Must(v1alpha1.AddToScheme(scheme))
 	//+kubebuilder:scaffold:scheme
+
+	logger := log.NewLogger(
+		log.SetLevel(log.ToLevel(logLevel)),
+		log.SetMode(log.ToMode(logMode)),
+		log.WriteTo(os.Stdout),
+	)
+	log.SetLogger(logger)
```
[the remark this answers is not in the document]
ehearneRedHat: Sure sounds good.

ehearneRedHat:
@mikenairn sorry for taking a while to get back to this PR... I was PTO last week and have been catching up with another PR previous . 

I've amended changes where there wasn't any questions I had for the suggestions. As for the rest, I'll await your response first :)

cmd/main.go
```diff
@@ -62,6 +65,10 @@ func init() {
 	//+kubebuilder:scaffold:scheme
 }
 
+func printControllerMetaInfo() {
+	setupLog.Info("", "version", version.Version, "commit", gitSHA, "dirty", dirty)
```
eguzki, line 69:
```suggestion
	setupLog.Info("build information", "version", version.Version, "commit", gitSHA, "dirty", dirty)
```

eguzki (commented):
overall LGTM

I left one nit, though

mikenairn (approved):
Changes look fine. As discussed offline, consider looking into using rebase to bring your dev branch up to date with the base(main) to avoid unnecessary merge commits.

I'll force a quash merge on this one.

PR #197, opened 2024-07-24
Update logging

### LINKED ISSUE #117: Add audit logs for actions taken against a given zone

makslion:
closes #117 

PR has two changes: 
- Rework logging to ensure logs make sense regardless of the logging level and provide needed information for debugging. 
- Document the logging and provide an example of querying logs locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,6 +103,52 @@ make manifests
 
 More information can be found via the [Kubebuilder Documentation](https://book.kubebuilder.io/introduction.html)
 
+## Logging
+Logs are following the general guidelines: 
+- `logger.Info()` describe a high-level state of the resource such as creation, deletion and which reconciliation path was taken.  
+- `logger.Error()` describe only those errors that are not returned in the result of the reconciliation. If error is occurred there should be only one error message. 
+- `logger.V(1).Info()` debug level logs to give information about every change or event caused by the resource as well as every update of the resource.
+
+The `--zap-devel` argument will enable debug level logs for the output. Otherwise, all `V()` logs are ignored.
+
+### Common metadata 
+Not exhaustive list of metadata for DNSRecord controller:
+- `level` - logging level. Values are: `info`,`debug` or `error`
+- `ts` - timestamp
+- `logger` - logger name
+- `msg`
+- `controller` and `controllerKind` - controller name, and it's kind respectively to output the log
+- `DNSRecord` - name and namespace of the DNS Record CR that is being reconciled
+- `reconcileID`
+- `ownerID` - ID the of owner of the DNS Record 
+- `txtPrefix`/`txtSuffix` - prefix and suffix of the TXT record in provider. 
+- `zoneEndpoints` - endpoints that exist in the provider
+- `specEdnoinds` - endpoints defined in the spec
+- `statusEndpoints` - endpoints that were processed previously
+
+> Note that not all the metadata values are present at each of the logs statements. 
+
+### Examples
+To query logs locally you can use `jq`. For example:
+Retrieve logs by 
+```shell
+kubectl get deployments -l app.kubernetes.io/part-of=dns-operator -A
+
+NAMESPACE             NAME                              READY 
+dns-operator-system   dns-operator-controller-manager   1/1   
+```
+And query them. For example:
+```shell
+kubectl logs -l control-plane=dns-operator-controller-manager -n dns-operator-system --tail -1 | sed '/^{/!d' | jq 'select(.controller=="dnsrecord" and .level=="info")'
+```
+or 
+```shell
+kubectl logs -l control-plane=dns-operator-controller-manager -n dns-operator-system --tail -1 | sed '/^{/!d' | jq 'select(.controller=="dnsrecord" and .DNSRecord.name=="test" and .reconcileID=="2be16b6d-b90f-430e-9996-8b5ec4855d53")' | jq '.level, .msg, .zoneEndpoints, .specEndpoints, .statusEndpoints '
+
+```
+You could use selector in the `jq` with `and`/`not`/`or` to restrict.
+
+
 ## License
 
 Copyright 2024.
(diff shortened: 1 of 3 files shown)


internal/controller/dnsrecord_controller.go
```diff
@@ -200,13 +204,15 @@ func (r *DNSRecordReconciler) updateStatus(ctx context.Context, previous, curren
...
 
+	logger.V(1).Info("Updating status of DNSRecord")
 	// short loop. We don't publish anything so not changing status
 	if prematurely, requeueIn := recordReceivedPrematurely(current); prematurely {
 		return reconcile.Result{RequeueAfter: requeueIn}, nil
 	}
 
 	// failure
 	if specErr != nil {
+		logger.Error(specErr, "Error reconciling DNS Record")
```
mikenairn, line 215: IMO we shouldn't need to do this since we should be returning any specErr in the response from this function which would be logged by controller runtime automatically.

internal/controller/dnsrecord_controller.go
```diff
@@ -174,6 +176,7 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	var reason, message string
 	err = dnsRecord.Validate()
 	if err != nil {
+		logger.V(1).Error(err, "Failed to validate DNSRecord")
```
mikenairn, line 179: I don't think the level makes any difference on Error logs i.e. `logger.Error` does the same

makslion: yes, you are correct

internal/controller/dnsrecord_controller.go
makslion: I was to generous with errors. However, we never return the `specErr` and it could lead to awkward debugging.

internal/controller/managedzone_controller.go
```diff
@@ -403,7 +414,7 @@ func (r *ManagedZoneReconciler) parentZoneNSRecordReady(ctx context.Context, man
 
 	nsRecordReady := meta.IsStatusConditionTrue(nsRecord.Status.Conditions, string(v1alpha1.ConditionTypeReady))
 	if !nsRecordReady {
-		return fmt.Errorf("the ns record is not in a ready state : %s", nsRecord.Name)
+		return fmt.Errorf("the ns record is not in a ready state : %s. Reason: %s", nsRecord.Name, meta.FindStatusCondition(nsRecord.Status.Conditions, string(v1alpha1.ConditionTypeReady)).Reason)
 	}
 	return nil
 }
```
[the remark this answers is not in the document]
makslion: got rid of those changes

README.md
```diff
@@ -105,6 +105,28 @@ make manifests
...
+
+NAMESPACE             NAME                              READY 
+dns-operator-system   dns-operator-controller-manager   1/1   
+```
+And query them. For example:
+```shell
+kubectl logs -l control-plane=dns-operator-controller-manager -n dns-operator-system --tail -1 | sed '/^{/!d' | jq 'select(.controller=="dnsrecord" and .level=="info")'
+```
+You could use selector in the `jq` with `and`/`not`/`or` to restrict. 
+
```
mikenairn, line 128: Should we not list here the common metadata that can be attached to a log statement and what they mean (ownerID, zoneID etc..), anything we add at different parts of the reconcile using [log.WithValues](https://github.com/Kuadrant/dns-operator/blob/main/internal/controller/dnsrecord_controller.go#L124-L125) ? Maybe a generic example of how you can filter logs using jq for any of these or a combination of values.  I kind of thought that was the main point of #117 (Maybe I'm wrong).

I'm less concerned about the testing it in an aggregator part of that issue, if the logs are structured (which they are), and the metadata is on the requests then they can be filtered using whatever tool you choose to use for log aggregation and querying.

I'm not sure this PR is doing enough to cover the first two asks of #117 though:

* Log metadata defined and documented
* Logging reviewed in the controllers and providers updated

makslion: Added section with `common metadata` with some of the keys (not all of them as maintaining it doesn't sound feasible) and another example of `jq` usage.

makslion: The task also says "logs metadata defined and documented". The documentation seems straightforward, but with "defined" - all the important ones (the ones that will allow to narrow down logs) are there anyway. Reconcile ID and CR name will be added anyway so I'm not sure how else we could "define" them :thinking:

README.md
```diff
@@ -103,6 +103,52 @@ make manifests
...
+- `logger` - logger name
+- `msg`
+- `controller` and `controllerKind` - controller name, and it's kind respectively to output the log
+- `DNSRecord` - name and namespace of the DNS Record CR that is being reconciled
+- `reconcileID`
+- `ownerID` - ID the of owner of the DNS Record 
+- `txtPrefix`/`txtSuffix` - prefix and suffix of the TXT record in provider. 
+- `zoneEndpoints` - endpoints that exist in the provider
+- `specEdnoinds` - endpoints defined in the spec
+- `statusEndpoints` - endpoints that were processed previously
```
mikenairn, line 127: nit: I would have kept this list down to the important pieces of data that are attached to all requests and not include some that are logged in one log statement. zoneEndpoint, specEndpoints and statusEndpoints txtPrefix are not goign to be on all requests.

mikenairn (approved):
Lets move on, log statements are just going to keep getting changed as we progress so i don't see any point dragging this PR out any logger. 

The important thing is the logger is using structured logs, is configurable, and we have at least some important meatadata attached to requests to allow filtering.

PR #200, opened 2024-07-25
Unify e2e test suite

### LINKED ISSUE #186: tests: Combine e2e-multi and e2e test suites
**What**

Create one single e2e test suite under `test/e2e` that can handle both single and multi instance(deployment and cluster) tests. Test cases that are explicitly for multi cluster scenarios can be labelled as such.

**Why** 

Initially it seemed likely that we needed to keep these separate, but as the multi instance tests have progressed it's became more clear that we could just use the same test setup in both negating the need for two different suites. 

**How**

* Update e2e test suite(`test/e2e`) setup to follow the same pattern as the multi instance suite(`test/e2e/multi_instance`)
* Replace `test/e2e/multi_record_test.go` with `test/e2e/multi_instance/multi_record_test.go`
* Remove `test/e2e/multi_instance/` and all associated make targets
* Update all other tests to load the first managed zone and use that for the test
* Label all test cases appropriately (single vs multi)
* Update github action (We coudl consider adding 2 clusters here so we can test the multi tests correctly)

mikenairn:
closes #186 

Combines the e2e and e2e-multi test suites into one unified test suite. Tests for single and multi instance scenarios can still be executed individually if required, but all tests should work in both cases.

`make test-e2e` - Runs all e2e tests
`make test-e2e-multi` - Runs only e2e tests for multiple record scenarios

Remove local-setup-multi in favour of adding the cluster count option to local-setup.

```
make local-setup DEPLOY=true CLUSTER_COUNT=2
```

Only noticeable change is that all clusters generated by local-setup, single or multi cluster, have a suffix for the count.

```
make local-setup DEPLOY=true
kind get clusters
kuadrant-dns-local-1
```


README.md
```diff
@@ -84,15 +84,13 @@ kubectl logs -f deployments/dns-operator-controller-manager -n dns-operator-syst
 The e2e test suite can be executed against any cluster running the DNS Operator with configuration added for any supported provider.
 
 ```
-make test-e2e TEST_DNS_MANAGED_ZONE_NAME=<My managed zone name> TEST_DNS_ZONE_DOMAIN_NAME=<My domain name> TEST_DNS_NAMESPACE=<My test namesapace> TEST_DNS_PROVIDER=<aws|gcp>
+make test-e2e TEST_DNS_MANAGED_ZONE_NAME=<My managed zone name> TEST_DNS_NAMESPACES=<My test namesapace(s)>
```
philbrookes, line 87: nit: `namesapace` > `namespace`

test/e2e/README.md
```diff
@@ -32,9 +41,40 @@ dnstest          dev-mz-aws   mn.hcpapps.net
 dnstest          dev-mz-gcp   mn.google.hcpapps.net
 ```
 
+### Cluster scoped on multiple clusters
+
+Deploy the operator on two kind cluster each with one operator instance watching all namespaces:
```
philbrookes, line 46: clusters

test/e2e/README.md
```diff
@@ -32,9 +41,40 @@ dnstest          dev-mz-aws   mn.hcpapps.net
...
+```shell
+make local-setup DEPLOY=true CLUSTER_COUNT=2
+```
+
+### Namespace scoped on multiple clusters
+
+Deploy the operator on two local kind cluster with two operator instances in two namespaces watching their own namespace only:
```
philbrookes, line 53: clusters

test/e2e/suite_test.go
```diff
@@ -61,45 +98,174 @@ var _ = BeforeSuite(func(ctx SpecContext) {
...
 	SetTestEnv("testGeoCode", geoCode)
 })
 
+// setConfigFromEnvVars loads test suite runtime configurations from env vars.
+// dnsManagedZoneName managed zone name expected to exist in each test namespace (i.e. dev-mz-aws).
+// dnsNamespaces test namespaces, comma seperated list (i.e. dns-operator-1,dns-operator-2)
+// deploymentCount number of test namespaces expected. Appends an index suffix to the dnsNamespaces, only used if dnsNamespaces is a single length array.
+//
+// Examples:
+// dnsNamespaces=dns-operator deploymentCount=<unset> = dnsNamespaces=dns-operator
```
philbrookes, line 124: I'm struggling with the syntax here, why is there a floating `=`?

philbrookes (approved):
looks fine, a few small comments/questions but looks OK to merge to me.

test/e2e/suite_test.go
mikenairn: I had a go at trying to make the example a bit clearer. They are trying to show if you give these inputs, this is what the resulting config will be.

test/e2e/suite_test.go
```diff
@@ -61,45 +98,174 @@ var _ = BeforeSuite(func(ctx SpecContext) {
...
+		namespacesStr = os.Getenv("TEST_DNS_NAMESPACE")
+		if namespacesStr == "" {
+			return fmt.Errorf("env variable '%s' must be set", dnsNamespaces)
+		}
 	}
-	if testDNSProvider = os.Getenv(dnsProvider); testDNSProvider == "" {
-		return fmt.Errorf("env variable '%s' must be set", dnsProvider)
+
+	namespaces := strings.Split(namespacesStr, ",")
+	if len(namespaces) == 1 {
```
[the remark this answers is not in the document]
mikenairn: I'm not sure, maybe there is  a better way of doing it but it's not totally clear to me what that would look like. 

Current logic is if we have a single namespace passed as input then we should check to see if we are expecting to test against multiple namespaces based on "DEPLOYMENT_COUNT". If that is set, we append a suffix to the single  namespace given, one for each deployment. 

```
input:
namespaces: dnstest
deployment count: 2
output
namespaces: dnstest-1,dnstest-2
```

If there are multiple namespaces passed as input then you are explicitly saying what the namespaces to test in are and we ignore "DEPLOYMENT_COUNT" entirely.

```
input:
namespaces: dnstest,someotherns
deployment count: 200
output
namespaces: dnstest, someotherns
```

Same logic applies to contexts.

philbrookes (approved):

PR #203, opened 2024-07-31
WIP chore: Remove ManagedZone API

mikenairn:


.github/workflows/ci-e2e.yaml
```diff
@@ -5,6 +5,7 @@ on:
     branches:
       - main
       - "release-*"
+      - remove_managed_zone_api
```
mikenairn, line 8: ToDo remove before merge

config/manager/kustomization.yaml
```diff
@@ -7,4 +7,6 @@ kind: Kustomization
 images:
 - name: controller
   newName: quay.io/kuadrant/dns-operator
-  newTag: latest
+  newTag: remove_managed_zone_api
+
+#ToDo mnairn: Revert tag before merge
```
mikenairn, line 12: ToDo revert before merge

internal/controller/dnsrecord_controller.go
```diff
@@ -395,29 +448,30 @@ func setDNSRecordCondition(dnsRecord *v1alpha1.DNSRecord, conditionType string,
...
-		DomainFilter:   externaldnsendpoint.NewDomainFilter([]string{managedZone.Spec.DomainName}),
+		DomainFilter:   externaldnsendpoint.NewDomainFilter([]string{dnsRecord.Status.ZoneDomainName}),
 		ZoneTypeFilter: externaldnsprovider.NewZoneTypeFilter(""),
-		ZoneIDFilter:   externaldnsprovider.NewZoneIDFilter([]string{managedZone.Status.ID}),
+		ZoneIDFilter:   externaldnsprovider.NewZoneIDFilter([]string{dnsRecord.Status.ZoneID}),
 	}
 
-	return r.ProviderFactory.ProviderFor(ctx, managedZone, providerConfig)
+	return r.ProviderFactory.ProviderFor(ctx, dnsRecord, providerConfig)
 }
```
mikenairn, line 459: This is only being used by the health check reconciler now. Since that will all being getting changed when we remove provider health checks I'm just leaving it for now.

pkg/builder/provider.go
```diff
@@ -0,0 +1,97 @@
...
+func (pb *ProviderBuilder) Build() *corev1.Secret {
+	return &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      pb.name,
+			Namespace: pb.namespace,
+		},
+		StringData: pb.strDataItems,
+		Type:       pb.secretType,
+	}
+}
```
mikenairn, line 97: I added this provider builder as a public API so it could be used in both the tests here and in kuadrant operator. Unsure of the best location for it, I'm not really sure if an API like this should be inside the `api/v1alpha1` package, or just be outside and always work with the latest version of the APIs.

Makefile
```diff
@@ -184,8 +184,8 @@ local-setup-cluster: $(KIND) ## Setup local development kind cluster, dependenci
 
 	@echo "local-setup: Check dns operator deployments"
 	$(KUBECTL) get deployments -l app.kubernetes.io/part-of=dns-operator -A
-	@echo "local-setup: Check managedzones"
-	$(KUBECTL) get managedzones -A
+	@echo "local-setup: Check dns providers"
+	$(KUBECTL) get secrets -A
```
philbrookes, line 188: won't this list unrelated secrets, i.e. all the secrets in the cluster? Could we add a label selector or something, or am I mistaken?

internal/controller/dnsrecord_controller.go
```diff
@@ -106,40 +107,48 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+				return ctrl.Result{}, err
+			}
+			// if hadChanges - the deleteRecord has successfully applied changes
+			// in this case we need to queue for validation to ensure DNS Provider retained changes
+			// before removing finalizer and deleting the DNS Record CR
+			if hadChanges {
+				return ctrl.Result{RequeueAfter: randomizedValidationRequeue}, nil
+			}
+		} else {
+			logger.Info(fmt.Sprintf("unable to perform cleanup of dns provider for zone status id: '%s', dominaName: '%s'",
```
maleck13, line 150: is it possible to give more info than this? The way I read it, this can happen if we failed to set a zid or domain inside the zone. As this is a delete also makes me wonder if this could happen. I guess if a record was created and there was a bad credential and then you tried to delete it?

internal/controller/dnsrecord_controller.go
```diff
@@ -106,40 +107,48 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+			}
+			hadChanges, err := r.deleteRecord(ctx, dnsRecord, dnsProvider)
+			if err != nil {
+				logger.Error(err, "Failed to delete DNSRecord")
+				return ctrl.Result{}, err
+			}
+			// if hadChanges - the deleteRecord has successfully applied changes
+			// in this case we need to queue for validation to ensure DNS Provider retained changes
+			// before removing finalizer and deleting the DNS Record CR
+			if hadChanges {
```
maleck13, line 146: this flow and letting it fall through to https://github.com/Kuadrant/dns-operator/pull/203/files#diff-adacb6a4589c65c2d853c0f59ca02d6b8fb50bb388b41de1d13e99b0d128cfe1R173 seem the same? Wondering if it could be removed as it will aways result in same result?

maleck13: Oh actually never mind

internal/controller/dnsrecord_controller.go
```diff
@@ -164,34 +173,83 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
 		if err != nil {
-			return ctrl.Result{}, err
+			setDNSRecordCondition(dnsRecord, string(v1alpha1.ConditionTypeReady), metav1.ConditionFalse,
+				"DNSProviderError", fmt.Sprintf("The dns provider could not be loaded: %v", err))
+			return r.updateStatus(ctx, previous, dnsRecord, false, err)
 		}
-		err = r.Client.Update(ctx, dnsRecord)
-		return ctrl.Result{}, err
+
+		z, err := dnsProvider.DNSZoneForHost(ctx, dnsRecord.Spec.RootHost)
```
maleck13, line 199: Should we do this or should we just expect the zoneid to be specified? Wondering a user wouldn't just specify the zone id and instead allow us to find it?

Makefile
mikenairn: :+1: There are labels being added to the secrets so we can use a the same on as we are using for deployments `-l app.kubernetes.io/part-of=dns-operator`

output now:

```
local-setup: Check dns providers                           
NAMESPACE   NAME                                TYPE                   DATA   AGE
dnstest     dns-provider-credentials-aws        kuadrant.io/aws        5      1s
dnstest     dns-provider-credentials-azure      kuadrant.io/azure      3      1s                                     
dnstest     dns-provider-credentials-gcp        kuadrant.io/gcp        4      1s
dnstest     dns-provider-credentials-inmemory   kuadrant.io/inmemory   1      0s
local-setup: Complete!!
```

internal/controller/dnsrecord_controller.go
maleck13: Based on conversation around this, decided to go with removing the secret filter options and instead relying on the credential to be limited down to what we should see. We will add docs giving examples of how to do this. This will mean we can avoid having things sepecified in two places (in the secret config and in the actual IAM policies etc). If we find two zones for the same domain, we will error out as we don't want to just choose one. In this case the permissions the credential has should be modified to remove the duplicate

internal/controller/dnsrecord_controller.go
mikenairn: Yes, if the record is created and for whatever reason it isn't able to assign a zone e.g. invalid credentials, no zone that matches the root host etc... we shouldn't try and cleanup when the record is deleted since it won't have been able to create anything in the first place.

maleck13 (approved):

PR #211, opened 2024-08-19
Add Azure provider

philbrookes:
Add full Azure support, and enable all e2e tests.

### DIFF
--- a/examples/kuadrant/combined-loadbalanced-dns.yaml
+++ b/examples/kuadrant/combined-loadbalanced-dns.yaml
@@ -33,6 +33,6 @@ spec:
     setIdentifier: GEO-EU
     targets:
     - geo-eu.klb.${KUADRANT_SUB_DOMAIN}
-  managedZone:
-    name: dev-mz-azure
+  providerRef:
+    name: dns-provider-credentials-azure
   rootHost: ${KUADRANT_SUB_DOMAIN}
--- a/examples/kuadrant/combined-loadbalanced-dns.yaml
+++ b/examples/kuadrant/combined-loadbalanced-dns.yaml
@@ -1,0 +1,38 @@
+apiVersion: kuadrant.io/v1alpha1
+kind: DNSRecord
+metadata:
+  name: combined-loadbalanced-dns
+  namespace: dnstest
+spec:
+  endpoints:
+  - dnsName: cluster2.klb.${KUADRANT_SUB_DOMAIN}
+    recordTTL: 60
+    recordType: A
+    targets:
+    - 127.0.0.2
+  - dnsName: ${KUADRANT_SUB_DOMAIN}
+    recordTTL: 300
+    recordType: CNAME
+    targets:
+    - klb.${KUADRANT_SUB_DOMAIN}
+  - dnsName: geo-eu.klb.${KUADRANT_SUB_DOMAIN}
+    providerSpecific:
+    - name: weight
+      value: "200"
+    recordTTL: 60
+    recordType: CNAME
+    setIdentifier: cluster2.klb.${KUADRANT_SUB_DOMAIN}
+    targets:
+    - cluster2.klb.${KUADRANT_SUB_DOMAIN}
+  - dnsName: klb.${KUADRANT_SUB_DOMAIN}
+    providerSpecific:
+    - name: geo-code
+      value: GEO-EU
+    recordTTL: 300
+    recordType: CNAME
+    setIdentifier: GEO-EU
+    targets:
+    - geo-eu.klb.${KUADRANT_SUB_DOMAIN}
+  managedZone:
+    name: dev-mz-azure
+  rootHost: ${KUADRANT_SUB_DOMAIN}
--- a/examples/kuadrant/geo-loadbalanced-healthchecks-dns.yaml
+++ b/examples/kuadrant/geo-loadbalanced-healthchecks-dns.yaml
@@ -1,7 +1,7 @@
 apiVersion: kuadrant.io/v1alpha1
 kind: DNSRecord
 metadata:
-  name: loadbalanced-healthchecks-dns
+  name: geo-loadbalanced-healthchecks-dns
 spec:
   endpoints:
     - dnsName: 14byhk-2k52h1.klb.${KUADRANT_SUB_DOMAIN}
@@ -63,6 +63,6 @@ spec:
     failureThreshold: 3
     port: 80
     protocol: HTTPS
-  providerRef:
-    name: dns-provider-credentials-azure
+  managedZone:
+    name: dev-mz-azure
   rootHost: ${KUADRANT_SUB_DOMAIN}
--- a/examples/kuadrant/geo-loadbalanced-healthchecks-dns.yaml
+++ b/examples/kuadrant/geo-loadbalanced-healthchecks-dns.yaml
@@ -63,6 +63,4 @@ spec:
     failureThreshold: 3
     port: 80
     protocol: HTTPS
-  managedZone:
-    name: dev-mz-azure
   rootHost: ${KUADRANT_SUB_DOMAIN}
--- a/examples/kuadrant/weighted-loadbalanced-healthchecks.yaml
+++ b/examples/kuadrant/weighted-loadbalanced-healthchecks.yaml
@@ -54,6 +54,4 @@ spec:
     failureThreshold: 3
     port: 80
     protocol: HTTPS
-  managedZone:
-    name: dev-mz-azure
   rootHost: ${KUADRANT_SUB_DOMAIN}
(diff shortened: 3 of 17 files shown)


philbrookes:
ping @maleck13 @makslion 

obviously ignore all the test changes for now, but any feedback on the azure implementation is appreciated.

examples/kuadrant/geo-loadbalanced-healthchecks-dns.yaml
```diff
@@ -0,0 +1,68 @@
...
+    - dnsName: klb.${KUADRANT_SUB_DOMAIN}
+      providerSpecific:
+        - name: geo-code
+          value: '*'
+      recordTTL: 300
+      recordType: CNAME
+      setIdentifier: default
+      targets:
+        - eu.klb.${KUADRANT_SUB_DOMAIN}
+  healthCheck:
```
maleck13, line 61: so do these work with Azure then? The health checks?

examples/kuadrant/weighted-loadbalanced-healthchecks.yaml
```diff
@@ -1,48 +1,56 @@
 apiVersion: kuadrant.io/v1alpha1
 kind: DNSRecord
 metadata:
-  name: loadbalanced-healthchecks-dns
+  name: weighted-loadbalanced-healthchecks-dns
```
maleck13, line 4: not clear why all these have changed is it just a different subdomain or something more structural

internal/external-dns/provider/azure/azure.go
```diff
@@ -96,7 +109,7 @@ func NewAzureProviderFromConfig(ctx context.Context, azureConfig Config) (*Azure
 // NewAzureProvider creates a new Azure provider.
 //
 // Returns the provider or an error if a provider could not be created.
-func NewAzureProvider(ctx context.Context, configFile string, domainFilter endpoint.DomainFilter, zoneNameFilter endpoint.DomainFilter, zoneIDFilter provider.ZoneIDFilter, resourceGroup string, userAssignedIdentityClientID string, dryRun bool) (*AzureProvider, error) {
+func NewAzureProvider(ctx context.Context, configFile string, domainFilter endpoint.DomainFilter, zoneNameFilter endpoint.DomainFilter, zoneIDFilter provider.ZoneIDFilter, resourceGroup string, userAssignedIdentityClientID string, DryRun bool) (*AzureProvider, error) {
```
maleck13, line 112:
```suggestion
func NewAzureProvider(ctx context.Context, configFile string, domainFilter endpoint.DomainFilter, zoneNameFilter endpoint.DomainFilter, zoneIDFilter provider.ZoneIDFilter, resourceGroup string, userAssignedIdentityClientID string, dryRun bool) (*AzureProvider, error) {
```

internal/external-dns/provider/azure/azure.go
```diff
@@ -105,7 +118,7 @@ func NewAzureProvider(ctx context.Context, configFile string, domainFilter endpo
 	cfg.DomainFilter = domainFilter
 	cfg.ZoneNameFilter = zoneNameFilter
 	cfg.IDFilter = zoneIDFilter
-	cfg.DryRun = dryRun
+	cfg.DryRun = DryRun
```
maleck13, line 121:
```suggestion
	cfg.DryRun = dryRun
```

internal/external-dns/provider/azure/azure.go
```diff
@@ -167,26 +180,26 @@ func (p *AzureProvider) ApplyChanges(ctx context.Context, changes *plan.Changes)
...
 		if err != nil {
 			return nil, err
 		}
 		for _, zone := range nextResult.Value {
-			if zone.Name != nil && p.domainFilter.Match(*zone.Name) && p.zoneIDFilter.Match(*zone.ID) {
+			if zone.Name != nil && p.DomainFilter.Match(*zone.Name) && p.zoneIDFilter.Match(*zone.ID) {
 				zones = append(zones, *zone)
-			} else if zone.Name != nil && len(p.zoneNameFilter.Filters) > 0 && p.zoneNameFilter.Match(*zone.Name) {
-				// Handle zoneNameFilter
+			} else if zone.Name != nil && len(p.ZoneNameFilter.Filters) > 0 && p.ZoneNameFilter.Match(*zone.Name) {
```
maleck13, line 201: did these all change for tests?

internal/external-dns/provider/azure/azure_test.go
```diff
@@ -223,22 +223,22 @@ func createMockRecordSetMultiWithTTL(name, recordType string, ttl int64, values
 }
 
 // newMockedAzureProvider creates an AzureProvider comprising the mocked clients for zones and recordsets
-func newMockedAzureProvider(domainFilter endpoint.DomainFilter, zoneNameFilter endpoint.DomainFilter, zoneIDFilter provider.ZoneIDFilter, dryRun bool, resourceGroup string, userAssignedIdentityClientID string, zones []*dns.Zone, recordSets []*dns.RecordSet) (*AzureProvider, error) {
+func newMockedAzureProvider(domainFilter endpoint.DomainFilter, zoneNameFilter endpoint.DomainFilter, zoneIDFilter provider.ZoneIDFilter, DryRun bool, resourceGroup string, userAssignedIdentityClientID string, zones []*dns.Zone, recordSets []*dns.RecordSet) (*AzureProvider, error) {
```
maleck13, line 226:
```suggestion
func newMockedAzureProvider(domainFilter endpoint.DomainFilter, zoneNameFilter endpoint.DomainFilter, zoneIDFilter provider.ZoneIDFilter, dryRun bool, resourceGroup string, userAssignedIdentityClientID string, zones []*dns.Zone, recordSets []*dns.RecordSet) (*AzureProvider, error) {
```

internal/external-dns/registry/txt.go
```diff
@@ -264,6 +264,7 @@ func (im *TXTRegistry) ApplyChanges(ctx context.Context, changes *plan.Changes)
 		}
 		r.Labels[endpoint.OwnerLabelKey] = im.ownerID
 
+		im.logger.Info("creating TXT record", "TXT record", im.generateTXTRecord(r))
```
maleck13, line 267: prob can remove

internal/provider/azure/azure.go
```diff
@@ -123,3 +134,365 @@ func (p *AzureProvider) getManagedZone(ctx context.Context, zoneID string) (prov
...
+// Records gets the current records.
+//
+// Returns the current records or an error if the operation failed.
+func (p *AzureProvider) Records(ctx context.Context) (endpoints []*externaldnsendpoint.Endpoint, _ error) {
+	zones, err := p.Zones(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	p.logger.Info("getting records from azure")
```
maleck13, line 147: prob needs to be a bit more informative if we want to keep it

examples/kuadrant/geo-loadbalanced-healthchecks-dns.yaml
philbrookes: no, the health checks are only in AWS still, for now. on-cluster health checks is my next piece of work.

examples/kuadrant/weighted-loadbalanced-healthchecks.yaml
philbrookes: I was using "loadbalanced" for what was really "geo-loadbalanced" and now I also want to have "weighted-loadbalanced" so I renamed them to keep them unique.

internal/provider/azure/azure.go
philbrookes: this is very much debugging haha, there is probably a lot of log statements I need to clean up yet.

examples/kuadrant/combined-loadbalanced-dns.yaml
```diff
@@ -0,0 +1,38 @@
+apiVersion: kuadrant.io/v1alpha1
```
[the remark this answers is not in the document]
philbrookes: no, they're just useful DNS Records for testing the various use-cases (instead of going through all the mess of setting up kuadrant-operator and creating a DNS Policy to output these same DNS Records).

philbrookes:
ping @mikenairn would appreciate a review, if you get a chance.

mikenairn (commented):
I haven't went through the provider code too much as i think generally we know we want to go back through all of the provider parts once we have all three providers in place and look at how we can improve them with external-dns compatibility in mind.

Concentrated on testing it actually works. It does work, but as already stated, its really slow. Changing it to API request in a thread was an improvement, but it does still take over 20 seconds on an initial create of a basic loadblanaced record which will block the main thread for that amount of time. In comparison on google or aws this takes about two seconds. This is mainly due to the azure API not supporting batch processing of requests like aws or google, so possibly a bit out of our control. 

This also does mean that the chances of things getting left dangling in the provider is increased because we can no longer ensure that all related resources (TXT owner records, and records) are created in a single transaction.

Not sure if this has been tested at scale (multiple record instances), but based on what i believe the default API limits are for Azure it does seem likely that we could have issues on that front considering how many API calls are involved for a single record. 

As long as the e2 test are passing i see no real reason why we couldn't merge this and iterate on it as required.

PR #216, opened 2024-08-27
ci: Update olm bundle/helm chart verification

mikenairn:
* Update verify-bundle to check config directory as it also modifies the contents of the manager kustomization.
* Add verify-helm-build to check the results of `make helm-build` are committed.
* Update helm-build to be consistent with `make bundle` in what image will be set in the manager kustomization.

### DIFF
--- a/make/helm.mk
+++ b/make/helm.mk
@@ -1,25 +1,9 @@
 ##@ Helm Charts
 
-# TODO: If this is found useful, consider adding it to the main Makefile
-# Semantic versioning (i.e. Major.Minor.Patch)
-is_semantic_version = $(shell [[ $(1) =~ ^[0-9]+\.[0-9]+\.[0-9]+(-.+)?$$ ]] && echo "true")
-
-
-# OPERATOR_IMG defines the dns-operator image with version tag.
-# If the version is not semantic, will use the default one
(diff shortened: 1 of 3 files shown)


make/helm.mk
```diff
@@ -1,25 +1,9 @@
...
-else
-OPERATOR_IMG ?= $(DEFAULT_IMAGE_TAG)
-endif
-
 .PHONY: helm-build
 helm-build: yq manifests kustomize operator-sdk ## Build the helm chart from kustomize manifests
-	# Replace the controller image
-	cd config/manager && $(KUSTOMIZE) edit set image controller=$(OPERATOR_IMG)
+	# Replace the controller image (Should remain consistent with what `make bundle` does)
+	cd config/manager && $(KUSTOMIZE) edit set image controller=$(IMG)
```
mikenairn, line 6: Release branches in this repo set all the expected variable values in a release file which is loaded by Make. The above was not talking any of that into account and  `OPERATOR_IMG` was being evaluated before those values were loaded leading to it using the default "latest" version.

Aside from that, it should be kept consistent with how the bundle is generated to ensure the same version of everything for all installation options (kustomize, olm and helm) .

mikenairn: @didierofrivia fyi

philbrookes (approved):

PR #219, opened 2024-09-03
chore: Immutable RootHost field on DNSRecord

mikenairn:
The dnsrecord controller does not currently handle changes to the rootHost so is now marked as immutable to prevent it from being changed and causing odd behaviour.

In the case where you want to change the host, a new record should be created and the old one removed to ensure proper cleanup of resources in the provider.

part of: https://github.com/Kuadrant/kuadrant-operator/issues/794

### DIFF
--- a/api/v1alpha1/dnsrecord_types.go
+++ b/api/v1alpha1/dnsrecord_types.go
@@ -81,6 +81,7 @@ type DNSRecordSpec struct {
 	// rootHost is the single root for all endpoints in a DNSRecord.
 	// it is expected all defined endpoints are children of or equal to this rootHost
 	// Must contain at least two groups of valid URL characters separated by a "."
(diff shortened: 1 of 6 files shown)


api/v1alpha1/dnsrecord_types.go
```diff
@@ -81,6 +81,7 @@ type DNSRecordSpec struct {
 	// rootHost is the single root for all endpoints in a DNSRecord.
 	// it is expected all defined endpoints are children of or equal to this rootHost
 	// Must contain at least two groups of valid URL characters separated by a "."
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="RootHost is immutable"
```
eguzki, line 84: TIL this :+1:

philbrookes (approved):

Issue #218, opened 2024-09-03
Constantly updating status on invalid dns provider credentials


averevki:
When a secret with *invalid* dns provider credentials is referenced in the newly created dnspolicy, dnspolicy jumps between 2 statuses infinitely every second. It looks like requests are sent every half of a second by continually updating request id's in the dnsrecord. The second status in dnspolicy I see, besides `DNSProviderError`, is `DNSPolicy has encountered some issues: reconcile DNSRecords error error reconciling dns records for gateway gw-averevki--5qem: the object has been modified; please apply your changes to the latest version and try again`. I believe it is caused by constantly changing status in dnsrecord, but not sure. 

With some time waiting, I'm getting just the modified object error most of the time. All of this generates hundreds of logs every second. I'd expect a consistent status to be set and be there until something changes for the better or something is changed by the user himself

PR #220, opened 2024-09-03
bug: Sanitize provider error messages on zone lookup

mikenairn:
fixes: #218  

Fixes an issue where if invalid credentials are used in the provider secret when looking for a suitable zone in the provider it would continually change the status on every reconcile. This is caused by the AWS provider adding a request id to the error which wasn't being removed in this case prior to update the record status.


makslion (approved):

makslion:
Looks good. In case there are other errors that could cause the same bug or in the future someone might forget to sanitize the error should we also clean it up [here](https://github.com/Kuadrant/dns-operator/blob/10aeef82a0423dbb6211219bb42f01b01974a041/internal/controller/dnsrecord_controller.go#L429)?

PR #221, opened 2024-09-05
feat: Add dns record delete validation

### LINKED ISSUE #181: Improve DNSRecord deletion to better ensure changes are persisted
**What**

Update the dnsrecord [deletion logic](https://github.com/Kuadrant/dns-operator/blob/main/internal/controller/dnsrecord_controller.go#L132-L142) to better cover the situation where multiple writes could be happening to the same records.
We currently do a single validation loop after a delete that ensures that the expected result of a delete request is persisted, but this happens very quickly and it is possible, at scale, for other controllers to add back values after this validation attempt.

The logic needs to be changed here to allow longer and more validation attempts after a delete, especially when multiple owners are in play.

Note: This is only really observable when multiple DNSRecords, all contributing to a shared record set, are deleting at the same time.  

Can be re-created by running the `test-e2e-multi` test suite on AWS:
```
make local-setup DEPLOY=true DEPLOYMENT_SCOPE=namespace DEPLOYMENT_COUNT=20
make test-e2e-multi TEST_DNS_MANAGED_ZONE_NAME=dev-mz-aws TEST_DNS_NAMESPACES=dns-operator DEPLOYMENT_COUNT=20
```

mikenairn:
closes #181

[fix e2e multi record test issues](https://github.com/Kuadrant/dns-operator/commit/cf6c6bc65f10f6f3096c8e3c0d92912885ba9bd9)

* Fix an issue with the domain owners checks when multiple instances are being used, allOwners was being appended with multiple duplicates.
* Add additional checks to ensure all remaining records are eventually updated with the correct domain owners.
* Add logging of time taken around dnsrecord create/update/delete checks.

[feat: Add dns record delete validation](https://github.com/Kuadrant/dns-operator/commit/dfb7260bf57d508b8662cfbce505ba7a85c0fe63)

Adds validation around the deletion of DNSRecord resources to reduce the chances of unexpected records being left in the external dns provider (AWS Routes53 etc..) after the record has been garbage collected. It now takes into account the possibly that multiple records could be manipulating the same zone record set simultaneously which can happen if
a significant number of DNSRecord resources were contributing to the same rootHost record set in the provider and were all deleted at roughly the same time.

The deletion logic was changed to always run validation checks, at random intervals, for an amount of time (currently 15 seconds) after the last time a change was detected. A change here can either be that it had to update the provider itself or that some status on the record was changed, the most important one being `Status.DomainOwners`. If either of these changes occur the delete validation loop of 15 seconds is restarted and will only move on to remove the finalizer once it reaches the 15 seconds duration without any changes having been made.

Note: Some providers, such as google, handle this better as they require the API requests to know the current state of the record you are updating/deleting before it will be accepted. Others such as AWS(Route53) unfortunately do not meaning that some things one record just deleted, may get added back by another depending on when it last queried the providers zone records.

**_Validation_**

**Terminal 1:**
Setup local env with multiple instances and watch all dnsrecords: 
```
make local-setup DEPLOY=true DEPLOYMENT_SCOPE=namespace DEPLOYMENT_COUNT=20
kubectl get dnsrecord -w -o json -A | jq '.metadata.namespace + ":" + .metadata.name + " = deletedAt: " + (.metadata.deletionTimestamp | tostring) + " queuedAt: " + (.status.queuedAt | tostring) + " domainOwners: " + (.status.domainOwners | tostring)'
```

**Terminal 2:**
Tail all operator logs: 
```
kubectl stern -l control-plane=dns-operator-controller-manager --all-namespaces
```

**Terminal 3:**
Run multi record e2e with 20 instances (AWS)
```
make test-e2e-multi TEST_DNS_ZONE_DOMAIN_NAME=<your aws domain> TEST_DNS_PROVIDER_SECRET_NAME=dns-provider-credentials-aws TEST_DNS_NAMESPACES=dns-operator DEPLOYMENT_COUNT=20
```
Run multi record e2e with 20 instances (GCP)
```
make test-e2e-multi TEST_DNS_ZONE_DOMAIN_NAME=<your google domain> TEST_DNS_PROVIDER_SECRET_NAME=dns-provider-credentials-gcp TEST_DNS_NAMESPACES=dns-operator DEPLOYMENT_COUNT=20
```

### DIFF
--- a/api/v1alpha1/dnsrecord_types.go
+++ b/api/v1alpha1/dnsrecord_types.go
@@ -112,7 +112,9 @@ type DNSRecordStatus struct {
 	// +optional
 	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
 
-	// QueuedAt is a time when DNS record was received for the reconciliation
+	// QueuedAt is a time when DNS record was received for the reconciliation.
+	// During deletion QueuedAt is used to track the last time a change was detected when processing the delete request.
+	// Used for validation purposes to ensure the record is completely removed from the provider.
 	QueuedAt metav1.Time `json:"queuedAt,omitempty"`
 
 	// QueuedFor is a time when we expect a DNS record to be reconciled again
--- a/bundle/manifests/dns-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/dns-operator.clusterserviceversion.yaml
@@ -38,7 +38,7 @@ metadata:
     capabilities: Basic Install
     categories: Integration & Delivery
     containerImage: quay.io/kuadrant/dns-operator:latest
-    createdAt: "2024-09-03T15:38:08Z"
+    createdAt: "2024-09-10T10:36:47Z"
     description: A Kubernetes Operator to manage the lifecycle of DNS resources
     operators.operatorframework.io/builder: operator-sdk-v1.33.0
     operators.operatorframework.io/project_layout: go.kubebuilder.io/v4
--- a/bundle/manifests/kuadrant.io_dnsrecords.yaml
+++ b/bundle/manifests/kuadrant.io_dnsrecords.yaml
@@ -470,8 +470,10 @@ spec:
                   of this record.
                 type: string
               queuedAt:
-                description: QueuedAt is a time when DNS record was received for the
-                  reconciliation
+                description: |-
+                  QueuedAt is a time when DNS record was received for the reconciliation.
+                  During deletion QueuedAt is used to track the last time a change was detected when processing the delete request.
+                  Used for validation purposes to ensure the record is completely removed from the provider.
                 format: date-time
                 type: string
               queuedFor:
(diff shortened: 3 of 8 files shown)


api/v1alpha1/dnsrecord_types.go
```diff
@@ -112,7 +112,9 @@ type DNSRecordStatus struct {
 	// +optional
 	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
 
-	// QueuedAt is a time when DNS record was received for the reconciliation
+	// QueuedAt is a time when DNS record was received for the reconciliation.
+	// During deletion QueuedAt is used to track the last time a change was detected when processing the delete request.
+	// Used for validation purposes to ensure the record is completely removed from the provider.
 	QueuedAt metav1.Time `json:"queuedAt,omitempty"`
```
mikenairn, line 118: I'm not really sure about re-using this for a different purpose, but it didn't seem like a great idea adding yet another time filed just for tracking the delete updates.

PR #226, opened 2024-09-11
tests: Use unique zone hostnames for each integration test

mikenairn:
The integration test suite uses the inmemory provider which is shared across all test specs. This works fine if a record is created and deleted successfully very quickly as the records in the provider(memory) will be removed, but if for any reason they aren't (secret gets removed before the record has completed its deletion reconcile for instance) they can be left behind and cause odd issues with other tests that are using the same hostname.

The changes here update the integration test suite to:
* Use a generated zone/hostname unique for each single test, this is similar in approach to the current e2e tests.
* Remove all cleanup logic from the integration test suite for record and secrets(providers). Alternative would be to wait for each deletion to complete before allowing it to continue but this could add significant amount of time to each test for little gain. Tests that explicitly test the deletion logic already exist in the integration and e2e suites.

### DIFF
--- a/internal/controller/dnsrecord_controller_test.go
+++ b/internal/controller/dnsrecord_controller_test.go
@@ -20,6 +20,8 @@ package controller
 
 import (
 	"context"
+	"fmt"
+	"strings"
 	"time"
 
 	. "github.com/onsi/ginkgo/v2"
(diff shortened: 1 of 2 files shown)


maleck13 (approved):
/lgtm

PR #227, opened 2024-09-11
chore: Remove unused field "queuedFor" from record status

mikenairn:
QueuedFor was only being set once in the status at the time of adding the finalizer and never updated again. Appears to be not used anywhere for anything and is meaningless as information in the status.

If we want to keep this for some reason, and instead have it get updated somewhere, I'll close this and create an issue for it.


philbrookes (approved):

makslion:
The `queueedFor` is and always was redundant. We don't base any decisions based on that time yet still bother to update it. At the moment of creation, this field was indented to complement the `validFor` and `queuedAt` for:
1. Testing. When dealing with lots of records through automation (not yet happening, but we had QE in mind) we would have access to information about the validity and reconciliation that happened and when we expect the record. 
2. Observability. Again, not yet there, but that piece of information was a candidate to be exposed to use in dashboards. 
If none of those use cases are of concern any more removing `queuedFor` would make the `continuous reconciliation` easier to understand

PR #231, opened 2024-09-13
add perms and docs for azure

maleck13:
Adds specific perms needed for Azure and AWS


docs/provider.md
```diff
@@ -34,6 +34,55 @@ We have tested using the available policy `AmazonRoute53FullAccess` however it s
...
+            "Sid": "VisualEditor1",
+            "Effect": "Allow",
+            "Action": [
+                "route53:ListHostedZones"
+            ],
+            "Resource": "*"
+        }
+    ]
+}
+```
```
mikenairn, line 83: Presumably after we drop support for provider health checks we won't really need those HealthCheck permissions? Should we keep them?

mikenairn (approved):
Looks fine.

docs/provider.md
maleck13: I will add a note to the health check epic to update the needed perms

Issue #222, opened 2024-09-09
Multiple DNS Records e2e test for DNS Operator


philbrookes:
Create many DNS Records at the same time, to see how the DNS Operator handles the load. This should be optional and able to create upto 64 DNS Records at once, all configured with GEO and Weighted load balancing.

PR #233, opened 2024-09-16
Concurrent DNS Records test

philbrookes:
Add a test to allow us to time how long it takes for a single threaded controller to bring N records to ready when they are created more-or-less simultaneously.

closes: https://github.com/Kuadrant/dns-operator/issues/222

```
64 records:
  Azure: failed at an hour
  GCP: 3m57s
  AWS: 6m7s
37 records (the most Azure did under an hour in my testing):
  Azure: 51m37s
  GCP: 2m16s
  AWS: 3m11s
```

### DIFF
--- a/test/e2e/single_record_test.go
+++ b/test/e2e/single_record_test.go
@@ -4,7 +4,9 @@ package e2e
 
 import (
 	"context"
+	"fmt"
 	"net"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


philbrookes:
closes: #222

mikenairn (approved):
I'd like to change this slightly and make it part of the "multi record" test suite that is intended to allow variables to influence the tests and setup different scenarios in a similar way to the current "distributed dns" multi record [tests](https://github.com/Kuadrant/dns-operator/blob/main/test/e2e/multi_record_test.go) and can be run as part of a different test suite `make test-e2e-multi`.

That doesn't need to be done in this PR though, looks fine to merge as is.

PR #235, opened 2024-09-16
gh-232: traffic manager profile hash generation update

### LINKED ISSUE #232: Azure traffic manager profile hash is not sufficiently unique
The hash generated for traffic manager profiles is not being calculated correctly and as such is not sufficiently unique.

philbrookes:
closes #232

### DIFF
--- a/internal/provider/azure/azure.go
+++ b/internal/provider/azure/azure.go
@@ -3,7 +3,6 @@ package azure
 import (
 	"context"
 	"crypto/md5"
-	"encoding/hex"
 	"errors"
 	"fmt"
 	"strconv"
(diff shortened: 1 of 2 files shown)


internal/provider/azure/azure.go
```diff
@@ -500,6 +499,12 @@ func FindDefaultGeoTarget(endpoints []*externaldnsendpoint.Endpoint) string {
 	return ""
 }
 
+func GenerateProfileName(resourceGroup string, ep *externaldnsendpoint.Endpoint) string {
+	data := []byte(ep.DNSName)
```
mikenairn, line 503: Is DNSName unique enough? Multiple endpoints can have the same DNSName depending on how the provider works.

mikenairn (commented):
I'm not sure i totally understand the difference between what it was doing and what it is doing now and how it wasn't unique enough before based on the DNSName. Going by the test you added i assume it was able to generate the same name for different DNSName inputs before?

Needs a rebase also.

internal/provider/azure/azure.go
philbrookes: As DNS name is the host of the CNAME that will point to the traffic manager profile, I'd have thought if that same hostname shows up again somewhere else, it would need to resolve to the same traffic manager profile?

philbrookes:
> I'm not sure i totally understand the difference between what it was doing and what it is doing now and how it wasn't unique enough before based on the DNSName. Going by the test you added i assume it was able to generate the same name for different DNSName inputs before?
> 
> Needs a rebase also.

yeah, before it was always generating the same 2 hashes, I suspect I had generated the hashes from the pointers somehow rather than the actual values.

internal/provider/azure/azure.go
mikenairn: true, so this is a name that we want linked to the dnsname i guess

mikenairn (approved):

PR #236, opened 2024-09-17
custom test timeouts for specific providers (Azure)

philbrookes:


test/e2e/multi_record_test.go
```diff
@@ -824,7 +824,7 @@ var _ = Describe("Multi Record Test", Labels{"multi_record"}, func() {
 					err := tr.cluster.k8sClient.Get(ctx, client.ObjectKeyFromObject(tr.record), tr.record)
 					g.Expect(err).To(MatchError(ContainSubstring("not found")))
 				}
-			}, recordsRemovedMaxDuration, 5*time.Second, ctx).Should(Succeed())
+			}, GetRecordsDeletedTimeout(testDNSProvider), 5*time.Second, ctx).Should(Succeed())
 			GinkgoWriter.Printf("[debug] all records removed in %v\n", time.Since(checkStarted))
 
 			By("checking provider zone records are all removed")
```
mikenairn, line 830: Rather than calling the function all over the place you might be better to just set the values at the start [here](https://github.com/Kuadrant/dns-operator/blob/f7aa99474d84888eb6bf45cad229b996b1dd5cb2/test/e2e/multi_record_test.go#L45C2-L45C27) 

```
recordsReadyMaxDuration := GetRecordsReadyTimeout(testDNSProvider)
recordsRemovedMaxDuration := GetRecordsDeletedTimeout(testDNSProvider)
```

All the [log statements](https://github.com/Kuadrant/dns-operator/blob/f7aa99474d84888eb6bf45cad229b996b1dd5cb2/test/e2e/multi_record_test.go#L130) would be printing the correct thing that way.

mikenairn:
Looks good, small suggestion, up to you.

mikenairn (approved):

test/e2e/multi_record_test.go
philbrookes: yeah, that's neater. I've updated it now.

test/e2e/healthcheck_test.go
```diff
@@ -35,6 +35,9 @@ var _ = Describe("Health Check Test", Serial, Labels{"health_checks"}, func() {
 
 	var dnsRecord *v1alpha1.DNSRecord
 
+	recordsReadyMaxDuration := GetRecordsReadyTimeout(testDNSProvider)
+	recordsRemovedMaxDuration := GetRecordsDeletedTimeout(testDNSProvider)
```
mikenairn, line 39: You might need to actually call these in a `BeforeX` call to be sure `testDNSProvider` is set(It could be nil here and defaulting to your lower timeouts for Azure). Since these are global, you could just set these variables once in `BeforeSuite` after the point where we know `testDNSProvider` has been [set](https://github.com/Kuadrant/dns-operator/blob/main/test/e2e/suite_test.go#L101).

mikenairn (approved):
I think we really might want to consider if we want the full e2e test suite running for Azure on all PRs:

image

PR #250, opened 2024-09-25
ci: Fix image build jobs to use ref instead of sha on tags

mikenairn:
Update the image build jobs to ensure the bundle references the operator image, and the catalog references the bundle image using the tag for the git ref_name(main, v0.9.0, release-0.9 etc..) rather than the commit sha.

Before:
```
docker cp $(docker create --name temp-container quay.io/kuadrant/dns-operator-bundle:v0.6.0 tail):/manifests dns-operator-v0.6.0 && docker rm temp-container
cat dns-operator-v0.6.0/dns-operator.clusterserviceversion.yaml | yq .spec.install.spec.deployments[0].spec.template.spec.containers[0].image
quay.io/kuadrant/dns-operator:a1cc37d47d1980e9b2187587afcdb1cfb2a4bac0
```

After:
```
docker cp $(docker create --name temp-container quay.io/kuadrant/dns-operator-bundle:ci_fix_image_build_job tail):/manifests dns-operator-ci_fix_image_build_job && docker rm temp-container
cat dns-operator-ci_fix_image_build_job/dns-operator.clusterserviceversion.yaml | yq .spec.install.spec.deployments[0].spec.template.spec.containers[0].image
quay.io/kuadrant/dns-operator:ci_fix_image_build_job
```

### DIFF
--- a/.github/workflows/build-images.yaml
+++ b/.github/workflows/build-images.yaml
@@ -11,6 +11,7 @@ on:
 
 env:
   IMG_TAGS: ${{ github.sha }} ${{ github.ref_name }}
+  IMG_REF: ${{ github.sha }}
   IMG_REGISTRY_HOST: quay.io
   IMG_REGISTRY_ORG: kuadrant
   IMG_REGISTRY_REPO: dns-operator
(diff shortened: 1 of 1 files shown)


eguzki (approved):
LGTM

The only minor thing (that is why I was suggesting to update default `IMG_TAGS` env var) is that for tags, the workflow will still be generating docker tags with git sha, which is needless. But good enough. Feel free to fix it or leave it as it is.

mikenairn:
> LGTM
> 
> The only minor thing (that is why I was suggesting to update default `IMG_TAGS` env var) is that for tags, the workflow will still be generating docker tags with git sha, which is needless. But good enough. Feel free to fix it or leave it as it is.

Yes, but your suggestion is only adding the sha back into the operator build tag list when the branch is "main", so would fail on bundle-build for any other branch such as "release-*" where we expect the build stage to add that sha tag, seemed safer to just leave it in all the time. Other option would be to remove it from the IMG_TAGS based on the it being a tag :shrug: .

PR #244, opened 2024-09-24
metrics: Instrumented provider http clients

mikenairn:
closes #206 

Updates all dns provders to use a http client wrapped with prometheus client middleware to add generic http metrics for all requests made to external dns provider platforms (aws, gcp, azure).

Metrics produced are:

```
# HELP http_client_in_flight_requests A gauge of in-flight requests for the wrapped client.
# TYPE http_client_in_flight_requests gauge
http_client_in_flight_requests{client="aws"} 0
# HELP http_client_request_latency_seconds A histogram of request latencies.
# TYPE http_client_request_latency_seconds histogram
http_client_request_latency_seconds_bucket{client="aws",code="200",method="get",le="0.005"} 0
...
http_client_request_latency_seconds_bucket{client="aws",code="200",method="get",le="+Inf"} 157
http_client_request_latency_seconds_sum{client="aws",code="200",method="get"} 31.374643888999994
http_client_request_latency_seconds_count{client="aws",code="200",method="get"} 157
...
# HELP http_client_requests_total A counter for requests from the wrapped client.
# TYPE http_client_requests_total counter
http_client_requests_total{client="aws",code="200",method="delete"} 3
http_client_requests_total{client="aws",code="200",method="get"} 157
http_client_requests_total{client="aws",code="200",method="post"} 30
http_client_requests_total{client="aws",code="201",method="post"} 3
http_client_requests_total{client="aws",code="400",method="post"} 14
http_client_requests_total{client="aws",code="403",method="get"} 12
http_client_requests_total{client="aws",code="404",method="get"} 2
```

Additional changes:
* Removes the dependency on `github.com/linki/instrumented_http` in favour of just using the[ prometheus http middleware](https://github.com/prometheus/client_golang/blob/86f34968685552f0993747630ba5be1ab2f3e3cd/prometheus/promhttp/instrument_client.go).
* Removes the old, unused, AWS instrumented client code.

Metrics output from an e2e test run for all providers(aws, google,azure) against a single instance: https://gist.github.com/mikenairn/4fc6f3437ea4af07e55eb3a018af78b9

Related [PR](https://github.com/Kuadrant/dns-operator/pull/246) to archive dns operator metrics at end of e2e run.

PR #260, opened 2024-10-02
adding probe worker

philbrookes:
Adding probe reconciler behind a flag
Adding probe worker
Adding probe CRD.


internal/probes/main.go
```diff
@@ -0,0 +1,224 @@
...
+
+func performRequest(ctx context.Context, protocol, host, path, ip string, port int, allowInsecure bool, headers v1alpha1.AdditionalHeaders) ProbeResult {
+	probeClient := instrumented_http.NewClient(&http.Client{
+		Transport: TransportWithDNSResponse(map[string]string{host: ip}, allowInsecure),
+	}, &instrumented_http.Callbacks{
+		PathProcessor: func(path string) string {
+			parts := strings.Split(path, "/")
+			return parts[len(parts)-1]
+		},
+	})
```
mikenairn, line 136: I removed the dependency on `github.com/linki/instrumented_http` in favour of just using the prometheus http middleware in [this PR]( https://github.com/Kuadrant/dns-operator/pull/244). Consider using [this](https://github.com/Kuadrant/dns-operator/blob/main/internal/metrics/http.go#L39) instrumented http client instead.

mikenairn: If you do intend to use this, you will also need to register it with the controller runtime metrics somewhere:

```
metrics.Registry.MustRegister(instrumented_http.RequestDurationSeconds)
```

philbrookes: I've updated it as you suggested, so I don't need to do that now, right?

mikenairn: > I've updated it as you suggested, so I don't need to do that now, right?

No

mikenairn: This will create a single set of metrics for all probes which might not be that useful. I'd go with it just now though, we can always modify the client to make the metrics it's producing more configurable since presumably you would want them keyed off the host/path/ip to be anyway useful.

philbrookes: yes, I considered adding a host and endpoint to the name, but when I looked at your names they were all quite short, and I wasn't sure if they needed to stay short or not.

api/v1alpha1/dnshealthcheckprobe_types.go
```diff
@@ -97,14 +97,25 @@ type DNSHealthCheckProbeList struct {
 	Items           []DNSHealthCheckProbe `json:"items"`
 }
 
+func (p *DNSHealthCheckProbe) GetIPAddresses() []string {
+	if net.ParseIP(p.Spec.Address) != nil {
+		return []string{
+			p.Spec.Address,
+		}
+	}
+	ips, _ := net.LookupHost(p.Spec.Address)
```
maleck13, line 106: why do we ignore an error here?

api/v1alpha1/dnshealthcheckprobe_types.go
```diff
@@ -31,9 +31,9 @@ type DNSHealthCheckProbeSpec struct {
 	// Hostname is the value sent in the host header, to route the request to the correct service
 	// +kubebuilder:validation:Pattern=`^[a-z][a-z0-9\-]+\.([a-z][a-z0-9\-]+\.)*[a-z][a-z0-9\-]+$`
 	Hostname string `json:"hostname,omitempty"`
```
maleck13, line 33: should these be required?

api/v1alpha1/dnshealthcheckprobe_types.go
```diff
@@ -97,14 +97,25 @@ type DNSHealthCheckProbeList struct {
...
+
 func (p *DNSHealthCheckProbe) Default() {
 	if p.Spec.Protocol == "" {
 		p.Spec.Protocol = HttpProtocol
 	}
 }
 
 func (p *DNSHealthCheckProbe) ToString() string {
-	return fmt.Sprintf("%v://%v:%v/%v", p.Spec.Protocol, p.Spec.Hostname, p.Spec.Port, p.Spec.Path)
+	return fmt.Sprintf("%v://%v:%v%v", p.Spec.Protocol, p.Spec.Hostname, *p.Spec.Port, p.Spec.Path)
```
maleck13, line 118: `p.Spec.Port` could nil pointer panic here right?

api/v1alpha1/dnshealthcheckprobe_types.go
maksymvavilov: They are. Unless you use `+optional` it thinks it is required.  [Provider ref ](https://github.com/Kuadrant/dns-operator/blob/main/api/v1alpha1/dnsrecord_types.go#L95-L96), for example.

api/v1alpha1/dnshealthcheckprobe_types.go
maksymvavilov: This whole function seems to not be needed. It can be replaced just be `net.LookupHost()` call (it checks for address to be IP inside). Not sure why it was decided to ignore error. I see not harm in checking for it.

api/v1alpha1/dnshealthcheckprobe_types.go
maksymvavilov: It could be. I am not sure why we need a port to be a pointer - we only read value anyway. And it being a pointer makes kubebuilder validation useless. Changed it to just int

api/v1alpha1/dnshealthcheckprobe_types.go
```diff
@@ -42,13 +41,13 @@ type DNSHealthCheckProbeSpec struct {
...
 	// Interval defines how frequently this probe should execute
-	Interval time.Duration `json:"interval,omitempty"`
+	Interval metav1.Duration `json:"interval,omitempty"`
 	// AdditionalHeadersRef refers to a secret that contains extra headers to send in the probe request, this is primarily useful if an authentication
 	// token is required by the endpoint.
 	AdditionalHeadersRef *AdditionalHeadersRef `json:"additionalHeadersRef,omitempty"`
 	// FailureThreshold is a limit of consecutive failures that must occur for a host to be considered unhealthy
 	// +kubebuilder:validation:XValidation:rule="self > 0",message="Failure threshold must be greater than 0"
-	FailureThreshold *int `json:"failureThreshold,omitempty"`
+	FailureThreshold int `json:"failureThreshold,omitempty"`
```
maleck13, line 50: same here the default will be 0

api/v1alpha1/dnshealthcheckprobe_types.go
```diff
@@ -104,7 +103,7 @@ func (p *DNSHealthCheckProbe) Default() {
 }
 
 func (p *DNSHealthCheckProbe) ToString() string {
-	return fmt.Sprintf("%v://%v:%v/%v", p.Spec.Protocol, p.Spec.Hostname, p.Spec.Port, p.Spec.Path)
+	return fmt.Sprintf("%v://%v:%v%v", p.Spec.Protocol, p.Spec.Hostname, p.Spec.Port, p.Spec.Path)
```
maleck13, line 106: so here we will end up with "https://host:0/path" I think?

api/v1alpha1/dnshealthcheckprobe_types.go
```diff
@@ -27,13 +26,13 @@ import (
 type DNSHealthCheckProbeSpec struct {
 	// Port to connect to the host on. Must be either 80, 443 or 1024-49151
 	// +kubebuilder:validation:XValidation:rule="self in [80, 443] || (self >= 1024 && self <= 49151)",message="Only ports 80, 443, 1024-49151 are allowed"
-	Port *int `json:"port,omitempty"`
+	Port int `json:"port,omitempty"`
```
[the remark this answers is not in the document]
maksymvavilov: The CEL validation will not allow for those values in the CR. It will allow only ports 80, 443 and from 1024 to 49151

api/v1alpha1/dnshealthcheckprobe_types.go
maksymvavilov: This will only alow integers grater than 0

api/v1alpha1/dnshealthcheckprobe_types.go
maksymvavilov: we will always end up with a valid address here. It even checks for the `path` to start with `/` or `?`. So something like `https://host:80health` won't be allowed.

internal/probes/main.go
```diff
@@ -0,0 +1,216 @@
...
+		} else if err != nil {
+			logger.Info("could not find probe CR, killing related worker")
+			return false
+		}
+	}
+
+	result := performProbe(ctx, k8sClient, probe)
+
+	patchFrom := client.MergeFrom(probe.DeepCopy())
+	if !result.Healthy {
```
maleck13, line 78: could this be done in the performProbe function and just leave the update here?
```
if !result.Healthy {
		probe.Status.ConsecutiveFailures += 1
	} else {
		probe.Status.ConsecutiveFailures = 0
	}
	probe.Status.Healthy = &result.Healthy
	probe.Status.LastCheckedAt = result.CheckedAt
	probe.Status.Reason = result.Reason
```

maleck13: eh maybe it works better as is... ignore

internal/probes/main.go
```diff
@@ -0,0 +1,216 @@
...
+		err := k8sClient.Get(ctx, client.ObjectKeyFromObject(originalProbe), probe)
+		if client.IgnoreNotFound(err) != nil {
+			logger.Error(err, "error reading probe")
+			return true
+		} else if err != nil {
+			logger.Info("could not find probe CR, killing related worker")
+			return false
+		}
+
+		timeUntilProbe = time.Until(probe.Status.LastCheckedAt.Time.Add(probe.Spec.Interval.Duration))
```
maleck13, line 53: what happens here if the pobe has no status.LastCheckedAt?

internal/probes/main.go
```diff
@@ -0,0 +1,216 @@
...
+			logger.V(1).Info("probe ready to execute", "timeUntilProbe", timeUntilProbe, "lastChecked", probe.Status.LastCheckedAt, "interval", probe.Spec.Interval, "now", time.Now())
+			break
+		}
+
+		logger.V(1).Info("sleeping until probe is ready", "timeUntilProbe", timeUntilProbe, "lastChecked", probe.Status.LastCheckedAt, "interval", probe.Spec.Interval, "now", time.Now())
+		time.Sleep(timeUntilProbe)
+	}
+
+	//if we slept for a while, refresh the probe from the api
+	if timeUntilProbe > time.Second*3 {
```
maleck13, line 64: I am not getting this bit, it we sleep it is in the for loop so the probe will get refreshed next time around right?

maleck13: we are not exiting the loop after the sleep so no need to refresh I mean

internal/probes/main.go
maksymvavilov: It will default to the first of January year 1

maleck13:
replaced by https://github.com/Kuadrant/dns-operator/pull/273

PR #266, opened 2024-10-08
add tooling to convert endpoints to a tree and back

philbrookes:
add tooling to convert endpoints to a tree and back

### DIFF
--- a/internal/common/tree.go
+++ b/internal/common/tree.go
@@ -1,0 +1,208 @@
+package common
+
+import (
+	"slices"
+
+	"sigs.k8s.io/external-dns/endpoint"
+
+	"github.com/kuadrant/dns-operator/api/v1alpha1"
+)
+
+// DNSTreeNode stores a relation between endpoints that were parsed into a tree
+type DNSTreeNode struct {
+	Name     string
+	Children []*DNSTreeNode
+	DataSets []DNSTreeNodeData
+}
+
+// DNSTreeNodeData holds a data for the enpoint(s) that correspond to this node
+type DNSTreeNodeData struct {
+	RecordType       string
+	SetIdentifier    string
+	RecordTTL        endpoint.TTL
+	Labels           endpoint.Labels
+	ProviderSpecific endpoint.ProviderSpecific
+	Targets          []string
+}
+
+// RemoveNode removes a node from a tree.
+// If the node was the only child of the parent node,
+// the parent will be removed as well unless the parent is a root node
+func (n *DNSTreeNode) RemoveNode(deleteNode *DNSTreeNode) {
+	// store indexes of dead branches
+	var deadBranches []int
+	var deadDataSets []string
+
+	if deleteNode == nil {
+		return
+	}
+
+	for i, node := range n.Children {
+		if node.Name == deleteNode.Name {
+			n.Children = append(n.Children[:i], n.Children[i+1:]...)
+
+			// we removed child, but need to clean up data sets
+			for j, dataset := range n.DataSets {
+				index := slices.Index(dataset.Targets, deleteNode.Name)
+				if index >= 0 {
+					n.DataSets[j].Targets = append(dataset.Targets[:index], dataset.Targets[index+1:]...)
+				}
+			}
+			return
+		}
+
+		// no children matched, try on each child
+		node.RemoveNode(deleteNode)
+
+		// the removed node was the only child - we have a dead branch (it is a leaf now)
+		// children are nil on leaf node
+		// not checking for it will nuke the whole tree as leafs will be considered dead branches
+		if node.Children != nil && isALeafNode(node) {
+			// store the index. indexes are in ascending order
+			deadBranches = append(deadBranches, i)
+
+			// we can't rely on indexes for data sets, so store node name
+			deadDataSets = append(deadDataSets, node.Name)
+		}
+	}
+
+	// prune dead branches separately from the main for loop.
+	// doing it inside will shift indexes the for loop is iterating through
+	for count, deadBranchIndex := range deadBranches {
+		// after the first removal, all subsequent deadBranchIndexes will de one to high,
+		// but since we have them ascending, we can use count as modifier
+		n.Children = append(n.Children[:deadBranchIndex-count], n.Children[deadBranchIndex-count+1:]...)
+	}
+
+	var healthyDataSets []DNSTreeNodeData
+	// clean up data nodes from dead branches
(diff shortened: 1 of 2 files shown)


internal/common/tree.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+
+}
+
+// GetLeafsTargets returns IP or CNAME of the leafs of a tree.
+// alternatively, it can populate the passed in array with pointers to targets
+func GetLeafsTargets(node *DNSTreeNode, targets *[]string) *[]string {
+	// no children means this is pointing to an IP or a host outside the DNS Record
+	if len(node.Children) == 0 {
+		*targets = append(*targets, node.Name)
```
maleck13, line 63: should ensure targets not nil here

maleck13: should also check node not nil otherwise could end up with panic

maleck13:
@maksymvavilov I think we should check for nils more aggressively and add some nil tests also as there are a lot of pointers and derefrencing pointers to get the value in here so could easilly end up with a panic

maksymvavilov:
> @maksymvavilov I think we should check for nils more aggressively and add some nil tests also as there are a lot of pointers and derefrencing pointers to get the value in here so could easilly end up with a panic

safeguarded all public calls from nil pointers

internal/common/tree.go
```diff
@@ -0,0 +1,201 @@
...
+	if deleteNode == nil {
+		return
+	}
+
+	for i, node := range n.Children {
+		if node.Name == deleteNode.Name {
+			n.Children = append(n.Children[:i], n.Children[i+1:]...)
+
+			// we removed child, but need to clean up data sets
+			for j, dataset := range n.DataSets {
```
maleck13, line 43: so this looks fine but I am a bit confused as to why the data set isn't just attached to the node

internal/common/tree.go
```diff
@@ -0,0 +1,201 @@
...
+				index := slices.Index(dataset.Targets, deleteNode.Name)
+				if index > 0 {
```
maleck13, line 45: what if the index is 0? should this be >=0 ? looks like `index` will return -1 if not present

internal/common/tree.go
maksymvavilov: We can have more datasets than nodes. For example, we can have 3 `klb.testdomain.com` records that target two (`eu.klb.testdomain.com` and `us.klb.testdomain.com`) and by having 3 datasets we know to create 3 enpoints of `klb.tesdomain.com`. 

We could "push" datasets to the child (this will solve the need to search on removal) but we will end up with an intricate logic for converting them to endpoints: for each node, we will need to check each child for the number of datasets. And it doesn't look right to store in a child the data you need to convert the parent into the endpoint :thinking: 
here is what i mean: 
```
endpoints := []externaldns.Endpoint{}

for _, child := range node.Children {

    ToEndpoints(child, endpoints)

    for _, dataSet := child.Datasets {
        endpoints = append(endpoints, externaldnsEnpoints{
            DNSName: node.Name
            Targets: dataSet.Targets
            .....
        }
    }
}
```
or something similar. Will have a Node that has dataset for it's parent.

maleck13: eh perhaps it is something we can do beyond this initial impl would rather get this in for now

internal/common/tree.go
```diff
@@ -0,0 +1,201 @@
...
+		n.Children = append(n.Children[:deadBranchIndex-count], n.Children[deadBranchIndex-count+1:]...)
+	}
+
+	var healthyDataSets []DNSTreeNodeData
+	// clean up data nodes from dead branches
+	for _, dataSet := range n.DataSets {
+
+		// we are dealing with CNAMES only here.
+		// the A record is already removed from datasets
+		if !slices.Contains(deadDataSets, dataSet.Targets[0]) {
```
maleck13, line 81: why targets[0]?

internal/common/tree.go
```diff
@@ -0,0 +1,201 @@
...
+	"slices"
+
+	"sigs.k8s.io/external-dns/endpoint"
+
+	"github.com/kuadrant/dns-operator/api/v1alpha1"
+)
+
+type DNSTreeNode struct {
+	Name     string
+	Children []*DNSTreeNode
```
maleck13, line 13: maybe we could comment on these to say what they are as it is not hugely clear with the current names

maleck13: like what is a dataset in this struct

internal/common/tree.go
maleck13: ok because it is a CNAME record? I am not following how we know it is CNAME though

internal/common/tree.go
```diff
@@ -0,0 +1,201 @@
...
+
+// ToEndpoints transforms a tree into an array of endpoints.
+// The array could be returned or passed in to be populated
+func ToEndpoints(node *DNSTreeNode, endpoints *[]*endpoint.Endpoint) *[]*endpoint.Endpoint {
+	if node == nil || endpoints == nil {
+		return &[]*endpoint.Endpoint{}
+	}
+
+	// no children means this is pointing to an IP or a host outside of the DNS Record
+	if len(node.Children) == 0 {
```
maleck13, line 115: I wonder would it read a little better if this check was done in a function name something like ```isFinalDNSEntry(node *DNSTreeNode)bool```

internal/common/tree.go
```diff
@@ -0,0 +1,201 @@
...
+		return endpoints
+	}
+	targets := []string{}
+	for _, child := range node.Children {
+		targets = append(targets, child.Name)
+		ToEndpoints(child, endpoints)
+	}
+
+	if node.DataSets == nil {
```
maleck13, line 124: its not clear what node.DataSets being nil means here

maleck13 (approved):
some minor things to consider but generally looks good

internal/common/tree.go
maksymvavilov: it was suggested to use topology-like approach :thinking:  we will definitely revisit it

internal/common/tree.go
maksymvavilov: if it will be A record we will return much earlier. I could add a check here but it is redundant

internal/common/tree.go
maksymvavilov: added comments on structs

internal/common/tree.go
maksymvavilov: it is a "prevent panic is the tree was created manually and with a mistake"

PR #243, opened 2024-09-23
add DNSHealthProbe CRD

philbrookes:
- Add a DNS Health Probe CRD
- Add an empty DNS Health Probe controller
- Add runtime flag to disable controller by default.

PR #270, opened 2024-10-11
add DNSHealthProbe CRD

maksymvavilov:
- Add a DNS Health Probe CRD
- Add an empty DNS Health Probe controller
- Add runtime flag to disable controller by default.

rebase of #243


maksymvavilov:
turns out i can force push tp Phil's branch. This is not needed

Issue #151, opened 2024-05-31
Add health checks to DNS Record status


philbrookes:
When a health check is present for a leaf record in a DNS Record, the status (healthy or unhealthy) should be recorded in the DNS Record status. Along with a message explaining why the record either was or was not included in the request to the DNS Provider.

some example explanations:
- The record was included as the health check is unhealthy but stale
- The record was excluded as the health check has failed X times, and is above the fault tolerance level Y
- The record was included even though the health check is failing as no other records exist - in order to avoid an NXDOMAIN response.

PR #271, opened 2024-10-15
Create/Delete HealthCheckProbeCR from DNSRecord Reconciler

maksymvavilov:
Create and delete `DNSHealthProbe` CRs during reconciliation of the DNSRecords. 
This replaced logic in `internal/controller/dnsrecord_healthchecks.go` to create new healthchecks but does not remove the code that would generate old healthchecks.

### DIFF
--- a/internal/controller/dnsrecord_healthchecks.go
+++ b/internal/controller/dnsrecord_healthchecks.go
@@ -2,180 +2,135 @@ package controller
 
 import (
 	"context"
-	"crypto/md5"
 	"fmt"
-	"io"
 	"reflect"
 
-	"k8s.io/apimachinery/pkg/api/meta"
+	"github.com/go-logr/logr"
+	"github.com/hashicorp/go-multierror"
+
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	externaldns "sigs.k8s.io/external-dns/endpoint"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/utils/ptr"
+	controllerruntime "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
 
 	"github.com/kuadrant/dns-operator/api/v1alpha1"
-	"github.com/kuadrant/dns-operator/internal/provider"
+	"github.com/kuadrant/dns-operator/internal/common"
 )
 
-// healthChecksConfig represents the user configuration for the health checks
-type healthChecksConfig struct {
-	Endpoint         string
-	Port             *int64
-	FailureThreshold *int64
-	Protocol         *provider.HealthCheckProtocol
+func (r *DNSRecordReconciler) ReconcileHealthChecks(ctx context.Context, dnsRecord *v1alpha1.DNSRecord, allowInsecureCerts bool) error {
+	logger := log.FromContext(ctx).WithName("healthchecks")
+	logger.Info("Reconciling healthchecks")
+
+	// Probes enabled but no health check spec yet. Nothing to do
+	if dnsRecord.Spec.HealthCheck == nil {
+		return nil
+	}
+
+	desiredProbes := buildDesiredProbes(dnsRecord, common.GetLeafsTargets(common.MakeTreeFromDNSRecord(dnsRecord), ptr.To([]string{})), allowInsecureCerts)
+
+	for _, probe := range desiredProbes {
+		// if one of them fails - health checks for this record are invalid anyway, so no sense to continue
+		if err := controllerruntime.SetControllerReference(dnsRecord, probe, r.Scheme); err != nil {
+			return err
+		}
+		if err := r.ensureProbe(ctx, probe, logger); err != nil {
+			return err
+		}
+	}
+	logger.Info("Healthecks reconciled")
+	return nil
 }
 
-func (r *DNSRecordReconciler) ReconcileHealthChecks(ctx context.Context, dnsRecord *v1alpha1.DNSRecord) error {
-	var results []provider.HealthCheckResult
-	var err error
+// DeleteHealthChecks deletes all v1alpha1.DNSHealthCheckProbe that have ProbeOwnerLabel of passed in DNSRecord
+func (r *DNSRecordReconciler) DeleteHealthChecks(ctx context.Context, dnsRecord *v1alpha1.DNSRecord) error {
+	logger := log.FromContext(ctx).WithName("healthchecks")
+	logger.Info("Deleting healthchecks")
 
-	dnsProvider, err := r.getDNSProvider(ctx, dnsRecord)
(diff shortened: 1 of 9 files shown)


internal/controller/dnsrecord_healthchecks_test.go
```diff
@@ -0,0 +1,181 @@
...
+				ProviderRef: v1alpha1.ProviderRef{
+					Name: dnsProviderSecret.Name,
+				},
+				Endpoints:   getTestLBEndpoints(testHostname),
+				HealthCheck: getTestHealthCheckSpec(),
+			},
+		}
+	})
+
+	It("Should create valid probe CRs and remove them on DNSRecord deletion", func() {
```
maksymvavilov, line 57: Creating a trivial integration check. At this moment we can only verify the creation of CR to our liking and a correct removal. Will expand in tests as the work progresses

internal/controller/dnsrecord_healthchecks.go
```diff
@@ -2,180 +2,161 @@ package controller
...
+		return err
 	}
 
-	previousCondition = meta.FindStatusCondition(dnsRecord.Status.HealthCheck.Conditions, "HealthProbesSynced")
-	if previousCondition != nil {
-		probesCondition = previousCondition
+	for _, probe := range healthProbes.Items {
+		logger.V(1).Info(fmt.Sprintf("Deleting probe: %s", probe.Name))
+		if err := r.Delete(ctx, &probe); err != nil {
+			return err
```
maleck13, line 66: should we gather the errors and attempt a delete on each of them to avoid leavng around ones we could have deleted due to an issue with another probe?

internal/controller/dnsrecord_healthchecks.go
```diff
@@ -2,180 +2,161 @@ package controller
...
-			spec := provider.HealthCheckSpec{
-				Id:               endpointId,
-				Name:             fmt.Sprintf("%s-%s-%s", dnsRecord.Spec.RootHost, dnsEndpoint.DNSName, address),
-				Host:             &dnsRecord.Spec.RootHost,
-				Path:             config.Endpoint,
-				Port:             config.Port,
-				Protocol:         config.Protocol,
-				FailureThreshold: config.FailureThreshold,
-			}
+	desiredProbes := buildDesiredProbes(dnsRecord, common.GetLeafsTargets(common.MakeTreeFromDNSRecord(dnsRecord), ptr.To([]string{})), allowInsecureCerts)
```
maleck13, line 32: This was confusing me a little, It is fine, but I think it highlights a change needed in the tree code

```
if targets == nil{
    targets = ptr.To([]string{})
}
```

internal/controller/dnsrecord_healthchecks.go
```diff
@@ -2,180 +2,161 @@ package controller
...
-
-	return &healthChecksConfig{
-		Endpoint:         dnsRecord.Spec.HealthCheck.Path,
-		Port:             &port,
-		FailureThreshold: &failureThreshold,
-		Protocol:         (*provider.HealthCheckProtocol)(&dnsRecord.Spec.HealthCheck.Protocol),
+func buildDesiredProbes(dnsRecord *v1alpha1.DNSRecord, leafs *[]string, allowInsecureCerts bool) []*v1alpha1.DNSHealthCheckProbe {
+	var probes []*v1alpha1.DNSHealthCheckProbe
+
+	for _, leaf := range *leafs {
```
maleck13, line 99: should do a nil pointer check

internal/controller/dnsrecord_healthchecks.go
```diff
@@ -2,180 +2,161 @@ package controller
...
+			// the name of the probe will be fine - it has a limit of 253
+			shortHost := strings.Join(strings.Split(record.Name, ".")[:3], ".")
+
+			// if this the case we can reduce just host
+			if len(record.Name)-len(shortHost) > overshoot {
+				value = fmt.Sprintf("%s_%s", record.Namespace, shortHost)
+			} else {
+				// we can't deal with it just by shortening one of them
+				// default to UID of record
+				value = record.GetUIDHash()
```
maleck13, line 157: why don't we just use this all the time?

internal/controller/dnsrecord_healthchecks_test.go
```diff
@@ -0,0 +1,181 @@
...
+		//Create default test dnsrecord
+		Expect(k8sClient.Create(ctx, dnsRecord)).To(Succeed())
+		Eventually(func(g Gomega) {
+			err := k8sClient.Get(ctx, client.ObjectKeyFromObject(dnsRecord), dnsRecord)
+			g.Expect(err).NotTo(HaveOccurred())
+			g.Expect(dnsRecord.Status.Conditions).To(
```
maleck13, line 63: do we intend to have a condition here that reflect the health check probe has successfully been created?

maleck13 (commented):
couple of questions but is coming together well

internal/controller/dnsrecord_healthchecks.go
maksymvavilov: Yes, it isn't pretty and I will revisit this part once we have it working. If this would comfort - we will never have a nil pointer with the current [implementation](https://github.com/Kuadrant/dns-operator/blob/main/internal/common/tree.go#L94)

internal/controller/dnsrecord_healthchecks.go
maksymvavilov: Hmm, at first glance I thought that it would never happen, but, in fact, it could if we have nil for the DNS record. Added a check

internal/controller/dnsrecord_healthchecks_test.go
maksymvavilov: [yes](https://github.com/Kuadrant/dns-operator/issues/151). I will make there a healthy condition and it will be affecting ready condition as well

internal/controller/dnsrecord_healthchecks.go
maksymvavilov: > actually the pobe will always be in the same namespace as the record. The record name has to be unique per namespace. So the label can just be the name of the DNSRecord it came from right? Don't need namespace here ?

if we have multiple namespaces and the same names of records among them this might case a problem for the controller. Right now I'm [restricting](https://github.com/Kuadrant/dns-operator/pull/271/files#diff-d5275729eb4df664c962c5be81ebe5010e601471d12429fcc42376d6d440a142R59) it to a namespace but I'd rather have it safe and have those label values be as unique as we can

maksymvavilov: however, if that makes no sense I'd rather have this logic as simple as possible (either name of record or uid if the name is too long). I might be just too paranoid :thinking:

maleck13: I think just go with the hash as the simplest option for the controller to use and add an additional label of the name for the human? When you do the list you could also restrict to namespace. So the list of probes is always done in the same namespace as the policy and record

maksymvavilov: changed

PR #273, opened 2024-10-17
Updated worker code

maleck13:
updated worker model allowing for a  new worker per generation of the probe, dealing with context being closed and no longer needing to read the probe constantly 

@maksymvavilov 

Verify via creating DNSHealthCheckProbe, editing it and deleting it

tests being added now, just wanted to push this to get feedback on the new model


internal/controller/dnshealthcheckprobe_reconciler.go
```diff
@@ -29,26 +44,86 @@ func (r *DNSProbeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
 	logger := baseLogger
 
-	logger.Info("TODO Reconciling DNSHealthCheckProbe")
+	logger.Info("Reconciling DNSHealthCheckProbe")
 
 	previous := &v1alpha1.DNSHealthCheckProbe{}
 	err := r.Client.Get(ctx, client.ObjectKey{Namespace: req.Namespace, Name: req.Name}, previous)
 	if err != nil {
 		if err = client.IgnoreNotFound(err); err == nil {
 			return ctrl.Result{}, nil
```
maleck13, line 53: seems odd to return nil here

maksymvavilov: We could return an error, but the idea is that DNSRecord controller will [make sure](https://github.com/Kuadrant/dns-operator/pull/271/files#diff-d5275729eb4df664c962c5be81ebe5010e601471d12429fcc42376d6d440a142R72) probes exist as long as DNSRecord exist. And this controller only acts on the presence of the CR and does not creates\updates it :thinking: 

That said I can't think of a case when we will reach this code. For this controller to receive a request we must do an action on the existing CR and since you added finalizers it will always be there

PR #277, opened 2024-10-22
consult healtcheck probes when publishing enpoints 

maksymvavilov:
Do not publish unhealthy endpoints or remove them from provider

### DIFF
--- a/internal/controller/dnsrecord_controller_test.go
+++ b/internal/controller/dnsrecord_controller_test.go
@@ -1,4 +1,4 @@
-//go:build integration
+//go:build integration2
 
 /*
 Copyright 2024.
@@ -76,7 +76,7 @@ var _ = Describe("DNSRecordReconciler", func() {
 				ProviderRef: v1alpha1.ProviderRef{
 					Name: dnsProviderSecret.Name,
 				},
-				Endpoints: getTestEndpoints(testHostname, "127.0.0.1"),
+				Endpoints: getTestEndpoints(testHostname, []string{"127.0.0.1"}),
 			},
 		}
 	})
@@ -92,7 +92,7 @@ var _ = Describe("DNSRecordReconciler", func() {
 				},
 				Spec: v1alpha1.DNSRecordSpec{
 					RootHost:    testHostname,
-					Endpoints:   getTestEndpoints(testHostname, "127.0.0.1"),
+					Endpoints:   getTestEndpoints(testHostname, []string{"127.0.0.1"}),
 					HealthCheck: nil,
 				},
 			}
@@ -114,7 +114,7 @@ var _ = Describe("DNSRecordReconciler", func() {
 					ProviderRef: v1alpha1.ProviderRef{
 						Name: dnsProviderSecret.Name,
 					},
-					Endpoints:   getTestEndpoints(testHostname, "127.0.0.1"),
+					Endpoints:   getTestEndpoints(testHostname, []string{"127.0.0.1"}),
 					HealthCheck: nil,
 				},
 			}
@@ -149,7 +149,7 @@ var _ = Describe("DNSRecordReconciler", func() {
 					ProviderRef: v1alpha1.ProviderRef{
 						Name: dnsProviderSecret.Name,
 					},
-					Endpoints: getTestEndpoints("bar.example.com", "127.0.0.1"),
+					Endpoints: getTestEndpoints("bar.example.com", []string{"127.0.0.1"}),
 					HealthCheck: &v1alpha1.HealthCheckSpec{
 						Path:             "health",
 						Port:             5,
@@ -197,7 +197,7 @@ var _ = Describe("DNSRecordReconciler", func() {
 				ProviderRef: v1alpha1.ProviderRef{
 					Name: dnsProviderSecret.Name,
 				},
-				Endpoints:   getTestEndpoints(testHostname2, "127.0.0.2"),
+				Endpoints:   getTestEndpoints(testHostname2, []string{"127.0.0.1"}),
 				HealthCheck: nil,
 			},
 		}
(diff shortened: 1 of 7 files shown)


internal/controller/dnsrecord_healthchecks.go
```diff
@@ -94,6 +95,69 @@ func (r *DNSRecordReconciler) ensureProbe(ctx context.Context, generated *v1alph
 	return nil
 }
 
+// removeUnhealthyEndpoints fetches all probes associated with this record and uses the following criteria while removing endpoints:
+//   - If the Leaf Address has no health check CR - it is healthy
+//   - If the health check CR has insufficient failures - it is healthy
+//   - If the health check CR is deleting - it is healthy
+//   - If the health check is a CNAME and any IP is healthy - the CNAME is healthy
```
maleck13, line 102: Is it also the case that if it is the last endpoint it healthy?

internal/controller/dnsrecord_healthchecks.go
```diff
@@ -94,6 +95,69 @@ func (r *DNSRecordReconciler) ensureProbe(ctx context.Context, generated *v1alph
...
+func (r *DNSRecordReconciler) removeUnhealthyEndpoints(ctx context.Context, endpoints []*endpoint.Endpoint, dnsRecord *v1alpha1.DNSRecord) ([]*endpoint.Endpoint, bool, error) {
```
maleck13, line 103: what is the boolean for?

maksymvavilov: for reporting to status. Changed to string slice for more information in the next PR. I decided to add it here so the next PR will not be changing this file, but I ended up refactoring it anyway. Could remove for this one if this matters

internal/controller/dnsrecord_healthchecks.go
maksymvavilov: Nope, if this would remove all the enpoints we will do nothing. I'll add more comments

maleck13: Discussed. We wont remove records only will remove if multi-value record

internal/controller/dnsrecord_controller_test.go
```diff
@@ -1,4 +1,4 @@
-//go:build integration
+//go:build integration2
```
maleck13, line 1: what is this for? Is something you setup for local use?

internal/controller/dnsrecord_healthchecks.go
```diff
@@ -94,6 +95,72 @@ func (r *DNSRecordReconciler) ensureProbe(ctx context.Context, generated *v1alph
...
 
+// removeUnhealthyEndpoints fetches all probes associated with this record and uses the following criteria while removing endpoints:
+//   - If the Leaf Address has no health check CR - it is healthy
+//   - If the health check CR has insufficient failures - it is healthy
+//   - If the health check CR is deleting - it is healthy
+//   - If the health check is a CNAME and any IP is healthy - the CNAME is healthy
+//
+// If this leads to an empty array of endpoints it:
+//   - Does nothing (prevents NXDomain response) if we already published
+//   - Returns empty array of nothing is published (prevent from publishing unhealthy EPs)
```
maleck13, line 106:
```suggestion
//   - Returns empty array if nothing is published (prevent from publishing unhealthy EPs)
```

internal/controller/dnsrecord_healthchecks.go
```diff
@@ -94,6 +95,72 @@ func (r *DNSRecordReconciler) ensureProbe(ctx context.Context, generated *v1alph
...
+//
+// it returns the list of healthy endpoints, an array of unhealthy addresses and an error
+func (r *DNSRecordReconciler) removeUnhealthyEndpoints(ctx context.Context, specEndpoints []*endpoint.Endpoint, dnsRecord *v1alpha1.DNSRecord) ([]*endpoint.Endpoint, []string, error) {
```
maleck13, line 109: We don't seem to be using the second return []string? Should we just remove it?

internal/controller/dnsrecord_healthchecks.go
```diff
@@ -94,6 +95,72 @@ func (r *DNSRecordReconciler) ensureProbe(ctx context.Context, generated *v1alph
...
+			tree.RemoveNode(&common.DNSTreeNode{
+				Name: probe.Spec.Address,
+			})
+			unhealthyAddresses = append(unhealthyAddresses, probe.Spec.Address)
+		}
+
+	}
+
+	// if at least one of the leaf probes was healthy return healthy probes
+	if haveHealthyProbes {
```
maleck13, line 157:
```suggestion
	if isHealthy {
```

internal/controller/dnsrecord_healthchecks.go
```diff
@@ -94,6 +95,72 @@ func (r *DNSRecordReconciler) ensureProbe(ctx context.Context, generated *v1alph
...
+		return *common.ToEndpoints(tree, ptr.To([]*endpoint.Endpoint{})), unhealthyAddresses, nil
+	}
+	// if none of the probes are healthy or probes don't exist - don't modify endpoints
+	return dnsRecord.Status.Endpoints, unhealthyAddresses, nil
```
maleck13, line 161: right so this is where, if we had an unhealthy record, we would want to validate that there would still be at least one record returned for a possible dns path (IE for a given geo record set or across all record sets (if no geo loadbalancing).

internal/controller/dnsrecord_healthchecks.go
```diff
@@ -94,6 +95,72 @@ func (r *DNSRecordReconciler) ensureProbe(ctx context.Context, generated *v1alph
...
+			})
+			unhealthyAddresses = append(unhealthyAddresses, probe.Spec.Address)
+		}
+
+	}
+
+	// if at least one of the leaf probes was healthy return healthy probes
+	if haveHealthyProbes {
+		return *common.ToEndpoints(tree, ptr.To([]*endpoint.Endpoint{})), unhealthyAddresses, nil
+	}
```
maleck13, line 159: is there value here of adding a status condition to the DNSRecord?
```
if !isHealthy{
conditions.AddCondition(...
{
type: Healthy
Reason: NoHealthChecksPassing
Status: False
Message: Record is unhealthy 
})```
}

maleck13 (commented):
This looks pretty good. I will test it some more tomorrow

PR #274, opened 2024-10-18
ci: Improve build images job for release tags

eguzki:
### What

New CI workflow named `Build and Publish Images For Tag Release` that:

- [X] Skip the bundle build [step](https://github.com/Kuadrant/dns-operator/blob/main/.github/workflows/build-images.yaml#L90)
- [x]  Don't set the BUNDLE_IMG value in the build-catalog [step](https://github.com/Kuadrant/dns-operator/blob/main/.github/workflows/build-images.yaml#L136)
- [x] Ensure current image values reference the same tag name currently being built
- [x] Ensure all image references are equal (operator, bundle, catalog)
- [ ] Golang version check

### Verification steps: create a release that should succeed.

```
git checkout -b release-0.7-testing 262-ci-improve-build-images-job-for-release-tags
```

Run prepare release for final version `v0.0.100`. 
```sh
make prepare-release VERSION=0.0.100 CHANNELS=stable REPLACES_VERSION=0.6.0
```
Verify local changes, commit, push and tag with the typo:
```sh
git add .
git commit -m "prepare-release: v0.7.0"
git tag v0.0.100 
git push origin release-0.7-testing
git push origin v0.0.100 
```
Verify that the build [image for release tag workflow](https://github.com/Kuadrant/dns-operator/actions/workflows/build-images-for-tag-release.yaml) is triggered and completes for the new tag

* Clean up
```
git checkout 262-ci-improve-build-images-job-for-release-tags
git push --delete origin v0.0.100
git tag -d v0.0.100
git push --delete origin release-0.7-testing
git branch -D release-0.7-testing
```

### Verification steps: create a release that should fail.

Let's prepare a release with a typo.

```
git checkout -b release-0.7-testing 262-ci-improve-build-images-job-for-release-tags
```

Run prepare release for final version `v0.0.200`. 
```sh
make prepare-release VERSION=0.0.200 CHANNELS=stable REPLACES_VERSION=0.6.0
```
Verify local changes, commit, push and tag with one typo: instead of tag `v0.0.200` which would be the right one, tag it with `v0.0.201`.
```sh
git add .
git commit -m "prepare-release: v0.7.0"
git tag v0.0.201
git push origin release-0.7-testing
git push origin v0.0.201
```
Verify that the build [image for release tag workflow](https://github.com/Kuadrant/dns-operator/actions/workflows/build-images-for-tag-release.yaml) is triggered and fails

* Clean up

```
git checkout 262-ci-improve-build-images-job-for-release-tags
[...]

PR #278, opened 2024-10-24
ci: on release check version string is the expected one

eguzki:
### What

Follow up work from https://github.com/Kuadrant/dns-operator/pull/274

CI check to ensure release version in go code matches git version tag

Release version passed in by ldflag

### Verification steps: create a release that should succeed.

```
git checkout -b release-0.7-testing ci-update-check-version-string
```

Run prepare release for final version `v0.0.100`. 
```sh
make prepare-release VERSION=0.0.100 CHANNELS=stable REPLACES_VERSION=0.6.0
```

Verify local changes, commit, push and tag with the typo:
```sh
git add .
git commit -m "prepare-release: v0.7.0"
git tag v0.0.100 
git push origin release-0.7-testing
git push origin v0.0.100 
```
Verify that the build [image for release tag workflow](https://github.com/Kuadrant/dns-operator/actions/workflows/build-images-for-tag-release.yaml) is triggered and completes for the new tag

* Clean up
```
git checkout ci-update-check-version-string
git push --delete origin v0.0.100
git tag -d v0.0.100
git push --delete origin release-0.7-testing
git branch -D release-0.7-testing
```

### Verification steps: create a release that should fail.

Let's prepare a release with a typo.

```
git checkout -b release-0.7-testing ci-update-check-version-string
```

Run prepare release for final version `v0.0.200`. 
```sh
make prepare-release VERSION=0.0.200 CHANNELS=stable REPLACES_VERSION=0.6.0
```
Verify local changes, commit, push and tag with one typo: instead of tag `v0.0.200` which would be the right one, tag it with `v0.0.201`.
```sh
git add .
git commit -m "prepare-release: v0.7.0"
git tag v0.0.201
git push origin release-0.7-testing
git push origin v0.0.201
```
Verify that the build [image for release tag workflow](https://github.com/Kuadrant/dns-operator/actions/workflows/build-images-for-tag-release.yaml) is triggered and fails.

Checking CI logs 
```
Git reference name = v0.0.201, release version = v0.0.200
```
So, the step named `Verify git reference name matches the release version` should fail.

* Clean up

```
git checkout ci-update-check-version-string
git push --delete origin v0.0.201
git tag -d v0.0.201
git push --delete origin release-0.7-testing
git branch -D release-0.7-testing
```


Makefile
```diff
@@ -529,7 +543,15 @@ prepare-release: ## Generates a makefile that will override environment variable
 	VERSION=$(VERSION)\nREPLACES_VERSION=$(REPLACES_VERSION)" > $(RELEASE_FILE)
 	$(MAKE) bundle
 	$(MAKE) helm-build VERSION=$(VERSION)
+	$(MAKE) update-release-version VERSION=$(VERSION)
+
+.PHONY: read-version
+read-release-version: ## Reads release version
+	@$(SED) -n 's/^.*Version = "\([^"]*\)".*/\1/p' $(PROJECT_PATH)/internal/version/version.go
```
eguzki, line 550: I am considering having release version being injected into golang code with ld flags like `gitSHA` parameter. Thoughts @mikenairn ? This reading/parsing and updating golang code does not feel robust, it is prone to errors. Also, for reading version set, I did not want to run go code to avoid installing golang toolkit.

maleck13: I think this is a good idea

mikenairn: I have no issue with this. It's just info, and doing it this way will be easier to deal with.

Makefile
```diff
@@ -528,6 +528,9 @@ prepare-release: ## Generates a makefile that will override environment variable
 	$(MAKE) bundle
 	$(MAKE) helm-build VERSION=$(VERSION)
 
+.PHONY: read-release-version
+read-release-version: ## Reads release version
+	@echo "v$(VERSION)"
```
mikenairn, line 533: :+1: Using the single version from the makefile makes sense to me. On release branches we set this as an override as part of the release process https://github.com/Kuadrant/dns-operator/blob/v0.6.0/make/release.mk#L7

mikenairn (approved):

Makefile
eguzki: I know :) 

I think that I will bring this release.mk approach to the other projects

PR #279, opened 2024-10-24
add default duration of 5m

maleck13:

rh-pre-commit.version: 2.2.0


maksymvavilov (approved):
The question is what the value should be: I'd go for ~5 minutes, but definitely not 0 seconds.

maksymvavilov (approved):

PR #283, opened 2024-10-25
add e2e test for healthchecks 

maksymvavilov:
Add a basic e2e test for healtchchecks. I'm splitting it from the rest of the e2e checks and running it with an in-memory provider. This makes it strange to have such a test in the e2e suite - the integration suite (that uses the same provider) mocks controllers and does 99% of the same job. The only difference is that here we actually rely on the probe worker to do a request (to the mock transport layer, but still a real request) compared to manually updating probe CRs in the integration suite 
Also allow for overriding of the transport layer for the HTTP client probe is using.

### DIFF
--- a/config/local-setup/dns-provider/inmemory/kustomization.yaml
+++ b/config/local-setup/dns-provider/inmemory/kustomization.yaml
@@ -10,4 +10,4 @@ secretGenerator:
   - name: dns-provider-credentials
     type: "kuadrant.io/inmemory"
     literals:
-      - INMEM_INIT_ZONES=kuadrant.local
+      - INMEM_INIT_ZONES=e2e.hcpapps.net
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -7,6 +7,18 @@ rules:
 - apiGroups:
   - ""
   resources:
+  - configmaps
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - ""
+  resources:
   - secrets
   verbs:
   - get
--- a/internal/controller/dnshealthcheckprobe_reconciler.go
+++ b/internal/controller/dnshealthcheckprobe_reconciler.go
@@ -40,6 +40,7 @@ type DNSProbeReconciler struct {
 //+kubebuilder:rbac:groups=kuadrant.io,resources=dnshealthcheckprobes,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=kuadrant.io,resources=dnshealthcheckprobes/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=kuadrant.io,resources=dnshealthcheckprobes/finalizers,verbs=update
+//+kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;watch;create;update;patch;delete
 
 func (r *DNSProbeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	baseLogger := log.FromContext(ctx).WithName("dnsprobe_controller")
--- a/internal/probes/worker.go
+++ b/internal/probes/worker.go
@@ -5,8 +5,10 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"strconv"
 	"time"
 
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	utilnet "k8s.io/apimachinery/pkg/util/net"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -203,11 +205,14 @@ func TransportWithDNSResponse(overrides map[string]string, allowInsecureCertific
 
 type ProbeManager struct {
 	probes map[string]context.CancelFunc
+	//main use for this is to configure a test transport
+	overrideTransport bool
 }
 
-func NewProbeManager() *ProbeManager {
+func NewProbeManager(overrideTransport bool) *ProbeManager {
 	return &ProbeManager{
-		probes: map[string]context.CancelFunc{},
+		probes:            map[string]context.CancelFunc{},
+		overrideTransport: overrideTransport,
 	}
 }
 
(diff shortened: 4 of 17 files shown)


config/deploy/local/kustomization.yaml
```diff
@@ -14,6 +14,9 @@ patches:
       - op: add
         path: /spec/template/spec/containers/0/args/-
         value: --zap-log-level=debug
+      - op: add
+        path: /spec/template/spec/containers/0/args/-
+        value: --e2e-test-only-use-mock-health-transport
```
maksymvavilov, line 19: This means that every time we use `local-deploy` or `local-setup` the controller will expect the map for healthchecks to be present. I.E. we will not be able to have probes use a real transport layer when doing a request. 
The only way to have a "real" IP/CNAME probed is to `make run` and have the controller running locally

config/local-setup/dns-provider/inmemory/kustomization.yaml
```diff
@@ -10,4 +10,4 @@ secretGenerator:
   - name: dns-provider-credentials
     type: "kuadrant.io/inmemory"
     literals:
-      - INMEM_INIT_ZONES=kuadrant.local
+      - INMEM_INIT_ZONES=e2e.hcpapps.net
```
mikenairn, line 13: Why change this?

config/rbac/role.yaml
```diff
@@ -4,6 +4,18 @@ kind: ClusterRole
...
+  resources:
+  - configmaps
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
```
mikenairn, line 18: Presumably this is only ever needed for testing? Unsure if this should be added into what will eventually be a real deployments rbac.

internal/controller/dnshealthcheckprobe_reconciler.go
```diff
@@ -40,6 +40,7 @@ type DNSProbeReconciler struct {
 //+kubebuilder:rbac:groups=kuadrant.io,resources=dnshealthcheckprobes,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=kuadrant.io,resources=dnshealthcheckprobes/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=kuadrant.io,resources=dnshealthcheckprobes/finalizers,verbs=update
+//+kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;watch;create;update;patch;delete
```
mikenairn, line 43: Do you really need all these permissions, also see above https://github.com/Kuadrant/dns-operator/pull/283/files#r1827561916

test/e2e/suite_test.go
```diff
@@ -1,4 +1,4 @@
-//go:build e2e
+//go:build e2e || e2e_healthchecks
```
mikenairn, line 1: The e2e test is meant to run on different providers, since these health check tests are only working with the inmemory provider adding them to the normal e2e test suite seems pointless, you will just end up running the same test multiple times.

config/local-setup/dns-provider/inmemory/kustomization.yaml
maksymvavilov: to align with the rest of the providers. It is a "taste" change not a "make it work" one

config/rbac/role.yaml
maksymvavilov: only for testing yes. I could not figure out the way to grant those permissions only when we decide to override the transport layer so I went with the second best :thinking: Would you have an idea how to improve? My thoughts were to use a secret with string data but add rbacs seemed like a better solution. WDYT?

internal/probes/worker.go
```diff
@@ -285,9 +290,47 @@ func (m *ProbeManager) EnsureProbeWorker(ctx context.Context, k8sClient client.C
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "mockhealthcheckresponses",
+				// we cannot guarantee this ns to be present
+				Namespace: "dns-operator-system",
+			},
+		}
+		if err := k8sClient.Get(ctx, client.ObjectKeyFromObject(cm), cm); err != nil {
+			return nil, fmt.Errorf("failed to load test only configmap based transport %w", err)
+		}
+		fmt.Println("CALLED MOCK HELATH CHECK")
```
maleck13, line 321: can remove this

internal/probes/worker.go
```diff
@@ -285,9 +290,47 @@ func (m *ProbeManager) EnsureProbeWorker(ctx context.Context, k8sClient client.C
...
+		if reponseCode, ok := cm.Data[ip]; ok {
+			code, err := strconv.ParseInt(reponseCode, 10, 64)
+			if err != nil {
+				fmt.Println("failed to parse response code from configmap", err)
```
maleck13, line 325: this can be removed

internal/probes/worker.go
```diff
@@ -285,9 +290,47 @@ func (m *ProbeManager) EnsureProbeWorker(ctx context.Context, k8sClient client.C
...
+				return nil, fmt.Errorf("failed to parse response code from configmap %w", err)
+			}
+			return &http.Response{
+				StatusCode: int(code),
+			}, nil
+		}
+		fmt.Println("no response code in configmap for uri", r.URL.Host)
```
maleck13, line 332: and this

test/e2e/healthcheck_test.go
```diff
@@ -47,9 +60,201 @@ var _ = Describe("Health Check Test", Serial, Labels{"health_checks"}, func() {
...
+					}),
+				))
+			}, TestTimeoutLong, time.Second, ctx).Should(Succeed())
+		})
 	})
 
-	Context("On-cluster healthchecks", func() {
-		// TODO test case
+	// multiple controllers against the same host
+	Context("Multiple publishers", func() {
```
maleck13, line 257: do you want to add this as part of this PR?

maleck13 (commented):
Couple of comments, the single A record multiple value test looks good though

config/local-setup/dns-provider/inmemory/kustomization.yaml
mikenairn: It doesn't make sense to me to change this to a domain that we actually use and is managed by route53. Not sure how this aligns with anything, its just a different domain, and one that is set to something obviously not meant for actual external use to avoid confusion.

test/e2e/healthcheck_test.go
maksymvavilov: Don't have a strong opinion. Removed

PR #284, opened 2024-10-25
fix: Multiple same host/target records

mikenairn:
Fixes an issue where if you had multiple records with the same host and target values all contributing to the same record set, removing one of them would cause invalid records (0 target values) to be returned from the plan.

Fixes an issue in the inmemory provider where it was allowing records with 0 target values to be added, and improves checks to avoid a panic.

Add --race option to make run task, enables the race detector on locally running instances https://go.dev/doc/articles/race_detector. This will report useful information in the logs about race related issues.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -239,13 +239,13 @@ build: manifests generate fmt vet ## Build manager binary.
 run: GIT_SHA=$(shell git rev-parse HEAD || echo "unknown")
 run: DIRTY=$(shell hack/check-git-dirty.sh || echo "unknown")
 run: manifests generate fmt vet ## Run a controller from your host.
-	go run -ldflags "-X main.gitSHA=${GIT_SHA} -X main.dirty=${DIRTY}" ./cmd/main.go --zap-devel --provider inmemory,aws,google,azure
+	go run -ldflags "-X main.gitSHA=${GIT_SHA} -X main.dirty=${DIRTY}" --race ./cmd/main.go --zap-devel --provider inmemory,aws,google,azure
 
 .PHONY: run-with-probes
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -239,13 +239,13 @@ build: manifests generate fmt vet ## Build manager binary.
...
 run: manifests generate fmt vet ## Run a controller from your host.
-	go run -ldflags "-X main.gitSHA=${GIT_SHA} -X main.dirty=${DIRTY}" ./cmd/main.go --zap-devel --provider inmemory,aws,google,azure
+	go run -ldflags "-X main.gitSHA=${GIT_SHA} -X main.dirty=${DIRTY}" --race ./cmd/main.go --zap-devel --provider inmemory,aws,google,azure
 
 .PHONY: run-with-probes
 run-with-probes: GIT_SHA=$(shell git rev-parse HEAD || echo "unknown")
 run-with-probes: DIRTY=$(shell hack/check-git-dirty.sh || echo "unknown")
 run-with-probes: manifests generate fmt vet ## Run a controller from your host.
-	go run -ldflags "-X main.gitSHA=${GIT_SHA} -X main.dirty=${DIRTY}" ./cmd/main.go --zap-devel --provider inmemory,aws,google,azure
+	go run -ldflags "-X main.gitSHA=${GIT_SHA} -X main.dirty=${DIRTY}" --race  ./cmd/main.go --zap-devel --provider inmemory,aws,google,azure
```
mikenairn, line 248: Is this supposed to do something different from `make run`? I assume it's missing an extra flag or something.

maleck13: tbh prob can be removed now. The flag is set to true anyway

maleck13 (approved):

PR #288, opened 2024-10-31
update probe logic not to share memory

maleck13:
Running with ```--race``` showed we were sharing some memory across routines. I have adjusted the design not to share memory like this now

rh-pre-commit.version: 2.2.0


maksymvavilov (approved):
looks good!

maleck13:
@maksymvavilov setting ```go run  --zap-log-level=2``` seemed to work well

PR #300, opened 2024-11-07
remove writing out lastCheckedAt status

maleck13:

rh-pre-commit.version: 2.2.0

### DIFF
--- a/internal/probes/worker.go
+++ b/internal/probes/worker.go
@@ -26,10 +26,13 @@ const (
 )
 
 type ProbeResult struct {
+	// CheckedAt the current helath check time
 	CheckedAt metav1.Time
-	Healthy   bool
-	Reason    string
-	Status    int
+	// PreviousCheck the time it was checked before the current check
+	PreviousCheck metav1.Time
+	Healthy       bool
+	Reason        string
+	Status        int
 }
 
 type RoundTripperFunc func(*http.Request) (*http.Response, error)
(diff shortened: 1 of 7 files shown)


internal/probes/worker.go
```diff
@@ -246,11 +252,11 @@ func (w *Probe) Start(clientctx context.Context, k8sClient client.Client, probe
 			} else {
 				freshProbe.Status.ConsecutiveFailures = 0
 			}
+			logger.V(1).Info("health: execution complete ", "result", probeResult, "checked at", probeResult.CheckedAt.String(), "previoud check at ", probeResult.PreviousCheck)
 			freshProbe.Status.Healthy = &probeResult.Healthy
 			freshProbe.Status.LastCheckedAt = probeResult.CheckedAt
```
maksymvavilov, line 257: We probably should not write `LastCheckedAt` to the probe CR anymore

internal/probes/worker.go
```diff
@@ -26,10 +26,13 @@ const (
 )
 
 type ProbeResult struct {
+	// CheckedAt the current helath check time
 	CheckedAt metav1.Time
```
maksymvavilov, line 30: Do we need this value any more? I thought that the only that matters is `LastCheckedAt` in the `localProbe` :thinking:

maksymvavilov (approved):
Seems like it is working 
image
The CR has no `LAstCheckedAt` field even when attempting to write it. No errors as well. 
It might be easier to remove `LastCheckedAt` completely from the probe CRD and have it as a local var. Or it will make things even more complicated :thinking:  
In any case - it is working

maleck13:
I will merge as is and then we can enhance it more afterwards as this will stop a lot of events happening anyway

PR #301, opened 2024-11-07
fix: Namespaced deployment overlay generation

mikenairn:
**fix: Namespaced deployment overlay generation**

* Remove dnshealthcheckprobes CRD for namespaced deployments
* Add crd to cluster kustomization
* Improve dns provider generation, avoid calling a base that requires a local file.

**bump kustomize version v5.5.0**

**fix: Multi record test**

Test was incorrectly checking for "GEO-" in the aws provider specific endpoint values.

Note: This test is executed on the e2e  but only with 1 cluster/deployment so it never uses the second geo, you can re-create the issue by running:

```
make local-setup DEPLOY=true DEPLOYMENT_SCOPE=namespace DEPLOYMENT_COUNT=2 CLUSTER_COUNT=2
make test-e2e-multi TEST_DNS_ZONE_DOMAIN_NAME=mn.hcpapps.net TEST_DNS_PROVIDER_SECRET_NAME=dns-provider-credentials-aws TEST_DNS_NAMESPACES=dns-operator DEPLOYMENT_COUNT=2 CLUSTER_COUNT=2
```

We should probably consider adding a GitHub action that executes it this way.


maleck13:
github action sounds good

config/local-setup/dns-operator/kustomization.yaml
```diff
@@ -8,3 +8,9 @@ patches:
       kind: CustomResourceDefinition
       metadata:
         name: dnsrecords.kuadrant.io
+  - patch: |-
```
maleck13, line 11: Not getting why this is needed?

maleck13 (approved):
lgtm just not sure why we have to remove the probe crd?

config/local-setup/dns-operator/kustomization.yaml
mikenairn: The `../../deploy/local` kustomization which this is modifying adds the crds which prevents you from using it multiple times in an overlay. When we install multiple instances of the dns operator we can use this kustomization which removes the CRDs multiple times, and install the crds separately once.

PR #303, opened 2024-11-13
refactor CEL validation on healthcheck specs

maksymvavilov:


internal/controller/dnsrecord_healthchecks_test.go
```diff
@@ -167,6 +167,64 @@ var _ = Describe("DNSRecordReconciler_HealthChecks", func() {
 		}, TestTimeoutMedium, time.Second).Should(Succeed())
 	})
 
+	It("Should create valid probe CRs with default values", func() {
+		//Create test dnsrecord with nils for optional fields
+		dnsRecord.Spec.HealthCheck = &v1alpha1.HealthCheckSpec{
+			Path: "/health",
```
maksymvavilov, line 173: This is not the "minimal" health spec that we could have in the policy. We use the [same](https://github.com/Kuadrant/kuadrant-operator/blob/main/api/v1/dnspolicy_types.go#L56) spec for the policy and DNSRecord

api/v1alpha1/dnsrecord_types.go
```diff
@@ -36,23 +36,36 @@ const HttpsProtocol Protocol = "HTTPS"
 // the listeners assigned to the target gateway
 type HealthCheckSpec struct {
 	// Port to connect to the host on. Must be either 80, 443 or 1024-49151
+	// Defaults to port 80
```
maleck13, line 39: should we perhaps default to 443 (IE default to secure)?

maleck13 (commented):
changes look fine just a question on whether to default to HTTP or HTTPS for health checks

api/v1alpha1/dnsrecord_types.go
maksymvavilov: Sure. Done

api/v1alpha1/dnsrecord_types.go
```diff
@@ -36,23 +36,36 @@ const HttpsProtocol Protocol = "HTTPS"
...
+
 	// Path is the path to append to the host to reach the expected health check.
 	// Must start with "?" or "/", contain only valid URL characters and end with alphanumeric char or "/". For example "/" or "/healthz" are common
 	// +kubebuilder:validation:Pattern=`^(?:\?|\/)[\w\-.~:\/?#\[\]@!$&'()*+,;=]+(?:[a-zA-Z0-9]|\/){1}$`
 	Path string `json:"path,omitempty"`
+
 	// Protocol to use when connecting to the host, valid values are "HTTP" or "HTTPS"
+	// Defaults to HTTP
 	// +kubebuilder:validation:XValidation:rule="self in ['HTTP','HTTPS']",message="Only HTTP or HTTPS protocols are allowed"
+	// +kubebuilder:default=HTTP
```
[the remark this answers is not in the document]
maksymvavilov: Defaulting to secure now

maleck13 (approved):

PR #308, opened 2024-11-15
Bug: ensure we watch health probes in dnsrecord controller

maksymvavilov:
Changes: 
- Watch probes in the dnsrecords reconciler 
- Refactor the `receivedPrematurely` to consult a list of probes (and not to retrieve probes twice I'm not passing them around) 
- If we have an operator running without `probes-enabled` flag set to true we now will ignore healthcheck spec. Before we would be waiting for probes to be created (they will not be created) 
- Added test case to capture premature reconciliation case

### DIFF
--- a/internal/controller/dnsrecord_controller.go
+++ b/internal/controller/dnsrecord_controller.go
@@ -30,6 +30,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -95,6 +96,7 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	logger.Info("Reconciling DNSRecord")
 
 	reconcileStart = metav1.Now()
+	probes := &v1alpha1.DNSHealthCheckProbeList{}
 
 	defer postReconcile(ctx)
 
@@ -125,7 +127,7 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 				reason := "DNSProviderError"
 				message := fmt.Sprintf("The dns provider could not be loaded: %v", err)
 				setDNSRecordCondition(dnsRecord, string(v1alpha1.ConditionTypeReady), metav1.ConditionFalse, reason, message)
-				return r.updateStatus(ctx, previous, dnsRecord, false, []string{}, err)
+				return r.updateStatus(ctx, previous, dnsRecord, probes, false, []string{}, err)
 			}
 
 			if probesEnabled {
@@ -174,7 +176,7 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		logger.Error(err, "Failed to validate record")
 		setDNSRecordCondition(dnsRecord, string(v1alpha1.ConditionTypeReady), metav1.ConditionFalse,
 			"ValidationError", fmt.Sprintf("validation of DNSRecord failed: %v", err))
-		return r.updateStatus(ctx, previous, dnsRecord, false, []string{}, err)
+		return r.updateStatus(ctx, previous, dnsRecord, probes, false, []string{}, err)
 	}
 
 	//Ensure an Owner ID has been assigned to the record (OwnerID set in the status)
@@ -197,14 +199,14 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		if err != nil {
 			setDNSRecordCondition(dnsRecord, string(v1alpha1.ConditionTypeReady), metav1.ConditionFalse,
 				"DNSProviderError", fmt.Sprintf("The dns provider could not be loaded: %v", err))
-			return r.updateStatus(ctx, previous, dnsRecord, false, []string{}, err)
+			return r.updateStatus(ctx, previous, dnsRecord, probes, false, []string{}, err)
 		}
 
 		z, err := p.DNSZoneForHost(ctx, dnsRecord.Spec.RootHost)
 		if err != nil {
 			setDNSRecordCondition(dnsRecord, string(v1alpha1.ConditionTypeReady), metav1.ConditionFalse,
 				"DNSProviderError", fmt.Sprintf("Unable to find suitable zone in provider: %v", provider.SanitizeError(err)))
-			return r.updateStatus(ctx, previous, dnsRecord, false, []string{}, err)
+			return r.updateStatus(ctx, previous, dnsRecord, probes, false, []string{}, err)
 		}
 
 		//Add zone id/domainName to status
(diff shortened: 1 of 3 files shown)


internal/controller/dnsrecord_controller.go
```diff
@@ -358,14 +368,34 @@ func (r *DNSRecordReconciler) SetupWithManager(mgr ctrl.Manager, maxRequeue, val
 			}
 			return toReconcile
 		})).
+		Watches(&v1alpha1.DNSHealthCheckProbe{}, handler.EnqueueRequestsFromMapFunc(func(ctx context.Context, o client.Object) []reconcile.Request {
+			logger := log.FromContext(ctx)
+			probe, ok := o.(*v1alpha1.DNSHealthCheckProbe)
```
maleck13, line 373: should we do a check here to see has the health of the probe changed? As I understand it on update this Watch function will be called twice (once with the old and once with the new) so if we do a Get(probe) and check status against the probe we received if that status has changed then and only then requeue?

maleck13: I guess we don't have to do that right now but if we agree lets track with a follow up issue

internal/controller/dnsrecord_controller.go
```diff
@@ -358,14 +368,34 @@ func (r *DNSRecordReconciler) SetupWithManager(mgr ctrl.Manager, maxRequeue, val
...
+			if !ok {
+				logger.V(1).Info("unexpected object type", "error", fmt.Sprintf("%T is not a *v1alpha1.DNSHealthCheckProbe", o))
+			}
+
+			records := &v1alpha1.DNSRecordList{}
```
maleck13, line 378: I think here we could do a get based on the owner ref on the probe? We only have one probe here and that probe can only belong to one record right?

internal/controller/dnsrecord_controller.go
```diff
@@ -220,25 +221,34 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+		return r.updateStatus(ctx, previous, dnsRecord, nil, false, []string{}, err)
 	}
 
+	probes := &v1alpha1.DNSHealthCheckProbeList{}
 	if probesEnabled {
 		if err = r.ReconcileHealthChecks(ctx, dnsRecord, allowInsecureCert); err != nil {
 			return ctrl.Result{}, err
 		}
+		// get all probes owned by this record
+		if err := r.List(ctx, probes, &client.ListOptions{
```
maleck13, line 233: would there ever be more than one for a given record

internal/controller/dnsrecord_controller.go
```diff
@@ -446,7 +505,7 @@ func setStatusConditions(record *v1alpha1.DNSRecord, hadChanges bool, notHealthy
 	setDNSRecordCondition(record, string(v1alpha1.ConditionTypeReady), metav1.ConditionTrue, string(v1alpha1.ConditionReasonProviderSuccess), "Provider ensured the dns record")
 
 	// probes are disabled or not defined, or this is a wildcard record
-	if record.Spec.HealthCheck == nil || strings.HasPrefix(record.Spec.RootHost, v1alpha1.WildcardPrefix) {
+	if record.Spec.HealthCheck == nil || strings.HasPrefix(record.Spec.RootHost, v1alpha1.WildcardPrefix) || !probesEnabled {
```
maleck13, line 508: starting to think we might remove the probes enabled flag soon

internal/controller/dnsrecord_controller.go
maksymvavilov: Yes, a probe per "leaf". If you have two addresses on your gateway you will end up with an A record with two IPs and you will have a probe per IP

internal/controller/dnsrecord_controller.go
maksymvavilov: Checking it here

internal/controller/dnsrecord_controller.go
```diff
@@ -358,14 +382,53 @@ func (r *DNSRecordReconciler) SetupWithManager(mgr ctrl.Manager, maxRequeue, val
...
+				return []reconcile.Request{}
+			}
+
+			// probe.Status.Healthy is not nil here
+			// we probed for the first time - reconcile
+			// or none are nil and status changed - reconcile
+			if onCluster.Status.Healthy == nil || *probe.Status.Healthy != *onCluster.Status.Healthy {
+				return []reconcile.Request{{NamespacedName: client.ObjectKeyFromObject(probeOwner)}}
+			}
+
```
maksymvavilov, line 420: There are 2 filters for the reconciliation. This is the first one and will let through only if we flipped ready `Healthy` or did the first probe. The second will list all probes and will do a full loop only if it will result in a change of the status

internal/controller/dnsrecord_controller.go
```diff
@@ -220,25 +221,34 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	if err != nil {
 		setDNSRecordCondition(dnsRecord, string(v1alpha1.ConditionTypeReady), metav1.ConditionFalse,
 			"DNSProviderError", fmt.Sprintf("The dns provider could not be loaded: %v", err))
-		return r.updateStatus(ctx, previous, dnsRecord, false, []string{}, err)
+		return r.updateStatus(ctx, previous, dnsRecord, nil, false, []string{}, err)
 	}
 
+	probes := &v1alpha1.DNSHealthCheckProbeList{}
```
maleck13, line 227: if we defined this early would it help with readability. IE jusst pass in the empty list of probes to updateStatus? as passing in nil is a bit obscure

maksymvavilov: defined up top and passing it always into the update status func

maleck13 (commented):
Looks good tried a few scenrios DNSRecord and Policy ended up as expectd

```
status:
  conditions:
  - lastTransitionTime: "2024-11-20T07:14:32Z"
    message: DNSPolicy has been accepted
    reason: Accepted
    status: "True"
    type: Accepted
  - lastTransitionTime: "2024-11-21T08:15:08Z"
    message: DNSPolicy has been successfully enforced
    reason: Enforced
    status: "True"
    type: Enforced
  - lastTransitionTime: "2024-11-22T11:20:05Z"
    message: 'DNSPolicy has encountered some issues: not all sub-resources of policy
      are passing the policy defined health check. Not healthy DNSRecords are: external-t1c
      external-t1d '
    reason: Unknown
    status: "False"
    type: SubResourcesHealthy
  observedGeneration: 1
  recordConditions:
    t1c.cb.hcpapps.net:
    - lastTransitionTime: "2024-11-22T11:20:03Z"
      message: 'Not healthy addresses: [af2b62688c5454c35a1827a82e7d7c68-1089732621.eu-west-1.elb.amazonaws.com]'
      observedGeneration: 1
      reason: HealthChecksFailed
      status: "False"
      type: Healthy
    t1d.cb.hcpapps.net:
    - lastTransitionTime: "2024-11-22T11:20:03Z"
      message: 'Not healthy addresses: [af2b62688c5454c35a1827a82e7d7c68-1089732621.eu-west-1.elb.amazonaws.com]'
      observedGeneration: 1
      reason: HealthChecksFailed
      status: "False"
      type: Healthy
  totalRecords: 2
```

```
status:
  conditions:
  - lastTransitionTime: "2024-11-20T07:14:32Z"
    message: DNSPolicy has been accepted
    reason: Accepted
    status: "True"
    type: Accepted
  - lastTransitionTime: "2024-11-21T08:15:08Z"
    message: DNSPolicy has been successfully enforced
    reason: Enforced
    status: "True"
    type: Enforced
  - lastTransitionTime: "2024-11-22T14:06:42Z"
    message: All sub-resources are healthy
    reason: SubResourcesHealthy
    status: "True"
    type: SubResourcesHealthy
  observedGeneration: 1
  totalRecords: 2

```

PR #312, opened 2024-11-22
fix: Azure provider error parsing

mikenairn:
Fixes an issue in the cleanup of azure error messages that could cause a panic.

### DIFF
--- a/internal/external-dns/provider/azure/azure.go
+++ b/internal/external-dns/provider/azure/azure.go
@@ -466,8 +466,10 @@ func CleanAzureError(err error) error {
 	reg := regexp.MustCompile(`\"message\": \".*\"`)
 	errMsg := err.Error()
 	msg := reg.FindString(errMsg)
-	msgBits := strings.SplitAfterN(msg, ":", 2)
-	msgBits = strings.Split(msgBits[1], `"`)
-	msg = msgBits[1]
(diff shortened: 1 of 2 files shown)


internal/external-dns/provider/azure/azure.go
```diff
@@ -466,8 +466,10 @@ func CleanAzureError(err error) error {
...
 	errMsg := err.Error()
 	msg := reg.FindString(errMsg)
-	msgBits := strings.SplitAfterN(msg, ":", 2)
-	msgBits = strings.Split(msgBits[1], `"`)
-	msg = msgBits[1]
+	if msgBits := strings.SplitAfterN(msg, ":", 2); len(msgBits) > 1 {
+		if msgBits = strings.Split(msgBits[1], `"`); len(msgBits) > 1 {
+			msg = msgBits[1]
+		}
+	}
```
mikenairn, line 473: This whole ting could probably be better done with regular expressions and groupings, but i want to avoid a bigger change here.

mikenairn:
The Azure tests don't run automatically on PRs, triggered a separate job for it here [to verify](https://github.com/Kuadrant/dns-operator/actions/runs/11970371253/job/33372966335)

internal/external-dns/provider/azure/azure.go
maleck13: bit gross I agree.

PR #315, opened 2024-11-25
probes unhealthy only if threshold exeeded

maksymvavilov:
This will ensure we mark probes as unhealthy only when consecutive failures exceed the threshold. Before we would mark it as unhealthy and refuse update the record until threshold exceeded. 
This leads to a situation when the premature reconciliation on the health status change will be triggered but result in to changes and will fall back to a default validation time.

### DIFF
--- a/internal/controller/dnsrecord_healthchecks.go
+++ b/internal/controller/dnsrecord_healthchecks.go
@@ -136,20 +136,18 @@ func removeUnhealthyEndpoints(specEndpoints []*endpoint.Endpoint, dnsRecord *v1a
 
 	var haveHealthyProbes bool
 	for _, probe := range probes.Items {
-		// if the probe is healthy or unknown, continue to the next probe
+		// if the probe is healthy, continue to the next probe
 		if probe.Status.Healthy != nil && *probe.Status.Healthy {
 			haveHealthyProbes = true
 			continue
 		}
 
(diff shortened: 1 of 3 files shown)


internal/controller/dnsrecord_healthchecks.go
```diff
@@ -136,20 +136,18 @@ func removeUnhealthyEndpoints(specEndpoints []*endpoint.Endpoint, dnsRecord *v1a
...
 
-		// if we exceeded a threshold or we haven't probed yet
-		if probe.Status.ConsecutiveFailures >= dnsRecord.Spec.HealthCheck.FailureThreshold || probe.Status.Healthy == nil {
-			//delete bad endpoint from all endpoints targets
-			tree.RemoveNode(&common.DNSTreeNode{
-				Name: probe.Spec.Address,
-			})
-			unhealthyAddresses = append(unhealthyAddresses, probe.Spec.Address)
-		}
+		// if unhealthy or we haven't probed yet
```
maleck13, line 145: ok good so assume not healthy until proven is

maksymvavilov: yes. Unless `heathy: true` we consider it unhealthy

maleck13 (approved):
changes look good

also verified locally both going to healthy and unhealthy updated the status of the DNSPolicy as expected

PR #317, opened 2024-11-26
reconcile premature on probes nil -> healthy

maksymvavilov:


maleck13 (approved):
checked the following

- [ ] new gateway no status on healthchecks while dns propigating then becomes healthy (records not published until health passes, dnspolicy shows correct status)
- [ ] new gateway no status on healthchecks while dns propigating then stays unhealthy (records not published dnspolicy shows correct status)
- [ ] make the health check pass after failing for sometime (records published and dnspolicy shows correct status)
- [ ] make healthcheck fail after publish (records remain published dnspolicy shows correct status)

PR #326, opened 2024-12-04
Add scale test

mikenairn:
Add a distributed dns with multiple namespaced dns operator deployments scale test using kube burner.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,5 @@ Dockerfile.cross
 tmp
 
 config/local-setup/**/*.env
+
+*.log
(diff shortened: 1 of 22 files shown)


maleck13:
@jsmolar @trepel this is how we are looking at approaching automatic scale testing

maleck13 (approved):
It would be cool to see how this works. But I don't see anything here that makes me concerned so 👍 nice work

mikenairn:
> It would be cool to see how this works. But I don't see anything here that makes me concerned so 👍 nice work

I'll setup a call when @philbrookes is back in to go through it

PR #330, opened 2024-12-09
display healthy cond on get

maksymvavilov:
display healthy cond on get 
image

### DIFF
--- a/api/v1alpha1/dnsrecord_types.go
+++ b/api/v1alpha1/dnsrecord_types.go
@@ -162,6 +162,7 @@ type DNSRecordStatus struct {
 //+kubebuilder:object:root=true
 //+kubebuilder:subresource:status
 //+kubebuilder:printcolumn:name="Ready",type="string",JSONPath=".status.conditions[?(@.type==\"Ready\")].status",description="DNSRecord ready."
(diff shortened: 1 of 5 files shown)


mikenairn:
@maksymvavilov What does this display if you have no health check? Nothing?

maksymvavilov:
@mikenairn yes. 
Changed priority so it is not always there 
image

maksymvavilov:
there is also a [sister PR](https://github.com/Kuadrant/kuadrant-operator/pull/1074) on the kuadrant side if you are up to it

api/v1alpha1/dnsrecord_types.go
```diff
@@ -162,6 +162,7 @@ type DNSRecordStatus struct {
 //+kubebuilder:object:root=true
 //+kubebuilder:subresource:status
 //+kubebuilder:printcolumn:name="Ready",type="string",JSONPath=".status.conditions[?(@.type==\"Ready\")].status",description="DNSRecord ready."
+//+kubebuilder:printcolumn:name="Healthy",type="string",JSONPath=".status.conditions[?(@.type==\"Healthy\")].status",description="DNSRecord healthy.",priority=2
```
philbrookes, line 165: is there a way to make this display something like "unknown" rather than nothing at all, when the field is missing?

philbrookes (approved):

PR #370, opened 2025-01-29
Signing Helm packages

didierofrivia:
Part of the work needed for https://github.com/Kuadrant/helm-charts/issues/18

This PR introduces the GPG signing of Helm chart packages upon creation. It also uploads its provenance file to the GH release page.

The job now requires to be passed an environment variable `GPG_KEYRING_BASE64` which represents the GPG keyring base64 encoded, in order to be stored as a GH action variable.

### DIFF
--- a/make/helm.mk
+++ b/make/helm.mk
@@ -29,18 +29,18 @@ helm-upgrade: $(HELM) ## Upgrade the helm chart
 	# Upgrade the helm chart in the cluster
 	$(HELM) upgrade $(CHART_NAME) $(CHART_DIRECTORY)
 
+.PHONY: helm-package
+helm-package: $(HELM) ## Package the helm chart
+	# Package the helm chart
+	$(HELM) package $(CHART_DIRECTORY)
+
 # GPG_KEY_UID: substring of the desired key's uid, the name or email
 GPG_KEY_UID ?= 'Kuadrant Development Team'
-# GPG_KEYRING_BASE64: the gpg keyring base64 encoded
-GPG_KEYRING_BASE64 ?= <KUADRANT_GPG_KEYRING_BASE64>
-
-.PHONY: helm-package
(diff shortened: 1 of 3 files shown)


make/helm.mk
```diff
@@ -1,33 +1,46 @@
...
+GPG_KEY_UID ?= 'Kuadrant Development Team'
+# GPG_KEYRING_BASE64: the gpg keyring base64 encoded
+GPG_KEYRING_BASE64 ?= <KUADRANT_GPG_KEYRING_BASE64>
 
 .PHONY: helm-package
-helm-package: $(HELM) ## Package the helm chart
+helm-package: $(HELM) ## Package the helm chart and GPG sign it
+	# Store the key
+	mkdir -p .gpg
+	echo $(GPG_KEYRING_BASE64) | base64 -d > .gpg/kuadrantsecring.gpg  #storing base64 GPG key into keyring
```
eguzki, line 41: writing the key to the disk does not seem to be a good idea. Maybe there is no other way, so let it be. As I see, the `helm-package` makefile target should assume that the key is already in the default keyring (or keystore) of the system, so it should only run `helm package --sign --key whatever`, without specifying the keyring. The only thing that matters is the key name that should exist. So, manually, the key should already been imported, and for GHA, it would be required a previous step for importing the key, something like:

```yaml
- name: Configure GPG Key
  run: |
    echo -n "$GPG_SIGNING_KEY" | base64 --decode | gpg --import
  env:
    GPG_SIGNING_KEY: ${{ secrets.HELM_CHARTS_SIGNING_KEY }}
```

wdyt?

didierofrivia: yup, it makes sense, let me do some changes addressing this, thanks for the suggestion!

eguzki (approved):
Looking good to me!

maleck13 (approved):

PR #376, opened 2025-02-06
update tree with propagate logic and tests

philbrookes:
This is the base logic to power the changes to allow soft delete features for various options in the future - plus it has been left fairly open ended in case currently unseen use cases emerge.

The main development here is to realise that the records are double-ended (i.e. `host1 CNAME host2` connects host1 to host2), which are reflected in the tree as datasets, and it is these datasets that hold the labels. Once I made this realisation the feature became clearer to implement.

I have added various unit tests too, to give more confidence that this works as expected, though they can be hard to reason about as the trees are hard to visualise when constructed from code - for this I have made a sort of text diagram to illustrate the before and after of each test.

The general flow starts in `PropagateStoppableLabel` is as follows:
- Spread labels as greedily as possible
  - Any labelled node labels all of it's children
  - Any node with all children labelled get's the label too
- Resolve the stop labels
  - Any node with the stop label and the propagated label:
    - Has the label removed from itself and all it's children
    - Has the label removed from any parent (or parent's parent) that has the label

### DIFF
--- a/internal/common/tree.go
+++ b/internal/common/tree.go
@@ -13,6 +13,7 @@ type DNSTreeNode struct {
 	Name     string
 	Children []*DNSTreeNode
 	DataSets []DNSTreeNodeData
+	Parent   *DNSTreeNode
 }
 
 // DNSTreeNodeData holds a data for the enpoint(s) that correspond to this node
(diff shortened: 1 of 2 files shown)


internal/common/tree.go
```diff
@@ -25,6 +26,173 @@ type DNSTreeNodeData struct {
...
+		}
+	}
+}
+
+func AddChild(parent *DNSTreeNode, child *DNSTreeNode) {
+	parent.Children = append(parent.Children, child)
+	child.Parent = parent
+}
+
+func RemoveLabelFromParents(node *DNSTreeNode, label string) {
```
maleck13, line 153: Wondering if there is a bettter name as it is not obvious that root is not a parent? Just reading this name I would expect it removes it from all parents

`RemoveLabelFromNonRootParents`?

internal/common/tree.go
```diff
@@ -25,6 +26,173 @@ type DNSTreeNodeData struct {
...
+	}
+
+	//remove label from stop labelled children
+	for _, c := range node.Children {
+		d := findDataSetForChild(node, c.Name)
+		//has label and stop label = remove label from this dataset and all children
+		if d != nil && d.Labels[stopLabel] != "" && d.Labels[label] != "" {
+			delete(d.Labels, label)
+			RemoveLabelFromTree(c, label)
+			RemoveLabelFromParents(node, label)
```
maleck13, line 68: reading this made me go read RemoveLabelFromParents as I couldn't infer that it wouldn't remove it from the root node. Its a nit not a big deal. Maybe `RemoveLabelFromSubParents`?

internal/common/tree.go
philbrookes: well, it's only returning early if the node is a root, as a root doesn't have parents (Which is all that `IsRoot` tests for).

Or in other words, if the node you've passed in to `RemoveLabelFromParents`  has no parents then there is no work to do.

philbrookes: ah, it shouldn't be public, I'll update that

maleck13 (changes requested):
Might be easier to review / follow once used in implementation

maleck13 (approved):

PR #384, opened 2025-02-21
[WIP] [feature] Implement soft_delete and stop_soft_delete flags in the plan

philbrookes:
Some useful DNS Records for local testing (will require some tweaking e.g. hostnames): https://gist.github.com/philbrookes/617e03b739f43f01423f9ea94284333d

To test having one or more removed, edit the following:
```
    - dnsName: cluster1-gw1-ns1.klb.loadbalanced.pb.hcpapps.net
      recordTTL: 60
      recordType: A
      targets:
        - 172.18.200.1
```

To look like:
```
    - dnsName: cluster1-gw1-ns1.klb.loadbalanced.pb.hcpapps.net
      recordTTL: 60
      recordType: A
      targets:
        - 172.18.200.1
      labels:
        soft_delete: "true"
```

- mark cluster1 or cluster2 individually for removal, to see it removed.
- mark cluster3 for removal and it will not remove (only record in the GEO)
- mark cluster1 for removal, then afterwards cluster2 and see cluster1 removed but not cluster2 (last record in GEO).

As discussed, this will not cover removing the label on cluster1 and cluster2 simultaneously (i.e. race conditions) that will be covered in future work: https://github.com/Kuadrant/dns-operator/issues/385

### DIFF
--- a/internal/common/helper.go
+++ b/internal/common/helper.go
@@ -1,9 +1,12 @@
 package common
 
 import (
+	"fmt"
+	"io"
 	"time"
 
 	"k8s.io/apimachinery/pkg/util/rand"
+	externaldnsendpoint "sigs.k8s.io/external-dns/endpoint"
 )
 
 // RandomizeValidationDuration randomizes duration for a given variance with a min value of 1 sec
@@ -26,4 +29,35 @@ func RandomizeDuration(variance, duration float64) time.Duration {
 	return time.Millisecond * time.Duration(rand.Int63nRange(
 		int64(lowerLimit),
 		int64(upperLimit)))
+}
+
+func WriteEndpoints(s io.Writer, endpoints []*externaldnsendpoint.Endpoint, title string) {
+	fmt.Fprintf(s, "\n====== %s ======\n", title)
+	for _, ep := range endpoints {
+		fmt.Fprintf(s, "	endpoint: %v > %+v with labels: %+v and flags: %+v\n", ep.DNSName, ep.Targets, ep.Labels, ep.ProviderSpecific)
+	}
+	fmt.Fprintf(s, "====== %s ======\n\n", title)
+}
+
+func RemoveLabelFromEndpoint(label string, endpoint *externaldnsendpoint.Endpoint) *externaldnsendpoint.Endpoint {
+	if endpoint.Labels == nil {
+		return endpoint
+	}
+
+	delete(endpoint.Labels, label)
+	return endpoint
+}
+
+func RemoveLabelFromEndpoints(label string, endpoints []*externaldnsendpoint.Endpoint) []*externaldnsendpoint.Endpoint {
+	for _, e := range endpoints {
+		RemoveLabelFromEndpoint(label, e)
+	}
+	return endpoints
+}
+func RemoveLabelsFromEndpoints(labels []string, endpoints []*externaldnsendpoint.Endpoint) []*externaldnsendpoint.Endpoint {
+	for _, l := range labels {
+		RemoveLabelFromEndpoints(l, endpoints)
+	}
+
+	return endpoints
 }
--- a/internal/common/tree.go
+++ b/internal/common/tree.go
@@ -1,29 +1,30 @@
 package common
 
 import (
+	"fmt"
+	"io"
 	"slices"
 
 	"sigs.k8s.io/external-dns/endpoint"
 
 	"github.com/kuadrant/dns-operator/api/v1alpha1"
+	"github.com/kuadrant/dns-operator/internal/common/slice"
 )
 
 // DNSTreeNode stores a relation between endpoints that were parsed into a tree
 type DNSTreeNode struct {
-	Name     string
-	Children []*DNSTreeNode
-	DataSets []DNSTreeNodeData
-	Parent   *DNSTreeNode
+	Name      string
+	Children  []*DNSTreeNode
+	Endpoints []*endpoint.Endpoint
+	Parent    *DNSTreeNode
 }
 
-// DNSTreeNodeData holds a data for the enpoint(s) that correspond to this node
-type DNSTreeNodeData struct {
-	RecordType       string
-	SetIdentifier    string
-	RecordTTL        endpoint.TTL
-	Labels           endpoint.Labels
-	ProviderSpecific endpoint.ProviderSpecific
-	Targets          []string
+func (d *DNSTreeNode) String() string {
+	return fmt.Sprintf("host: %s, endpoints: %+v, children: %+v", d.Name, d.Endpoints, d.Children)
+}
+
+func WriteTree(s io.Writer, node *DNSTreeNode, title string) {
+	WriteEndpoints(s, *ToEndpoints(node, &[]*endpoint.Endpoint{}), title)
 }
 
 // PropagateStoppableLabel takes a propLabel (and value) to propagate throughout a tree, and a stopLabel
@@ -42,13 +43,51 @@ type DNSTreeNodeData struct {
 //   - Any node with the stopLabel and the propLabel:
 //   - Has the propLabel removed from itself and all it's children
 //   - Has the propLabel removed from any parent (or parent's parent) that has the label
-
 func PropagateStoppableLabel(node *DNSTreeNode, propLabel, value, stopLabel string) {
 	//propagate labels regardless of stop labels
-	propagateLabel(node, propLabel, value)
+	PropagateLabel(node, propLabel, value)
 
 	//propagate stop labels
 	resolveStops(node, propLabel, value, stopLabel)
+}
+
+func CopyLabel(label string, from, to *DNSTreeNode) {
+	toNode := FindNode(to, from.Name)
+	if toNode != nil {
+		// copy label values to any matching datasets
+		for _, fromEP := range from.Endpoints {
+			for _, toEP := range toNode.Endpoints {
+				if fromEP.Key() == toEP.Key() {
+					if fromEP.Labels == nil || fromEP.Labels[label] == "" {
+						if toEP.Labels != nil {
+							delete(toEP.Labels, label)
+						}
+					} else {
+						if toEP.Labels == nil {
+							toEP.Labels = endpoint.NewLabels()
+						}
+						toEP.Labels[label] = fromEP.Labels[label]
+					}
+				}
+			}
+		}
+	}
+	for _, c := range from.Children {
+		CopyLabel(label, c, to)
+	}
+}
+
+func FindNode(tree *DNSTreeNode, name string) *DNSTreeNode {
+	if tree.Name == name {
+		return tree
+	}
+	for _, c := range tree.Children {
+		n := FindNode(c, name)
+		if n != nil {
+			return n
+		}
+	}
+	return nil
 }
 
 func resolveStops(node *DNSTreeNode, label, value, stopLabel string) {
@@ -60,13 +99,19 @@ func resolveStops(node *DNSTreeNode, label, value, stopLabel string) {
 
 	//remove label from stop labelled children
 	for _, c := range node.Children {
-		d := findDataSetForChild(node, c.Name)
-		//has label and stop label = remove label from this dataset and all children
-		if d != nil && d.Labels[stopLabel] != "" && d.Labels[label] != "" {
-			delete(d.Labels, label)
-			RemoveLabelFromTree(c, label)
-			RemoveLabelFromParents(node, label)
-		} else {
+		hadLabel := false
+		endpoints := findEndpointsForChild(node, c.Name)
+		for _, ep := range endpoints {
+			//has label and stop label = remove label from this dataset and all children
+			if ep.Labels != nil && ep.Labels[stopLabel] == value && ep.Labels[label] == value {
+				delete(ep.Labels, label)
+				RemoveLabelFromTree(c, label)
+				removeLabelFromParents(node, label)
+				hadLabel = true
+				break
+			}
+		}
+		if !hadLabel {
 			resolveStops(c, label, value, stopLabel)
 		}
 
@@ -74,15 +119,25 @@ func resolveStops(node *DNSTreeNode, label, value, stopLabel string) {
 
 }
 
-func propagateLabel(node *DNSTreeNode, label, value string) {
+func PropagateLabel(node *DNSTreeNode, label, value string) {
 	for _, c := range node.Children {
-		d := findDataSetForChild(node, c.Name)
-		if d != nil && d.Labels[label] != "" {
+		dsets := findEndpointsForChild(node, c.Name)
+		if len(dsets) == 0 {
+			PropagateLabel(c, label, value)
+		}
+		foundLabel := false
+		for _, d := range dsets {
+			if d.Labels[label] != "" {
+				foundLabel = true
+				break
+			}
+		}
+		if foundLabel {
 			//this child is labelled, indiscriminately label entire tree under this child
 			AddLabelToTree(c, label, value)
 		} else {
 			// this child is not labelled, continue descending to propagate label
-			propagateLabel(c, label, value)
+			PropagateLabel(c, label, value)
 		}
 	}
 
@@ -98,26 +153,28 @@ func isRoot(node *DNSTreeNode) bool {
 
 func allChildrenHaveLabel(node *DNSTreeNode, label, value string) bool {
 	for _, c := range node.Children {
-		if !HasLabelForBranch(node, c.Name, label, value) {
+		if !hasLabelForBranch(node, c.Name, label, value) {
 			return false
 		}
 	}
 	return true
 }
 
-func findDataSetForChild(node *DNSTreeNode, name string) *DNSTreeNodeData {
-	for _, d := range node.DataSets {
-		if slices.Contains(d.Targets, name) {
-			return &d
+func findEndpointsForChild(node *DNSTreeNode, name string) []*endpoint.Endpoint {
+	ret := []*endpoint.Endpoint{}
+	for _, ep := range node.Endpoints {
+		if slices.Contains(ep.Targets, name) {
+			ret = append(ret, ep)
 		}
 	}
-	return nil
+	return ret
 }
 
 func AddLabelToBranch(node *DNSTreeNode, branch, label, value string) {
-	d := findDataSetForChild(node, branch)
-	if d == nil {
-		node.DataSets = append(node.DataSets, DNSTreeNodeData{
+	eps := findEndpointsForChild(node, branch)
+	if len(eps) == 0 {
+		node.Endpoints = append(node.Endpoints, &endpoint.Endpoint{
+			DNSName: node.Name,
 			Labels: endpoint.Labels{
 				label: value,
 			},
@@ -126,19 +183,28 @@ func AddLabelToBranch(node *DNSTreeNode, branch, label, value string) {
 			},
 		})
 	} else {
-		if len(d.Targets) == 1 {
-			d.Labels[label] = value
-		} else {
-			//remove target from shared dataset and recreate uniquely
-			for i, t := range d.Targets {
-				if t == branch {
-					d.Targets = append(d.Targets[:i], d.Targets[i+1:]...)
-					newDS := DNSTreeNodeData{
-						Labels:  d.Labels.DeepCopy(),
-						Targets: []string{branch},
+		for _, ep := range eps {
+			if len(ep.Targets) == 1 {
+				if ep.Labels == nil {
+					ep.Labels = endpoint.NewLabels()
+				}
+				ep.Labels[label] = value
+			} else {
+				//remove target from shared endpoint and recreate uniquely
+				for i, t := range ep.Targets {
+					if t == branch {
+						ep.Targets = append(ep.Targets[:i], ep.Targets[i+1:]...)
+						newEP := &endpoint.Endpoint{
+							Labels:  ep.Labels.DeepCopy(),
+							Targets: []string{branch},
+						}
+						if newEP.Labels == nil {
+							newEP.Labels = endpoint.NewLabels()
+						}
+						newEP.Labels[label] = value
+						node.Endpoints = append(node.Endpoints, newEP)
+						break
 					}
-					newDS.Labels[label] = value
-					node.DataSets = append(node.DataSets, newDS)
 				}
 			}
 		}
@@ -150,24 +216,23 @@ func AddChild(parent *DNSTreeNode, child *DNSTreeNode) {
 	child.Parent = parent
 }
 
-func RemoveLabelFromParents(node *DNSTreeNode, label string) {
+func removeLabelFromParents(node *DNSTreeNode, label string) {
 	if isRoot(node) {
 		return
 	}
 
-	d := findDataSetForChild(node.Parent, node.Name)
-	if d == nil {
-		return
+	eps := findEndpointsForChild(node.Parent, node.Name)
+
+	for _, ep := range eps {
+		delete(ep.Labels, label)
+
 	}
-
-	delete(d.Labels, label)
-
-	RemoveLabelFromParents(node.Parent, label)
+	removeLabelFromParents(node.Parent, label)
 }
 
 func RemoveLabelFromTree(node *DNSTreeNode, label string) {
-	for _, d := range node.DataSets {
-		delete(d.Labels, label)
+	for _, ep := range node.Endpoints {
+		delete(ep.Labels, label)
 	}
 
 	for _, c := range node.Children {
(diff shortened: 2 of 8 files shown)


internal/external-dns/plan/plan.go
```diff
@@ -200,6 +206,60 @@ func (p *Plan) Error() error {
...
+	}
+	return endpoints
+}
+func CleanLabelsFromEndpoints(labels []string, endpoints []*endpoint.Endpoint) []*endpoint.Endpoint {
+	for _, l := range labels {
+		CleanLabelFromEndpoints(l, endpoints)
+	}
+
+	return endpoints
+}
```
mikenairn, line 230: nit: Might just be me, but  "Remove" or "Delete" seem more appropriate than "Clean" i.e RemoveLabelsFromEndpoints. These are also really helpers nothing to do with the plan, maybe they would be better elsewhere?

internal/external-dns/plan/plan.go
```diff
@@ -200,6 +206,60 @@ func (p *Plan) Error() error {
...
+	common.PropagateStoppableLabel(currentTree, SoftDeleteLabel, "true", StopSoftDeleteLabel)
+
+	// write everything back to the plan
+	newDesired := []*endpoint.Endpoint{}
+	newCurrent := []*endpoint.Endpoint{}
+	common.ToEndpoints(desiredTree, &newDesired)
+	common.ToEndpoints(currentTree, &newCurrent)
+	p.Desired = common.MergeEndpoints(newDesired, p.Desired)
+	p.Current = common.MergeEndpoints(newCurrent, p.Current)
+}
```
mikenairn, line 261: Does this have to create entirely new slices of endpoints? Can it not just label the endpoints in memory or is it because its going through this tree code, does it not just maintain a pointer to the endpoint structs in the tree?

internal/common/tree.go
```diff
@@ -343,11 +446,11 @@ func findChildren(name string, record *v1alpha1.DNSRecord) []*DNSTreeNode {
...
-	for _, ep := range record.Spec.Endpoints {
+func findDataSets(name string, endpoints []*endpoint.Endpoint) []*DNSTreeNodeData {
+	dataSets := []*DNSTreeNodeData{}
+	for _, ep := range endpoints {
 		if ep.DNSName == name {
-			dataSets = append(dataSets, DNSTreeNodeData{
+			dataSets = append(dataSets, &DNSTreeNodeData{
 				RecordType:       ep.RecordType,
 				RecordTTL:        ep.RecordTTL,
 				SetIdentifier:    ep.SetIdentifier,
```
mikenairn, line 456: I'm not 100% i understand all this code, but why doesn't it just keep a pointer to Endpoint resources? Why does it have it's own `DNSTreeNodeData` struct?

internal/controller/dnsrecord_controller.go
```diff
@@ -659,7 +659,7 @@ func (r *DNSRecordReconciler) applyChanges(ctx context.Context, dnsRecord *v1alp
 	}
 
 	// add related endpoints to the record
-	dnsRecord.Status.ZoneEndpoints = mergeZoneEndpoints(
+	dnsRecord.Status.ZoneEndpoints = common.MergeEndpoints(
```
mikenairn, line 662: I think you just moved this, with no changes from what i can tell, but I'd rather we looked at a way of getting rid of this entirely. This is combining slices that it thinks should now be in the provider, but really this should just be a list pulled directly from the provider unmodified or something that the plan has calculated and exposed.

mikenairn (commented):
Some questions but i think this looks fine.

I would eventually rather the whole processing of endpoints wasn't done in the plan but rather done as an external (optional) adapter to the provider requests (Like the TXT registry) that just adds the labels, and the plan happens to understand the meaning of them, much like the owner labels and the current registries.

internal/common/tree.go
philbrookes: because the same host (node, e.g. klb.domain.com) can have multiple sets of targets with unique labels. In the DNS Record this is acheived with multiple endpoints with the same host, in the tree the host == node so the different endpoints (e.g. GEO * and GEO eu) are shown as 2 datasets for the same host.

internal/external-dns/plan/plan.go
philbrookes: it does maintain a handle to the endpoints, and I originally did as you say, however the problem caused here was that the tree ignores any endpoints that aren't able to be traced back to the root, and in some tests those records are expected to be preserved. I changed this code to get those tests to pass again.

internal/controller/dnsrecord_controller.go
philbrookes: yes, I did just move it, and I would also like to see it removed eventually. As I mentioned in another comment this was primarily making a test that started failing pass again, I didn't see this as necessary in any of my real world testing.

philbrookes:
> Some questions but i think this looks fine.
> 
> I would eventually rather the whole processing of endpoints wasn't done in the plan but rather done as an external (optional) adapter to the provider requests (Like the TXT registry) that just adds the labels, and the plan happens to understand the meaning of them, much like the owner labels and the current registries.

I've added this issue to track moving this code to an adapter, if you want to put any thoughts on it there.

Ideally this will allow us to merge this PR now.

https://github.com/Kuadrant/dns-operator/issues/393

internal/external-dns/plan/plan.go
```diff
@@ -200,6 +206,37 @@ func (p *Plan) Error() error {
 	return errors.Join(p.Errors...)
 }
 
+func (p *Plan) processStopLabel() {
+	if p.RootHost == nil {
+		return
+	}
+	desiredTree := common.MakeTreeFromEndpoints(*p.RootHost, p.Desired)
+	currentTree := common.MakeTreeFromEndpoints(*p.RootHost, p.Current)
```
mikenairn, line 214: Noting, that the `rootHost` is concept we added to our DNSRecord, and pass in (optionally) to our version of the plan. Code in here, external-dns, should not rely on it.

mikenairn (changes requested):
I need to understand the current tree logic better, but it seems to me at a brief glance, too reliant on our current record setup, in particular the loadblanced set of endpoints that we expect to actually relate to one another. I would have expected a tree structure to just allow the insertion of any Endpoint resource, and the tree would put it in the correct place based on whatever logic it uses (target -> dnsName in this case).

That aside, the current implementation is adding a lot of extra manipulation of the endpoints for very little reason or gain that I can see. It's currently converting everything(the tree decides is related to the rootHost via target name) to a
different type, manipulating them in some way, converting it back to an Endpoint and merging it back with all the original slices, presumably to ensure we don't lose endpoints. Seems to be adding a lot of ways things could go wrong just to add labels.

Main issue though is basic functionality appears to be broken.

A record like the following can be created, but it doesn't update or delete endpoints correctly(Nothing to do with labels, just normal operation here). I create that and change weights or delete endpoints, it's never reflected correctly in the provider.
```
kind: DNSRecord
metadata:
  name: dnsrecord-test-aws
spec:
  rootHost: dnsrecord-test.mn.hcpapps.net
  endpoints:
    - dnsName: dnsrecord-test.mn.hcpapps.net
      recordTTL: 60
      recordType: CNAME
      targets:
        - lb.dnsrecord-test.mn.hcpapps.net
    - dnsName: lb.dnsrecord-test.mn.hcpapps.net
      providerSpecific:
        - name: weight
          value: "150"
      recordTTL: 60
      recordType: CNAME
      setIdentifier: cluster1.dnsrecord-test.mn.hcpapps.net
      targets:
        - cluster1.dnsrecord-test.mn.hcpapps.net
    - dnsName: lb.dnsrecord-test.mn.hcpapps.net
      providerSpecific:
        - name: weight
          value: "230"
      recordTTL: 60
      recordType: CNAME
      setIdentifier: cluster2.dnsrecord-test.mn.hcpapps.net
      targets:
        - cluster2.dnsrecord-test.mn.hcpapps.net
    - dnsName: cluster1.dnsrecord-test.mn.hcpapps.net
      recordTTL: 60
      recordType: A
      targets:
        - 127.0.0.2
    - dnsName: cluster2.dnsrecord-test.mn.hcpapps.net
      recordTTL: 60
      recordType: A
      targets:
        - 127.0.0.3
  providerRef:
    name: dns-provider-credentials-aws
```

Also think we need to review our current tests, if this wasn't caught by the current test suite we need better ones.

philbrookes:
> I need to understand the current tree logic better, but it seems to me at a brief glance, too reliant on our current record setup, in particular the loadblanced set of endpoints that we expect to actually relate to one another. I would have expected a tree structure to just allow the insertion of any Endpoint resource, and the tree would put it in the correct place based on whatever logic it uses (target -> dnsName in this case).
> 
> That aside, the current implementation is adding a lot of extra manipulation of the endpoints for very little reason or gain that I can see. It's currently converting everything(the tree decides is related to the rootHost via target name) to a different type, manipulating them in some way, converting it back to an Endpoint and merging it back with all the original slices, presumably to ensure we don't lose endpoints. Seems to be adding a lot of ways things could go wrong just to add labels.
> 
> Main issue though is basic functionality appears to be broken.
> 
> A record like the following can be created, but it doesn't update or delete endpoints correctly(Nothing to do with labels, just normal operation here). I create that and change weights or delete endpoints, it's never reflected correctly in the provider.
> 
> ```
> apiVersion: kuadrant.io/v1alpha1
> kind: DNSRecord
> metadata:
>   name: dnsrecord-test-aws
> spec:
>   rootHost: dnsrecord-test.mn.hcpapps.net
>   endpoints:
>     - dnsName: dnsrecord-test.mn.hcpapps.net
>       recordTTL: 60
>       recordType: CNAME
>       targets:
>         - lb.dnsrecord-test.mn.hcpapps.net
>     - dnsName: lb.dnsrecord-test.mn.hcpapps.net
>       providerSpecific:
>         - name: weight
>           value: "150"
>       recordTTL: 60
>       recordType: CNAME
>       setIdentifier: cluster1.dnsrecord-test.mn.hcpapps.net
>       targets:
>         - cluster1.dnsrecord-test.mn.hcpapps.net
>     - dnsName: lb.dnsrecord-test.mn.hcpapps.net
>       providerSpecific:
>         - name: weight
>           value: "230"
>       recordTTL: 60
>       recordType: CNAME
>       setIdentifier: cluster2.dnsrecord-test.mn.hcpapps.net
>       targets:
>         - cluster2.dnsrecord-test.mn.hcpapps.net
>     - dnsName: cluster1.dnsrecord-test.mn.hcpapps.net
>       recordTTL: 60
>       recordType: A
>       targets:
>         - 127.0.0.2
>     - dnsName: cluster2.dnsrecord-test.mn.hcpapps.net
>       recordTTL: 60
>       recordType: A
>       targets:
>         - 127.0.0.3
>   providerRef:
>     name: dns-provider-credentials-aws
> ```
> 
> Also think we need to review our current tests, if this wasn't caught by the current test suite we need better ones.

oh wow, I'm glad you tested this, I had assumed that was covered by our tests as you mention.

I don't currently understand why that's not working, I'll have to look into it when I get back.

internal/external-dns/plan/plan.go
philbrookes: yeah, this is a good point. I'll address this when I get back.

internal/external-dns/plan/plan.go
```diff
@@ -200,6 +206,29 @@ func (p *Plan) Error() error {
...
+	common.CopyLabel(SoftDeleteLabel, desiredTree, currentTree)
+	common.CopyLabel(StopSoftDeleteLabel, desiredTree, currentTree)
+
+	// propagate label in desired tree ignoring stop labels
+	common.PropagateLabel(desiredTree, SoftDeleteLabel, "true")
+
+	// propagate the stop label in the current tree
+	common.PropagateStoppableLabel(currentTree, SoftDeleteLabel, "true", StopSoftDeleteLabel)
+}
+
```
mikenairn, line 231: I have read through this several times and i don't really understand the flow here. Can you explain it?

* Why do we create two trees instead of a single tree that is a combination of both current and desired (current overlaid with desired) and represents the final combined structure?
* Why do we copy labels from desired to current?
* Why are the `SoftDeleteLabel` labels propagated on desired, but the `SoftDeleteLabel` and `StopSoftDeleteLabel` dealt with on current? Why is current being manipulated at all?

internal/common/tree.go
```diff
@@ -126,19 +182,24 @@ func AddLabelToBranch(node *DNSTreeNode, branch, label, value string) {
...
-						Labels:  d.Labels.DeepCopy(),
-						Targets: []string{branch},
+		for _, ep := range eps {
+			if len(ep.Targets) == 1 {
+				if ep.Labels == nil {
+					ep.Labels = endpoint.NewLabels()
+				}
+				ep.Labels[label] = value
+			} else {
+				//remove target from shared endpoint and recreate uniquely
```
mikenairn, line 192: What does this mean? Not sure why this would ever be creating an endpoint, it should just be working with endpoints in the tree(referenced from the lists) no?

internal/external-dns/plan/plan.go
```diff
@@ -292,6 +307,11 @@ func (p *Plan) Calculate() *Plan {
 			var rTypeUpdate endpointUpdate
 			recordsByType := t.resolver.ResolveRecordTypes(key, row)
 			for _, records := range recordsByType {
+				// handle soft_delete
+				if records.current != nil && records.current.Labels != nil && records.current.Labels[SoftDeleteLabel] == "true" {
+					p.processDelete(records, key, &managedChanges)
+					continue
+				}
```
mikenairn, line 314: Hmm :thinking:, i see you are actually checking for the existence of this label on the current endpoint rather than desired, why is that?

internal/common/tree.go
```diff
@@ -42,15 +43,53 @@ type DNSTreeNodeData struct {
...
+						td.Labels[label] = fd.Labels[label]
+					}
+				}
+			}
+		}
+	}
+	for _, c := range from.Children {
+		CopyLabel(label, c, to)
+	}
+}
```
mikenairn, line 78: Can you explain what this (`CopyLabel`) is doing? It looks like it is deciding equality of endpoints based on the targets, but not sure why you would ever do that.

internal/common/helper.go
```diff
@@ -27,3 +31,82 @@ func RandomizeDuration(variance, duration float64) time.Duration {
...
+	for _, p := range e1.ProviderSpecific {
+		if !slices.ContainsFunc(e2.ProviderSpecific, func(e externaldnsendpoint.ProviderSpecificProperty) bool {
+			return e.Name == p.Name && e.Value == p.Value
+		}) {
+			return false
+		}
+	}
+
+	return true
+}
```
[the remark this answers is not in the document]
philbrookes: yes it is unused now, but I thought it might be useful someday. I can remove it, if you'd prefer?

internal/common/tree.go
philbrookes: this scenario is very unlikely to occur, but it felt incorrect to leave it unhandled. To recreate it would go something like:
- Have one endpoint: cname.x with targets: cname1.x and cname2.x
- Add a label to cname1.x target

As I see this, we either:
- Add the label to the endpoint that contains the target (thereby inaccurately labelling cname2.x as well)
- Move the target to be labelled to it's own endpoint

In our usecase the label is being used to mark endpoints to be unpublished, so it made sense to me to remove it from the targets and make it in a new endpoint (that will be unpublished) with the label.

internal/common/tree.go
```diff
@@ -176,84 +239,63 @@ func RemoveLabelFromTree(node *DNSTreeNode, label string) {
...
-	for i, node := range n.Children {
-		if node.Name == deleteNode.Name {
-			n.Children = append(n.Children[:i], n.Children[i+1:]...)
-
-			// we removed child, but need to clean up data sets
-			for j, dataset := range n.DataSets {
-				index := slices.Index(dataset.Targets, deleteNode.Name)
-				if index >= 0 {
-					n.DataSets[j].Targets = append(dataset.Targets[:index], dataset.Targets[index+1:]...)
+func RemoveNode(container, removeNode *DNSTreeNode) {
```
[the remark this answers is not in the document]
philbrookes: yes exactly, once the new health check logic is in, I'd expect this to be deleted.

internal/common/tree.go
```diff
@@ -343,21 +391,14 @@ func findChildren(name string, record *v1alpha1.DNSRecord) []*DNSTreeNode {
...
-				Labels:           ep.Labels,
-				ProviderSpecific: ep.ProviderSpecific,
-				Targets:          ep.Targets,
-			})
+			retEndpoints = append(retEndpoints, ep)
 		}
 	}
-	return dataSets
+	return retEndpoints
 }
```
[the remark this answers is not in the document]
philbrookes: Unless I'm mistaken. This should return all matching endpoints, for all record types, whose DNSName matches the current nodes host.

internal/common/tree.go
philbrookes: this is because there can be multiple endpoints with the same host (which is matched on line 55), but with different targets (i.e. GEO CNAMES), so the endpoint is only considered to be the same endpoint if it has the same host and targets (arguably this should also compare record TYPE, though that hasn't presented a problem yet).

internal/external-dns/plan/plan.go
```diff
@@ -219,6 +248,8 @@ func (p *Plan) Calculate() *Plan {
 		p.DomainFilter = append(p.DomainFilter, &rootDomainFilter)
 	}
 
+	p.processStopLabel()
```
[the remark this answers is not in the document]
philbrookes: This is a great idea, I'll do it now.

internal/external-dns/plan/plan.go
```diff
@@ -249,8 +280,14 @@ func (p *Plan) Calculate() *Plan {
 		if len(row.current) == 0 {
 			recordsByType := t.resolver.ResolveRecordTypes(key, row)
 			for _, records := range recordsByType {
-				if len(records.candidates) > 0 {
-					managedChanges.creates = append(managedChanges.creates, t.resolver.ResolveCreate(records.candidates))
+				candidates := []*endpoint.Endpoint{}
+				for _, c := range records.candidates {
+					if _, ok := c.Labels[SoftDeleteLabel]; !ok {
```
maksymvavilov, line 285: I see that here we are checking for the presence of the label and later on we are checking for the value being `true`. Would `soft_delete=blah` break anything?

philbrookes: yeah, it's a good point I was vacillating between needing a particular value, and just requiring the label to be present. I think expecting a value makes sense though. I'll update this.

internal/common/tree.go
```diff
@@ -176,84 +239,63 @@ func RemoveLabelFromTree(node *DNSTreeNode, label string) {
...
-	d := findDataSetForChild(node, branch)
-	if d == nil {
-		return false
-	}
-	if v, ok := d.Labels[label]; ok {
-		return value == v
+func hasLabelForBranch(node *DNSTreeNode, branch, label, value string) bool {
+	for _, ep := range findEndpointsForChild(node, branch) {
+		if v, ok := ep.Labels[label]; ok {
+			if value == v {
```
maksymvavilov, line 261: would this work?
```suggestion
			return value == v
```

maksymvavilov, line 261: is 
```
if value == v {
	return true
} else {
	return false
}
```

equivalent to `return value == v `?

philbrookes: nice, cheers. I assume I had it like that while I was wanting to log the decisions, but I'll change it now :)

internal/common/tree.go
mikenairn: Targets can change, any difference in what desired wants the record targets to look like and what targets are already in the current will cause this to fail, they will never match.

An endpoint is unique within a record set when it's dnsName, Type and SetIdentifier are unique, this is what makes up it's [key](https://github.com/kubernetes-sigs/external-dns/blob/master/endpoint/endpoint.go#L306).

This is what you should be comparing when you are looking for endpoints that represent the same dns record in the provider, i.e. `fd.Key() == td.Key()`

internal/common/helper.go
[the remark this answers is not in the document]
mikenairn: :man_shrugging: Never seen the point of adding code that we don't plan to use, just more to maintain.

internal/external-dns/plan/plan.go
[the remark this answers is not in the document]
philbrookes: I've added some tests to the plan_test.go now.

internal/common/tree.go
```diff
@@ -176,84 +239,59 @@ func RemoveLabelFromTree(node *DNSTreeNode, label string) {
...
+	} else {
+		for _, ep := range node.Endpoints {
+			if ep.Labels == nil {
+				ep.Labels = endpoint.NewLabels()
+			}
+			ep.Labels[label] = value
+		}
+	}
 	for _, c := range node.Children {
 		AddLabelToBranch(node, c.Name, label, value)
```
maksymvavilov, line 253: The funny thing is that with this line commented out, all tests but one succeed. The `AddLabelToBranch` is basically "label the node" and `AddLabelToTree` is "label everything under including the node". I don't think that matters, tbh but if we migrate the "create new endpoint" bit to the `AddLabelToTree` this should do the same job with half as much code :thinking:

internal/common/tree.go
```diff
@@ -176,84 +239,59 @@ func RemoveLabelFromTree(node *DNSTreeNode, label string) {
...
-	if d == nil {
-		return false
-	}
-	if v, ok := d.Labels[label]; ok {
-		return value == v
+func hasLabelForBranch(node *DNSTreeNode, branch, label, value string) bool {
+	for _, ep := range findEndpointsForChild(node, branch) {
+		if v, ok := ep.Labels[label]; ok {
+			return value == v
+		}
```
maksymvavilov, line 262: The "it could be easier" bit (ignorable):
It is probably just me but "branch" means everything under the child, while in this code the "branch" seems to mean "node" :thinking:

internal/common/tree.go
```diff
@@ -60,29 +99,45 @@ func resolveStops(node *DNSTreeNode, label, value, stopLabel string) {
...
+
+		}
+		for _, d := range dsets {
+			//has label and stop label = remove label from this dataset and all children
+			if d.Labels[stopLabel] != "" && d.Labels[label] != "" {
+				delete(d.Labels, label)
+				RemoveLabelFromTree(c, label)
+				removeLabelFromParents(node, label)
+			} else {
+				resolveStops(c, label, value, stopLabel)
```
maksymvavilov, line 114: The "it could be easier" bit (ignorable):
Seems like we are calling functions inside the loop with non-loop-specific values. In other words for each endpoints/dataset we will make recursive calls wit the same values :thinking:

internal/common/tree.go
```diff
@@ -60,29 +99,45 @@ func resolveStops(node *DNSTreeNode, label, value, stopLabel string) {
...
-		d := findDataSetForChild(node, c.Name)
-		//has label and stop label = remove label from this dataset and all children
-		if d != nil && d.Labels[stopLabel] != "" && d.Labels[label] != "" {
-			delete(d.Labels, label)
-			RemoveLabelFromTree(c, label)
-			RemoveLabelFromParents(node, label)
-		} else {
+		dsets := findEndpointsForChild(node, c.Name)
+		if len(dsets) == 0 {
 			resolveStops(c, label, value, stopLabel)
```
maksymvavilov, line 104: The "it could be easier" bit (ignorable):
If there are no datasets on a node, will it ever have children? Is this jut an obscure way to `return`?

internal/common/tree.go
```diff
@@ -150,24 +211,26 @@ func AddChild(parent *DNSTreeNode, child *DNSTreeNode) {
...
-func RemoveLabelFromParents(node *DNSTreeNode, label string) {
+func removeLabelFromParents(node *DNSTreeNode, label string) {
 	if isRoot(node) {
 		return
 	}
 
-	d := findDataSetForChild(node.Parent, node.Name)
-	if d == nil {
+	eps := findEndpointsForChild(node.Parent, node.Name)
+	if len(eps) == 0 {
```
maksymvavilov, line 220: Seems like the tests only cover `if isRoot()` part. This is strange since we don't copy stop label so it should be found only on the leafs and I would expect the situation to be opposite. Can you please explain to me what I am missing?

internal/common/tree.go
philbrookes: yes, I agree with the phraseology, if we're saying each child of a node is a branch from that node. 

Then this function is asking "does this node have this label for it's child(branch) with this name".

internal/common/tree.go
philbrookes: actually this whole check isn't required. I'll remove it

internal/common/tree.go
philbrookes: good catch I'll address this.

internal/common/tree.go
```diff
@@ -42,15 +43,53 @@ type DNSTreeNodeData struct {
...
-	propagateLabel(node, propLabel, value)
+	PropagateLabel(node, propLabel, value)
 
 	//propagate stop labels
 	resolveStops(node, propLabel, value, stopLabel)
 }
 
+func CopyLabel(label string, from, to *DNSTreeNode) {
+	toNode := FindNode(to, from.Name)
+	if toNode != nil {
```
maksymvavilov, line 56: The "it could be easier" bit (ignorable):
If `toNode` is nil should we even bother?

internal/common/tree.go
```diff
@@ -42,15 +43,53 @@ type DNSTreeNodeData struct {
...
+		// copy label values to any matching datasets
+		for _, fd := range from.Endpoints {
+			for _, td := range toNode.Endpoints {
+				if fd.Key() == td.Key() {
+					if fd.Labels == nil || fd.Labels[label] == "" {
+						if td.Labels != nil {
+							delete(td.Labels, label)
+						}
+					} else if fd.Labels[label] != "" {
```
maksymvavilov, line 65: The "it could be easier" bit (ignorable):
Shouldn't it just be `else`? :thinking:

internal/common/tree.go
philbrookes: ah yes, this is because the stop labels in the test are only ever directly below root. I'll add a test to cover the missed code. Thanks for pointing that out.

internal/common/tree.go
philbrookes: no, but we still need to check if the children have labels to copy in that scenario.

internal/common/tree.go
philbrookes: :+1:

internal/common/tree.go
philbrookes: I've played around with what you're suggesting, but I can't seem to find a way to make that work. Let's leave it for now I think, as we are running out of time.

maksymvavilov (approved):
Looks good. Great work!

PR #403, opened 2025-03-10
Add CoreDNS plugin

### LINKED ISSUE #396: Add tests and harden Core DNS plugin

mikenairn:
Adds a kuadrant CoreDNS plugin that enables serving zone data directly from kuadrant DNSRecord resources.

The plugin sets up watch and listers on kuadrant's DNSRecord resources in the k8s cluster and as it discovers them processes them and adds the endpoints to the appropriate DNS zone with the correct GEO and Weighted data.
The plugin uses logic from the [CoreDNS file plugin](https://github.com/coredns/coredns/tree/master/plugin/file) to create a functioning DNS server.

Refer to the [README](https://github.com/Kuadrant/dns-operator/blob/14849be04b716038cc63b28668afe7f42c3fb2e5/coredns/plugin/README.md) for more details of it's usage.

closes #396 

Follow on tasks:

* https://github.com/Kuadrant/dns-operator/issues/409
* https://github.com/Kuadrant/dns-operator/issues/410
* https://github.com/Kuadrant/dns-operator/issues/411
* https://github.com/Kuadrant/dns-operator/issues/412
* https://github.com/Kuadrant/dns-operator/issues/413
* https://github.com/Kuadrant/dns-operator/issues/414
* https://github.com/Kuadrant/dns-operator/issues/416

Note: Additional unit tests will follow, just want to get the base merged before adding even more to it.


coredns/plugin/README.md
```diff
@@ -0,0 +1,2 @@
+# kuadrant CoreDNS Plugin
+
```
maleck13, line 2: Do we expect others might want to make use of this in their own core dns? Should we add more details and consider making it available to others to use?

mikenairn: I intend to add some content to this README before i move it from draft, but in terms of someone wanting to include this plugin in their own CoreDNS then it would be done in the same way as any other external plugin as described [here](https://coredns.io/2017/07/25/compile-time-enabling-or-disabling-plugins/).

coredns/plugin/zone.go
```diff
@@ -0,0 +1,233 @@
...
+}
+
+// parseGeoAnswers takes a slice of answers for a dns name and reduces it down to a single answer based on geo.
+func (z *Zone) parseGeoAnswers(ctx context.Context, request request.Request, grrs []dns.RR) []dns.RR {
+	log.Debugf("parsing geo answers for %s", grrs[0].Header().Name)
+	log.Debugf("source ip is %s", request.IP())
+	var answer *dns.RR
+	var geoRRs []geodRR
+
+	roundRobinShuffle(grrs)
```
mikenairn, line 177: This should probably be removed, but it makes me wonder what we would like the default behaviour to be when we can't match any geo to the current request. Do we want it to always just return the first RR in the list?

Currently this will cause it to randomise what it would return if there are multiple geos available.

coredns/plugin/k8s.go
```diff
@@ -0,0 +1,213 @@
...
+							opts.LabelSelector = labelSelector.String()
+							return c.DNSRecords(ns).List(ctx, opts)
+						},
+						WatchFunc: func(opts metav1.ListOptions) (watch.Interface, error) {
+							opts.LabelSelector = labelSelector.String()
+							return c.DNSRecords(ns).Watch(ctx, opts)
+						},
+					},
+					&v1alpha1.DNSRecord{},
+					defaultResyncPeriod,
```
maleck13, line 104: does it matter that we are passing 0 as default? Not sure what that means

coredns/plugin/kuadrant.go
```diff
@@ -0,0 +1,118 @@
...
+		log.Debugf("request %s has not matched any zones %v", qname, k.Zones)
+		return plugin.NextOrFailure(k.Name(), k.Next, ctx, w, r)
+	}
+
+	z, ok := k.Zones.Z[zone]
+	if !ok || z == nil {
+		return dns.RcodeServerFailure, nil
+	}
+
+	// If transfer is not loaded, we'll see these, answer with refused (no transfer allowed).
```
maleck13, line 66: what is transfer?

coredns/plugin/weighted.go
```diff
@@ -0,0 +1,102 @@
...
+
+	for _, r := range address {
+		wsum += uint(r.weight)
+	}
+
+	// Select the first (top) IP
+	sort.Slice(address, func(i, j int) bool {
+		return address[i].weight > address[j].weight
+	})
+	v := w.randUint(wsum)
```
maleck13, line 84: So generates a random number between 0 and the sum of all the weights. Then orders the addresses by weight desc and returns the first result that has a weight or combined weight greater than the random value. So if you have a weight of 200 and another of 100 you end up with 300 and a random number between 0 and 300 picked so you are twice as likely to get one < 200 than you are to get one between >= 200 <= 300. So kinda like a weighted coin flip it wont be super predictable but will give you the weighting. At least this is how I read it

coredns/plugin/zone.go
```diff
@@ -0,0 +1,233 @@
...
+			//Take the first answer in the default case, not geo or weighted
+			rrs = []dns.RR{rrs[0]}
+		}
+		return rrs[0]
+	}
+
+	ns := &dns.NS{Hdr: dns.RR_Header{Name: dns.Fqdn(name), Rrtype: dns.TypeNS, Ttl: ttlSOA, Class: dns.ClassINET},
+		Ns: dnsutil.Join("ns1", name),
+	}
+	z.file.Insert(ns)
```
maleck13, line 60: haven't checked but assume this checks before adding to remove duplication concern?

coredns/plugin/zone.go
maleck13: I think we should leverage the default GEO like AWS

maleck13 (dismissed):

coredns/plugin/README.md
mikenairn: Updated. I wrote the README in a similar style to all the other COreDNS plugins

coredns/plugin/zone.go
mikenairn: Added a follow on task to look into this https://github.com/Kuadrant/dns-operator/issues/409

coredns/plugin/cmd/coredns.go
```diff
@@ -0,0 +1,46 @@
...
+
+// https://github.com/ori-edge/k8s_gateway/blob/master/cmd/coredns.go
+func init() {
+	var directives []string
+	var alreadyAdded bool
+
+	for _, name := range dnsserver.Directives {
+
+		if dropPlugins[name] {
+			if !alreadyAdded {
```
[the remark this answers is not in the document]
mikenairn: This is no longer relevant since the changes to limit the number of plugins enabled by default, but it was done this way to add the plugin at a particular place in the list of directives since the order is important. i.e. must be after geoip/metadata etc..

coredns/plugin/k8s.go
mikenairn: AFAIK it just means that it won't resync when it's set to 0, i believe this is the default so could just be removed and it would do the same thing. https://github.com/kubernetes/client-go/blob/master/tools/cache/shared_informer.go#L323

coredns/plugin/weighted.go
mikenairn: Created follow on issue to look at improving this https://github.com/Kuadrant/dns-operator/issues/410

coredns/plugin/zone.go
```diff
@@ -0,0 +1,233 @@
...
+	file   *file.Zone
+	rrData map[dns.RR]rrData
+}
+
+func NewZone(name string) *Zone {
+	z := &Zone{
+		file.NewZone(name, ""),
+		map[dns.RR]rrData{},
+	}
+
```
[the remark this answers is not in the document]
mikenairn: comment added

.github/workflows/build-coredns-images.yaml
```diff
@@ -0,0 +1,68 @@
+name: Build and Publish CoreDNS Images
+
+on:
+  push:
+    branches:
+      - main
+      - coredns_plugin
```
mikenairn, line 7: @mikenairn Remove this before merge

mikenairn: Also double check the image is working correctly before merging

coredns/plugin/go.mod
```diff
@@ -0,0 +1,84 @@
+module github.com/kuadrant/coredns-kuadrant
+
+go 1.23.0
+
+toolchain go1.23.4
```
mikenairn, line 5: Hmm, kind of wondering if we should just set this back to 1.22 :thinking: It should probably be the same as what CoreDNS currently set anyway https://github.com/coredns/coredns/blob/master/go.mod#L3

maleck13: Yeah I tihnk this would make sense for now

coredns/plugin/README.md
```diff
@@ -0,0 +1,163 @@
...
+}
+```
+
+```yaml
+apiVersion: kuadrant.io/v1alpha1
+kind: DNSRecord
+metadata:
+  name: example.org
+  labels:
+    kuadrant.io/zone-name: example.org
```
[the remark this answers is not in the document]
mikenairn: renamed

coredns/plugin/Dockerfile
```diff
@@ -0,0 +1,40 @@
+# Build the manager binary
+FROM mirror.gcr.io/library/golang:1.23 AS builder
```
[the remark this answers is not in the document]
mikenairn: reverted, for now

coredns/plugin/go.mod
mikenairn: I have set this back to 1.22.7. I'm unsure about the best way forward here, ideally we wouldn't be restricted in updating things because of downstream build limitations. CoreDNS now set the toolchain to 1.23+ https://github.com/coredns/coredns/blob/master/go.mod#L3-L5 which allows the dependencies with issues to update.

Will come back to this.

coredns/plugin/zone_test.go
```diff
@@ -0,0 +1,222 @@
...
+							Value: "GEO-EU",
+						},
+					},
+				},
+			},
+			expectedRRData: map[string]rrData{
+				"cname.example.com.\t60\tIN\tCNAME\tt.example.com.": {
+					geo: ptr.To("GEO-EU"),
+				},
+			},
```
maleck13, line 203: this is a good set of tests. Should we include inserting a duplicate?

maleck13 (dismissed):
Happy to approve this at this stage given the follow up issues have been defined and this is to support a developer level feature 👍

.github/workflows/build-coredns-images.yaml
mikenairn: Was an issue with the image, Dockerfile had to be updated to include an additional stage to allow the coredns binary to work correctly in nonroot containers https://github.com/Kuadrant/dns-operator/blob/867c019829f5fe09d772ef82be6f4dcf0161a576/coredns/plugin/Dockerfile.

Based on the the [CoreDNS Dockerfile](https://github.com/coredns/coredns/blob/master/Dockerfile).

coredns/plugin/zone_test.go
mikenairn: Yeas i have some other tests coming, will add that as part of the next set of changes.

maleck13 (approved):

PR #415, opened 2025-03-19
coredns(install): Add CoreDNS install and config

### LINKED ISSUE #412: CoreDNS Plugin: Add Kuadrant CoreDNS deploymnet
**What**

Add configuration to deploy the Kuadrant CoreDNS instance. This should include all k8s resources required to run it in cluster.

mikenairn:
Adds a basic setup of CoreDNS using the upstream helm chart https://github.com/coredns/helm wrapped in a kustomization. The kustomization (config/coredns) will install a CoreDNS instance in the "kuadrant-coredns" namespace configured with a test zone (k.example.com) with all required plugins (kuadrant, geoip and metadata) enabled.

Geoip plugin configured to use `edns-subnet` in order to use the subnet option if it's available instead of the source ip making things easier to test locally. Currently uses the demo db embedded in the coredns image (/GeoLite2-City-demo.mmdb).

Adds metallb setup to allow access to the CoreDNS instance from the host if the host allows it.

Add grafana dashboard for CoreDNS https://grafana.com/grafana/dashboards/14981-coredns/.

Add unmonitored coredns installation configuration. Same installation without ServiceMonitor resource so it can be installed without observability existing.

Set the externalTrafficPolicy to Local in the coredns LoadBalancer services to preserve the client source ip.
https://kubernetes.io/docs/tasks/access-application-cluster/create-external-load-balancer/#preserving-the-client-source-ip

closes #412 

Requires #403


maleck13 (dismissed):
I see this is in draft. But is all looks ok to me here. We have been using it for a while now so happy for it to merge

PR #419, opened 2025-03-20
TXT records per hostname + owner

maksymvavilov:
Before, we would create TXT records per [DNSName+recordType](https://github.com/Kuadrant/dns-operator/blob/main/internal/external-dns/registry/txt.go#L241)

Now we have a record per DNSName + record type + ownerID. 
This means that we can have multiple TXTs for the same dnsName, so the changes include: 
1. Change the registry to generate more TXT records 
2. Change to the Endpoint's labels - the keys are now always ownersID, and we aren't using the owner as a separate label 
3. The tool to "unpack" and pack labels - depending on each individual use case, we might want to know if other owners specify the same label. For example, with health checks, we would like to know what other owners think of the health of specific endpoints. 
4. Change to the plan: adapting for the changed structure of labels 
5. Update to e2e tests since TXT records are not using their target to signify ownership. In our cases, it means that targets are empty 
6. Introduction of the interface that every registry must implement - depending on the registry we will need tools to pack and filter labels 

The PR is ready for a review, but there will still be more to this task:
1. Create an `.md` file with the explanation of how the registry works 
2. Make code more readable and friendly for others to look at

### DIFF
--- a/internal/external-dns/plan/conflict.go
+++ b/internal/external-dns/plan/conflict.go
@@ -20,18 +20,22 @@ import (
 	"sort"
 
 	"sigs.k8s.io/external-dns/endpoint"
+
+	"github.com/kuadrant/dns-operator/internal/external-dns/registry"
 )
 
 // ConflictResolver is used to make a decision in case of two or more different kubernetes resources
 // are trying to acquire same DNS name
 type ConflictResolver interface {
 	ResolveCreate(candidates []*endpoint.Endpoint) *endpoint.Endpoint
-	ResolveUpdate(current *endpoint.Endpoint, candidates []*endpoint.Endpoint) *endpoint.Endpoint
+	ResolveUpdate(current *endpoint.Endpoint, owner string, candidates []*endpoint.Endpoint) *endpoint.Endpoint
 	ResolveRecordTypes(key planKey, row *planTableRow) map[string]*domainEndpoints
 }
 
 // PerResource allows only one resource to own a given dns name
-type PerResource struct{}
+type PerResource struct {
+	packer registry.LabelsPacker
+}
 
 // ResolveCreate is invoked when dns name is not owned by any resource
 // ResolveCreate takes "minimal" (string comparison of Target) endpoint to acquire the DNS record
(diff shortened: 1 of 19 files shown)


internal/external-dns/plan/conflict.go
```diff
@@ -48,15 +52,19 @@ func (s PerResource) ResolveCreate(candidates []*endpoint.Endpoint) *endpoint.En
...
+	}
+	currentResource := ownedLabels[endpoint.ResourceLabelKey] // resource which has already acquired the DNS
 	// TODO: sort candidates only needed because we can still have two endpoints from same resource here. We sort for consistency
 	// TODO: remove once single endpoint can have multiple targets
 	sort.SliceStable(candidates, func(i, j int) bool {
 		return s.less(candidates[i], candidates[j])
 	})
 	for _, ep := range candidates {
-		if ep.Labels[endpoint.ResourceLabelKey] == currentResource {
+		if ep.Labels[endpoint.ResourceLabelKey] == currentResource || s.packer.UnpackLabels(ep.Labels)[owner][endpoint.ResourceLabelKey] == currentResource {
```
maleck13, line 67: do we need the same nil check here as is on 57?

internal/external-dns/plan/plan.go
```diff
@@ -417,6 +417,14 @@ type endpointUpdate struct {
 }
 
 func (e *endpointUpdate) ShouldUpdate() bool {
+	desiredTarget := e.desired.Targets[0]
+	currentTarget := e.current.Targets[0]
+	_ = desiredTarget + currentTarget
```
maleck13, line 422: doesn't look to be used?

internal/external-dns/plan/plan.go
```diff
@@ -417,6 +417,14 @@ type endpointUpdate struct {
...
+	updateOwner := shouldUpdateOwner(e.desired, e.current)
+	updateTTL := shouldUpdateTTL(e.desired, e.current)
+	target := targetChanged(e.desired, e.current)
+	updateProvider := shouldUpdateProviderSpecific(e.desired, e.current)
+	_ = updateOwner || updateTTL || target || updateProvider
```
maleck13, line 427: not getting this line

internal/external-dns/plan/conflict.go
maksymvavilov: If we only put endpoints from the registry here, they will never have nil labels, but they _could_ be nil here, so sure

internal/external-dns/plan/plan.go
maksymvavilov: Ah, missed those. Leftowers from running controller in debug mode.

maksymvavilov:
This approach is "too much". 
What I've done here works, but introduces unnecessary changes that won't be needed for a while. 
I still would have it as a reference for the future.

maksymvavilov:
This is replaced by #461

PR #421, opened 2025-03-24
core dns provider. Initial tests

maleck13:

rh-pre-commit.version: 2.2.0


maleck13:
to try this locally:

run the controller locally with core dns installed against a kind cluster

run
```
make local-setup-coredns-credentials TARGET_NAMESPACE=dnstest
k apply -f coredns/examples/dnsrecord-c1.yaml -n dnstest
dig @${coredns1IP} k.example.com +subnet=127.0.200.200

```

make/dnsproviders.mk
```diff
@@ -61,6 +62,13 @@ $(LOCAL_SETUP_AZURE_CREDS):
 	$(call ndef,KUADRANT_AZURE_CREDENTIALS)
 	$(call patch-config,${LOCAL_SETUP_AZURE_CREDS}.template,${LOCAL_SETUP_AZURE_CREDS})
 
+.PHONY: local-setup-coredns-credentials
+local-setup-coredns-credentials:
+	echo "local-setup: creating dns provider for coredns in ${TARGET_NAMESPACE} based on lb IP $(COREDNS_NAMESERVERS)";\
+	$(KUBECTL) delete secret dns-provider-core-dns -n ${TARGET_NAMESPACE};\
+	$(KUBECTL) -n ${TARGET_NAMESPACE} create secret generic dns-provider-core-dns --type=kuadrant.io/coredns --from-literal=NAMESERVERS="$(COREDNS_NAMESERVERS):53" --from-literal=ZONES="k.example.com";\
+	$(KUBECTL) label secret dns-provider-core-dns -n ${TARGET_NAMESPACE} app.kubernetes.io/part-of=dns-operator;\
```
philbrookes, line 70:
```
echo "local-setup: creating dns provider for coredns in ${TARGET_NAMESPACE} based on lb IP $(COREDNS_NAMESERVERS)"
	-$(KUBECTL) delete secret dns-provider-core-dns -n ${TARGET_NAMESPACE}
	$(KUBECTL) -n ${TARGET_NAMESPACE} create secret generic dns-provider-core-dns --type=kuadrant.io/coredns --from-literal=NAMESERVERS="$(COREDNS_NAMESERVERS):53" --from-literal=ZONES="k.example.com"
	$(KUBECTL) label secret dns-provider-core-dns -n ${TARGET_NAMESPACE} app.kubernetes.io/part-of=dns-operator
```

make/dnsproviders.mk
```diff
@@ -76,5 +84,11 @@ local-setup-dns-providers: kustomize ## Create AWS, Azure and GCP DNS Providers
 		echo "local-setup: creating dns provider for azure in ${TARGET_NAMESPACE}";\
 		${KUSTOMIZE} build ${LOCAL_SETUP_AZURE_DIR} | $(KUBECTL) -n ${TARGET_NAMESPACE} apply  -f -;\
 	fi
+
+	echo "local-setup: creating dns provider for coredns in ${TARGET_NAMESPACE} based on lb IP $(COREDNS_NAMESERVERS)";\
+	$(KUBECTL) delete secret dns-provider-core-dns -n ${TARGET_NAMESPACE};\
+	$(KUBECTL) -n ${TARGET_NAMESPACE} create secret generic dns-provider-core-dns --type=kuadrant.io/coredns --from-literal=NAMESERVERS="$(COREDNS_NAMESERVERS):53" --from-literal=ZONES="k.example.com";\
+	$(KUBECTL) label secret dns-provider-core-dns -n ${TARGET_NAMESPACE} app.kubernetes.io/part-of=dns-operator;\
```
philbrookes, line 91:
```
echo "local-setup: creating dns provider for coredns in ${TARGET_NAMESPACE} based on lb IP $(COREDNS_NAMESERVERS)"
	- $(KUBECTL) delete secret dns-provider-core-dns -n ${TARGET_NAMESPACE}
	$(KUBECTL) -n ${TARGET_NAMESPACE} create secret generic dns-provider-core-dns --type=kuadrant.io/coredns --from-literal=NAMESERVERS="$(COREDNS_NAMESERVERS):53" --from-literal=ZONES="k.example.com"
	$(KUBECTL) label secret dns-provider-core-dns -n ${TARGET_NAMESPACE} app.kubernetes.io/part-of=dns-operator
```

coredns/examples/dnsrecord-api-k-example-com_geo_weight.yaml
```diff
@@ -72,10 +72,10 @@ spec:
...
         - 127.0.0.1
-    - dnsName: "*.api.k.example.com"
+    - dnsName: '*.api.k.example.com'
       recordTTL: 60
       recordType: A
       targets:
         - 9.9.9.9
   providerRef:
-    name: dns-provider-credentials-coredns
+    name: dns-provider-core-dns
```
mikenairn, line 81: We should try and keep these consistent with all other providers generated by local-setup "dns-provider-credentials-\<provider name\>"

internal/controller/dnsrecord_controller.go
```diff
@@ -546,7 +550,7 @@ func setStatusConditions(record *v1alpha1.DNSRecord, hadChanges bool, notHealthy
 	}
 
 	// if we haven't published because of the health failure, we won't have changes but the spec endpoints will be empty
-	if record.Status.Endpoints == nil || len(record.Status.Endpoints) == 0 {
+	if len(record.Status.Endpoints) == 0 {
```
mikenairn, line 553: Why was this removed? Was this a required change for coredns?

test/e2e/single_record_test.go
```diff
@@ -77,7 +78,7 @@ var _ = Describe("Single Record Test", Labels{"single_record"}, func() {
 		}
 	})
 
-	It("correctly handles wildcard rootHost values", func(ctx SpecContext) {
+	FIt("correctly handles wildcard rootHost values", func(ctx SpecContext) {
```
mikenairn, line 81:
```suggestion
	It("correctly handles wildcard rootHost values", func(ctx SpecContext) {
```

make/dnsproviders.mk
```diff
@@ -76,5 +84,11 @@ local-setup-dns-providers: kustomize ## Create AWS, Azure and GCP DNS Providers
...
 	fi
+
+	echo "local-setup: creating dns provider for coredns in ${TARGET_NAMESPACE} based on lb IP $(COREDNS_NAMESERVERS)"
+	-$(KUBECTL) delete secret dns-provider-core-dns -n ${TARGET_NAMESPACE}
+	$(KUBECTL) -n ${TARGET_NAMESPACE} create secret generic dns-provider-core-dns --type=kuadrant.io/coredns --from-literal=NAMESERVERS="$(COREDNS_NAMESERVERS):53" --from-literal=ZONES="k.example.com"
+	$(KUBECTL) label secret dns-provider-core-dns -n ${TARGET_NAMESPACE} app.kubernetes.io/part-of=dns-operator
+
 	echo "local-setup: creating dns provider for inmemory in ${TARGET_NAMESPACE}";\
-    ${KUSTOMIZE} build ${LOCAL_SETUP_INMEM_DIR} | $(KUBECTL) -n ${TARGET_NAMESPACE} apply  -f -
+    ${KUSTOMIZE} build ${LOCAL_SETUP_INMEM_DIR} | $(KUBECTL) -n ${TARGET_NAMESPACE} apply  -f -;\
```
mikenairn, line 94: I think we would be better here to make this work in a more similar way to the existing local-setup provider generation for consistency. In this particular target it should just be generating something that will work with what is installed via local-setup (single instance of coredns in kuadrant-coredns). That is what this is doing, but in a different way from using kustomize to generate it.

This also doesn't take into account the creation of a secret which includes multiple name servers, what you would get if ran `make install-coredns-multi`. I see this is a separate though, since we don't do that as part of local-setup, but a target that can generate a secret for a given set of coredns instances (using the labels maybe) would probably be useful.

```
kubectl get service -A -l app.kubernetes.io/name=coredns,app.kubernetes.io/component!=metrics,app.kubernetes.io/part-of=coredns-multi -o json | jq -r '[.items[] | (.status.loadBalancer.ingress[].ip + ":53")] | join(",")'
```

internal/provider/provider.go
```diff
@@ -34,6 +48,10 @@ type Provider interface {
 	DNSZoneForHost(ctx context.Context, host string) (*DNSZone, error)
 
 	ProviderSpecific() ProviderSpecificLabels
+
+	Name() DNSProviderName
+
+	RecordsForHost(ctx context.Context, host string) ([]*externaldnsendpoint.Endpoint, error)
```
mikenairn, line 54: I'm not seeing why we need this new method `RecordsForHost`. Could we not just use the `DomainFilter` on the [provider](https://github.com/Kuadrant/dns-operator/blob/main/internal/controller/dnsrecord_controller.go#L603) (like all the rest) and call the existing `Records()` method? What it does with the filter internally is up to the the implementation.

```
kudrantHost := fmt.Sprintf("%s.%s", dnsRecord.Spec.RootHost, provider.KuadrantTLD)
hosts := []string{kudrantHost, fmt.Sprintf("w.%s", kudrantHost), fmt.Sprintf("g.%s", kudrantHost)}

providerConfig := provider.Config{
    DomainFilter: externaldnsendpoint.NewDomainFilter(hosts),
}
return r.ProviderFactory.ProviderFor(ctx, dnsRecord, providerConfig)
```

```
func (p *CoreDNSProvider) Records(ctx context.Context) ([]*endpoint.Endpoint, error) {
	hosts := p.domainFilter.Filters

	return []*endpoint.Endpoint{}, fmt.Errorf("no impl for core dns")
}
```

internal/controller/dnsrecord_controller.go
maleck13: hmm not sure now. May have been accidental

maleck13: ahh actually I think the nil check is not needed as it will default to 0 if not set

maleck13:
```
type val struct{}

func main() {

	var empty []*val
	fmt.Println(len(empty))
}
```

make/dnsproviders.mk
mikenairn: > I think we would be better here to make this work in a more similar way to the existing local-setup provider generation for consistency. In this particular target it should just be generating something that will work with what is installed via local-setup (single instance of coredns in kuadrant-coredns). That is what this is doing, but in a different way from using kustomize to generate it.

This is what i was meaning https://github.com/Kuadrant/dns-operator/pull/424.

internal/provider/provider.go
```diff
@@ -15,14 +15,28 @@ import (
...
 	requestIDRegexp         = regexp.MustCompile(`request id: [^\s]+`)
 	saxParseExceptionRegexp = regexp.MustCompile(`Invalid XML ; javax.xml.stream.XMLStreamException: org.xml.sax.SAXParseException; lineNumber: [^\s]+; columnNumber: [^\s]+`)
 
 	ErrNoZoneForHost = fmt.Errorf("no zone for host")
+
+	DNSProviderCoreDNS DNSProviderName = "coredns"
+	DNSProviderAWS     DNSProviderName = "aws"
+	DNSProviderAzure   DNSProviderName = "azure"
+	DNSProviderGCP     DNSProviderName = "google"
+	DNSProviderInMem   DNSProviderName = "inmemory"
```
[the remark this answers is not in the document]
maleck13: if you do that you end up with cyclical imports where the factory is importing the aws package and aws package is importing the provider pkg

internal/provider/provider.go
mikenairn: Removed it when @maleck13 wasn't looking

philbrookes (approved):

PR #425, opened 2025-04-01
tests: Update loadbalanced single record e2e test

mikenairn:
Updates the e2e test for a loadbalanced recordset created for a single DNSRecord (it contains endpoints for multiple geos and clusters), all providers should be able to handle this.

Note: These tests will fail for the [coredns provider](https://github.com/Kuadrant/dns-operator/pull/421)  currently, if we don't intend this to work initially for CoreDNS we should hold off on merging this.


mikenairn:
With the planned changes to the CoreDNS provider we should eventually be able to merge this. Will leave it open, as a draft, to remind me to come back to it.

PR #433, opened 2025-04-04
Core dns provider e2e tests

mikenairn:
[coredns(provider): Fix zone record retrieval](https://github.com/Kuadrant/dns-operator/commit/9312cc416569baa2add5b1d0e8ae2119616cdc20)

Removes the RecordsForHost() method from the provider and re-implements
coreds Records() with the same functionality.

The point of the provider.Records() method is to retrieve all the
records as they are currently in the external dns provider. In the case
of the coredns implementation that is the coredns servers themselves,
the "merged and "local" DNSRecords being created are not really relevant
here, these are just internal pieces used to achieve the update in the
provider.

Also note, the current implementation of Records() was just retrieving
the merged record from the currently configured cluster. This won't work
when executing the e2e tests against multiple clusters as it would need
to be configured with the correct cluster context.

The CoreDNS provider is slighty different from others in that it needs
to know the exact host(DNSRecord.spec.RootHost) it's dealing with, and
not just the zone domain, an additional filter was added to the provider
config to allow this to be passed in.

A records with the same dnsname should be merged.

[coredns(provider): Fix e2e tests](https://github.com/Kuadrant/dns-operator/commit/ec99df85cc0dd5e71172be7a29eac193237de8c3)

Update all tests to correctly check the existence of zone endpoints
based on wether or not the txt registry is enabled for a provider or
not. Disabled for CoreDNS provider.

Update checks that do a dns lookup to check all available DNS servers
to ensure the correct one is queried. In the case of coredns we don't
know for sure which one should be used, so we just have to try them all
until one or none work.


internal/provider/coredns/coredns.go
```diff
@@ -173,13 +153,13 @@ func (p *CoreDNSProvider) RecordsForHost(ctx context.Context, host string) ([]*e
...
 	answers := map[string]map[string]*dns.Msg{}
-	p.logger.Info("RecordsForHost", "total nameservers", len(p.nameservers), "hosts", hosts)
+	p.logger.Info("recordsForHost", "total nameservers", len(p.nameservers), "hosts", hosts)
 	for _, nServer := range p.nameservers {
 		p.logger.Info("checking nameserver ", "address", *nServer)
 		nsAnswer, err := p.DNSQueryFunc(hosts, *nServer)
 		if err != nil {
 			//TODO prob need to handle dns errors better here
-			return endpoints, err
+			continue
```
maleck13, line 162: should we at least log here?

internal/provider/coredns/coredns.go
```diff
@@ -128,43 +136,15 @@ func (p *CoreDNSProvider) ProviderSpecific() provider.ProviderSpecificLabels {
 	return provider.ProviderSpecificLabels{}
 }
 
-func (p *CoreDNSProvider) k8sClient() (DNSRecordInterface, error) {
```
maleck13, line 131: kinda glad we removed this

test/e2e/single_record_test.go
```diff
@@ -79,6 +80,9 @@ var _ = Describe("Single Record Test", Labels{"single_record"}, func() {
 	})
 
 	It("correctly handles wildcard rootHost values", func(ctx SpecContext) {
+		if testDNSProvider == provider.DNSProviderCoreDNS.String() {
+			Skip("coredns does not currently handle wildcard rootHosts correctly!!")
```
maleck13, line 84: if we don't have an issue for this we should add one

test/e2e/single_record_test.go
```diff
@@ -686,33 +703,43 @@ var _ = Describe("Single Record Test", Labels{"single_record"}, func() {
...
+
 			Eventually(func(g Gomega, ctx context.Context) {
 				var err error
 				var ips []string
-				if resolver == nil {
+				if nameServers == nil {
 					ips, err = net.LookupHost(testHostname)
 				} else {
-					ips, err = resolver.LookupHost(ctx, testHostname)
+					for _, nameServer := range nameServers {
```
maleck13, line 736: should we not expect every Auth NS to have the ips? rather than just one of them?

mikenairn: Good question, maybe, but it depends. The reason i decided to do this was to allow the tests to work if you specified more nameservers than you were actually testing with. i.e. a run local-setup with :

```
make local-setup DEPLOY=true DEPLOYMENT_SCOPE=namespace DEPLOYMENT_COUNT=2 CLUSTER_COUNT=2
``` 

but i want to run the tests with less: 

```
make test-e2e TEST_DNS_ZONE_DOMAIN_NAME=k.example.com TEST_DNS_PROVIDER_SECRET_NAME=dns-provider-credentials-coredns TEST_DNS_NAMESPACES=kuadrant-dns-operator DEPLOYMENT_COUNT=1 TEST_DNS_CLUSTER_CONTEXTS=kind-kuadrant-dns-local CLUSTER_COUNT=1
```

This isn't an issue for any other provider, it just picks a cluster and deployment to work against, but with coredns we need to have it verify the result using the correct nameserver, and there is no way currently of knowing which one that should be.

test/e2e/single_record_test.go
mikenairn: Yes, we need a follow in on issue for this. I know you had changed this, but it was really just pretending to work. The wildcard entry exists, but the other one is lost so its not really working as expected.

maleck13 (approved):
I see how this is working. It looks good to me. Thanks for taking it on @mikenairn

test/e2e/single_record_test.go
maleck13: good spot.

maleck13: https://github.com/Kuadrant/dns-operator/issues/434

internal/provider/coredns/coredns.go
maleck13: Yeah I think we would expect each NS to be reachable and respond (eventually at least) If one of the NS is not returning results then you may have a functional record but one gateway is likely not receiving traffic

mikenairn: Yeah, thinking on this, i think this wrong. I'm making this change to make testing easier which isn't really the correct solution.

PR #446, opened 2025-04-15
Add install overlay for OpenShift on AWS.

mikenairn:


mikenairn:
Closing for now. Will wait and see where the CoreDNS stuff lands, and how many different variants of installation are going to be needed.

PR #462, opened 2025-05-30
txt records per hostname and owner

maksymvavilov:


internal/external-dns/plan/labels.go
```diff
@@ -1,6 +1,48 @@
...
+import (
+	"slices"
+	"strings"
+)
+
 const (
-	// OwnerLabelDeliminator is a deliminator used between owners in the OwnerLabelKey value when multiple owners are assigned.
-	OwnerLabelDeliminator = "&&"
+	// LabelDelaminator is a default delaminator for labels if a label key has multiple values
+	LabelDelaminator = "&&"
```
philbrookes, line 10: not very important, but this should be: `LabelDelimiter`.

internal/external-dns/plan/labels.go
```diff
@@ -1,6 +1,48 @@
...
+	// SoftDeleteKey indicates that endpoint can be soft deleted
+	SoftDeleteKey = "kuadrant/soft-delete"
+
+	// StopSoftDeleteKey forbids soft deletion of endpoint
+	StopSoftDeleteKey = "kuadrant/stop-soft-delete"
 )
+
+func EnsureLabel(labels, label string) string {
+	labelsSplit := SplitLabels(labels)
+	labelsSplit = append(labelsSplit, label)
```
philbrookes, line 21: might be worth adding `if !slice.ContainsStr(labelsSplit, label)` here?

PR #463, opened 2025-06-09
feat: use provider credentials from operator environment

kevthompson:
Configuring static credentials in Secrets is less secure & inconsistent with the existing credential chains provided by the various cloud providers. Leveraging the default credential chains will allow use of dynamic creds via IRSA or the other workload identity flows.

I tried to update examples & docs, I haven't touched the RFC yet though.


philbrookes:
Hi @kevthompson thanks for the PR.

I'm just taking a look through it now, and was wondering if you could give me some quick step by step instructions on how I can test it out locally?

kevthompson:
> Hi @kevthompson thanks for the PR.
> 
> I'm just taking a look through it now, and was wondering if you could give me some quick step by step instructions on how I can test it out locally?

Sorry for the late reply - you can deploy as normal with the `make local-setup DEPLOY=true` command & the operator will mount the credentials from the existing Secrets. To use the provider with those creds, create a DNSRecord with a `provider` field instead of `providerRef`. For AWS with static creds it would look like this:

```
kind: DNSRecord
metadata:
  name: simple-dns
  namespace: dnstest
spec:
  endpoints:
    - dnsName: ${KUADRANT_SUB_DOMAIN}
      recordTTL: 60
      recordType: A
      targets:
        - 172.32.200.17
  # new field, mutually exclusive with `providerRef`
  provider:
    name: aws  # one of: aws, azure, google
  rootHost: ${KUADRANT_SUB_DOMAIN}
  ```

I haven't set it up to use IRSA, you'd just need to add the usual annotation to the operator's ServiceAccount.

If you're trying to use Azure or Google then you'd swap the provider name. They also take a couple of optional arguments:
Azure:
  - `.spec.provider.azureConfigFile` - location of the azure-creds.json file, you won't need to set this for the local setup command
Google:
  - `.spec.provider.googleProject` - ID for your Google Cloud Project, this is required when using GCP.

CoreDNS: passing these values from the spec instead of the secret
  `.spec.provider.corednsNameserver`
  `.spec.provider.corednsZones`

philbrookes:
Hey @kevthompson I will close this PR for now as it is quite old, and needs a rebase.

If you have the time to rebase it, please re-open this PR and we will take a look at it.

Issue #325, opened 2024-12-04
bug: plan calculate desired can panic


mikenairn:
Plan calculate desired can panic in certain cases. 

Observed only with Azure at scale (10+ records for the same host)

```
controller-manager-57db8d94d7-6gdr4 manager panic: runtime error: index out of range [4] with length 4 [recovered]
controller-manager-57db8d94d7-6gdr4 manager     panic: runtime error: index out of range [4] with length 4
controller-manager-57db8d94d7-6gdr4 manager 
controller-manager-57db8d94d7-6gdr4 manager goroutine 290 [running]:
controller-manager-57db8d94d7-6gdr4 manager sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile.func1()
controller-manager-57db8d94d7-6gdr4 manager     /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.18.0/pkg/internal/controller/controller.go:111 +0x1e5
controller-manager-57db8d94d7-6gdr4 manager panic({0x1dab8a0?, 0xc00084eab0?})
controller-manager-57db8d94d7-6gdr4 manager     /usr/local/go/src/runtime/panic.go:770 +0x132
controller-manager-57db8d94d7-6gdr4 manager github.com/kuadrant/dns-operator/internal/external-dns/plan.(*managedRecordSetChanges).calculateDesired(0xc0007af0c8, 0xc0006c7ac0)
controller-manager-57db8d94d7-6gdr4 manager     /workspace/internal/external-dns/plan/plan.go:542 +0x9b3
controller-manager-57db8d94d7-6gdr4 manager github.com/kuadrant/dns-operator/internal/external-dns/plan.(*managedRecordSetChanges).Calculate(0xc0007af0c8)
controller-manager-57db8d94d7-6gdr4 manager     /workspace/internal/external-dns/plan/plan.go:452 +0x28b
controller-manager-57db8d94d7-6gdr4 manager github.com/kuadrant/dns-operator/internal/external-dns/plan.(*Plan).Calculate(0xc000b5ac60)
controller-manager-57db8d94d7-6gdr4 manager     /workspace/internal/external-dns/plan/plan.go:345 +0x8ff
[...]

PR #468, opened 2025-07-02
prevent index out of range in plan

maksymvavilov:
fixes #https://github.com/Kuadrant/dns-operator/issues/325

Fixes a bug in the modified external-dns plan logic when manipulating endpoint target values.


mikenairn:
Is it possible to add a test to re-create the panic and see this fix work? If it is caused by calling `removeEndpointTarget` in the loop as i suspect, then it should be possible to add a plan test for it.

mikenairn (approved):

maksymvavilov:
I'll create a follow-up for the test

Issue #346, opened 2025-01-17
registry: Create per record+owner metadata records (TXT Registry)


mikenairn:
**What**

Update the TXT registry to create a registry entry per owner/record combination instead of a single record entry that is updated with all owners.

**Pros:**

* Each DNSRecord will only ever be updating it's own TXT records avoiding all possible conflicts on TXT record updates between records/clusters updating the same record set.
* Removes the limit of ~20 records max for the same dns name (rootHost) . Upper limit would only be limited by the number of records allowed in a zone.
* Would resolve a lot of issues we hit when deleting records https://github.com/Kuadrant/dns-operator/issues/181.
* More possibility for expanding the metadata saved in these records. Health, heartbeat etc.. 
* Is more like how you would do it if we were talking to a real database or cache i.e. redis.

**Cons:**

* More TXT records, one per owner on every record that is shared.

**Considerations**

* Need to migrate from the old format to the new
* Leave the old txt records in place to allow for older DNS operators not yet upgraded

maleck13:
This sounds like the ideal approach to me where the we remove more shared records and increase the upper limit. Any thoughts on how the migration would happen? I assume there would be no impact on existing non txt records right?
I think we should prioritise this change.

PR #475, opened 2025-07-10
evict name maper from the TXT registry code

maksymvavilov:
It was excluded from #346 for the ease of review. Just moving the mapper out.
It was with the registry because that is how `external-dns` [has](https://github.com/kubernetes-sigs/external-dns/blob/master/registry/txt.go) it. For them, it makes sense as there are multiple registry types. For us it _might_ make it easier to understand and maintain TXT registry.


maleck13 (approved):
Looks like  straight change all good

Issue #210, opened 2024-08-11
bug: dnsrecord never stops trying to update on conflict


mikenairn:
**What**

The dnsrecord controller no longer stops trying to update the record in the provider zone when it is trying to update the same generation too often. Look like a regression introduced by https://github.com/Kuadrant/dns-operator/pull/100, at least i can't find the equivalent of this anywhere now https://github.com/Kuadrant/dns-operator/pull/100/files#diff-adacb6a4589c65c2d853c0f59ca02d6b8fb50bb388b41de1d13e99b0d128cfe1L405
```.go
if dnsRecord.Status.WriteCounter < WriteCounterLimit {
   dnsRecord.Status.WriteCounter++
   wrtiteCounter.WithLabelValues(dnsRecord.Name, dnsRecord.Namespace).Inc()
   logger.V(1).Info("Changes needed on the same generation of record")
} else {
```

PR #483, opened 2025-07-11
gh-210 backoff on conflict

maksymvavilov:
Back off on conflict (writes to the provider on the same generation). 
Also update the conflict test to check for it

### DIFF
--- a/internal/controller/dnsrecord_controller.go
+++ b/internal/controller/dnsrecord_controller.go
@@ -384,20 +384,27 @@ func (r *DNSRecordReconciler) updateStatus(ctx context.Context, previous, curren
 		return reconcile.Result{RequeueAfter: requeueIn}, nil
 	}
 
+	readyCond := meta.FindStatusCondition(current.GetStatus().Conditions, string(v1alpha1.ConditionTypeReady))
+
 	// success
 	if hadChanges {
 		// generation has not changed but there are changes.
 		// implies that they were overridden - bump write counter
+		requeueTime = randomizedValidationRequeue
 		if !generationChanged(current.GetDNSRecord()) {
 			current.GetStatus().WriteCounter++
 			metrics.WriteCounter.WithLabelValues(current.GetDNSRecord().GetName(), current.GetDNSRecord().GetNamespace()).Inc()
 			logger.V(1).Info("Changes needed on the same generation of record")
+
+			// we weren't ready before - back off
+			if readyCond.Status == metav1.ConditionFalse {
+				requeueTime = exponentialRequeueTime(current.GetStatus().ValidFor)
+			}
 		}
-		requeueTime = randomizedValidationRequeue
 	} else {
 		logger.Info("All records are already up to date")
 
-		readyCond := meta.FindStatusCondition(current.GetStatus().Conditions, string(v1alpha1.ConditionTypeReady))
+		readyCond = meta.FindStatusCondition(current.GetStatus().Conditions, string(v1alpha1.ConditionTypeReady))
 
 		// this is the first reconciliation current.GetStatus().ValidFor is not set
 		if readyCond == nil {
--- a/internal/controller/dnsrecord_controller_test.go
+++ b/internal/controller/dnsrecord_controller_test.go
@@ -82,7 +82,7 @@ var _ = Describe("DNSRecordReconciler", func() {
 				ProviderRef: &v1alpha1.ProviderRef{
 					Name: dnsProviderSecret.Name,
 				},
-				Endpoints: getTestEndpoints(testHostname, []string{"127.0.0.1"}),
+				Endpoints: NewTestEndpoints(testHostname).Endpoints(),
 			},
 		}
 	})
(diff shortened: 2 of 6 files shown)


internal/controller/dnsrecord_controller.go
```diff
@@ -294,8 +294,10 @@ func (r *DNSRecordReconciler) updateStatus(ctx context.Context, previous, curren
 			current.Status.WriteCounter++
 			metrics.WriteCounter.WithLabelValues(current.Name, current.Namespace).Inc()
 			logger.V(1).Info("Changes needed on the same generation of record")
+			requeueTime = exponentialRequeueTime(randomizedValidationRequeue.String())
+		} else {
+			requeueTime = randomizedValidationRequeue
 		}
-		requeueTime = randomizedValidationRequeue
```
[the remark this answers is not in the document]
maksymvavilov: That is a good point. 
Added a statement to check if the condition was ready, and if so, set requeue to a small amount. 
In the situation you described, it will now have the least interval after detecting conflict again.

internal/controller/dnsrecord_controller.go
```diff
@@ -294,13 +296,20 @@ func (r *DNSRecordReconciler) updateStatus(ctx context.Context, previous, curren
...
+
+			// if it is the same gen, but we had ready cond set before - we don't want to keep the "long"
+			// validation time from before
+			if readyCond.Status == metav1.ConditionTrue {
+				requeueTime = randomizedValidationRequeue
+			} else {
+				requeueTime = exponentialRequeueTime(current.Status.ValidFor)
+			}
+		} else {
+			requeueTime = randomizedValidationRequeue
```
maleck13, line 308: minor: could this be made a little more readable with a the following change?

```go
requeueTime = exponentialRequeueTime(current.Status.ValidFor)
if !generationChanged(current) {
	current.Status.WriteCounter++
	metrics.WriteCounter.WithLabelValues(current.Name, current.Namespace).Inc()
	logger.V(1).Info("Changes needed on the same generation of record")

	// if it is the same gen, but we had ready cond set before - we don't want to keep the "long"
	// validation time from before
	if readyCond.Status == metav1.ConditionTrue {
		requeueTime = randomizedValidationRequeue
	}
}

```

maksymvavilov: Changed. Yep, it is confusing enough. I tried separating it into its own logic (the `calculateRequeue()`), but it did not help. Comments should help, but even with them, it is not ideal.

maleck13 (dismissed):
@mikenairn perhaps you could also give this another once over now?

internal/controller/dnsrecord_controller_test.go
```diff
@@ -527,6 +527,22 @@ var _ = Describe("DNSRecordReconciler", func() {
...
+
+			err = k8sClient.Get(ctx, client.ObjectKeyFromObject(dnsRecord), dnsRecord)
+			g.Expect(err).NotTo(HaveOccurred())
+
+			newRequeue, err = time.ParseDuration(dnsRecord.Status.ValidFor)
+			g.Expect(err).NotTo(HaveOccurred())
+
+			g.Expect(oldRequeue.Milliseconds()).To(BeNumerically("<", newRequeue.Milliseconds()))
+		}, TestTimeoutMedium, time.Second).Should(Succeed())
+
```
mikenairn, line 545: I'm not sure this test ("should detect a conflict and the resolution of a conflict") is providing any value, the two records that are being created don't conflict, tbh I'm not even sure off the top of my head how you can cause a conflict when using the inmem provider.

Point of this, as i understand it, is when two records are able to update the same field to a different value (aws default geo for example) but two A records will just merge these days and don't actually conflict. They will cause a write count increase, but that's expected

mikenairn (commented):
This has been approved and so i have no issues merging it if others think it's fine. The test I'm fairly sure is testing nothing useful, we need to look into how we can actually re-produce a "conflict" that would fall into this code.

Generally i find the whole requeue process confusing with the amount of variables involved and really can't tell just looking at it what it's going to do in any given situation.

internal/controller/dnsrecord_controller_test.go
mikenairn: Actually, you can get it to properly "conflict" if you set different TTL values on the two records. I would suggest making that change here.

maksymvavilov: This test is, indeed, checks for nothing. Seems like it was relying on a bug or something with the inmemory provider to recreate a conflict. I'll change a test in here to do a conflict. 

As for the mess in the requeue process I've created #513 yesterday. It could be the excuse to revisit the logic and improve it :thinking:

maksymvavilov: Changed the test to be a valid conflict test. With the changes, it didn't make sense to keep the `testEndpoints()` as it was, so I made a slight modification. It will also make it easier to adjust those in the future

philbrookes (approved):

Issue #447, opened 2025-04-16
Beta level Core DNS support


maleck13:
This feature leads directly on from the developer preview. The issues associated with it, should define what is needed to get to a more supportable and feature full Core DNS offering that could be used in none production environments for full evaluation. 
Once this is completed, there should be very little in the way of breaking changes. Not that we wont make them if needed at this stage, but ideally all is well known and we are  moving to a more complete feature set for Core DNS support.

PR #488, opened 2025-07-16
gh-447 optional provider ref

maksymvavilov:


internal/controller/dnsrecord_controller.go
```diff
@@ -193,6 +192,41 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+			if err != nil {
+				setDNSRecordCondition(dnsRecord, string(v1alpha1.ConditionTypeReady), metav1.ConditionFalse,
+					"DNSProviderError", fmt.Sprintf("The default dns provider secret could not be loaded: %v", err))
+				return r.updateStatus(ctx, previous, dnsRecord, probes, false, []string{}, err)
+			}
+
+			// no secrets
+			if len(defaultSecretList.Items) == 0 {
+				setDNSRecordCondition(dnsRecord, string(v1alpha1.ConditionTypeReady), metav1.ConditionFalse,
+					"DNSProviderError", fmt.Sprintf("No default provider secret labaled %s was found", v1alpha1.DefaultProviderSecretLabel))
```
maleck13, line 214:
```suggestion
					"DNSProviderError", fmt.Sprintf("No default provider secret was found", v1alpha1.DefaultProviderSecretLabel))
```

internal/controller/dnsrecord_controller.go
```diff
@@ -193,6 +192,41 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
 	}
 
+	// Ensure we have provider secret
+	if !dnsRecord.HasProviderSecretAssigned() {
+		if dnsRecord.Spec.ProviderRef.Name != "" {
+			dnsRecord.Status.ProviderRef = dnsRecord.Spec.ProviderRef
+		} else {
+			// try to find the default secret
+			defaultSecretList := &v1.SecretList{}
+			err = r.Client.List(ctx, defaultSecretList, client.InNamespace(dnsRecord.Namespace), client.HasLabels{v1alpha1.DefaultProviderSecretLabel})
```
mikenairn, line 202: Should this check for the label having a "truthy" value? Assume "kuadrant.io/default-provider=false" would also work here.

api/v1alpha1/dnsrecord_types.go
```diff
@@ -103,7 +103,8 @@ type DNSRecordSpec struct {
 	// +kubebuilder:validation:Pattern=`^(?:[\w\-.~:\/?#[\]@!$&'()*+,;=]+)\.(?:[\w\-.~:\/?#[\]@!$&'()*+,;=]+)$`
 	RootHost string `json:"rootHost"`
 
-	// providerRef is a reference to a provider secret.
+	// ProviderRef is a reference to a provider secret.
+	// +optional
 	ProviderRef ProviderRef `json:"providerRef"`
```
mikenairn, line 108: Should this be a pointer if its now optional?

api/v1alpha1/dnsrecord_types.go
```diff
@@ -103,7 +103,8 @@ type DNSRecordSpec struct {
 	// +kubebuilder:validation:Pattern=`^(?:[\w\-.~:\/?#[\]@!$&'()*+,;=]+)\.(?:[\w\-.~:\/?#[\]@!$&'()*+,;=]+)$`
 	RootHost string `json:"rootHost"`
 
-	// providerRef is a reference to a provider secret.
+	// ProviderRef is a reference to a provider secret.
```
mikenairn, line 106: Separate task, but we should go through all these and make them consistent one way or the other, we have a mix at the moment. Probably caused by the fact these get used in the CRD docs so the capitalization of the field name looks wrong there, although it's correct here.

mikenairn (commented):
Changes look good to me.

api/v1alpha1/dnsrecord_types.go
maksymvavilov: Made the `Spec` one a pointer (the `Status` reference is not a pointer) following API conventions

maleck13 (approved):
changes look good to me /lgtm

PR #489, opened 2025-07-18
Cluster aware dnsrecord controller

### LINKED ISSUE #484: Add multi cluster secrets and controller
**What**

Add the required controller logic to watch for multi cluster secrets and register them as "clusters" within the dns operator.

**How**

* Add watch on secrets in a specified namespace e.g. kuadrant-system with a given label e.g. `kuadrant.io/multicluster-kubeconfig: "true"`
* The namespace and label should be configurable via flags to the controller:
  * `--cluster-secret-namespace` The Namespace to look for cluster secrets. default: kuadrant-system
  * `--cluster-secret-label` The label that identifies a Secret resource as a cluster secret. default: kuadrant.io/multicluster-kubeconfig
* Cluster secrets are expected to contain kubeconfig data, when one is found it should be registered as a "cluster" and a watch for DNSRecord resources started inside a runnable go routine.
* If a cluster secret is removed, any cluster resource and watch created should be removed.
* When a DNSRecord resource gets an update it should call the "reconcile function". This task will just stub this out for now.
* Add a new role to the dns operator that contains the minimum required permissions needed for a "primary" cluster to process the DNSRecord resources on a "remote" cluster.
```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: remote-cluster-role
rules:
- apiGroups:
    - kuadrant.io
  resources:
    - dnsrecords
  verbs:
    - get
    - list
    - watch
- apiGroups:
    - kuadrant.io
  resources:
    - dnsrecords/status
  verbs:
    - get
    - patch
    - update
```

Note: For the [POC]( https://github.com/Kuadrant/dns-operator/blob/multicluster-poc/cmd/main.go#L156) i used the [multicluster controller runtime package](https://github.com/kubernetes-sigs/multicluster-runtime) which is created by the KCP folks, but experimental. It did work well, and was easily integrated using a [kubeconfig provider](https://github.com/kubernetes-sigs/multicluster-runtime/pull/45), but we should talk about the use of this and alternatives before committing to it. 

**Expected Outcomes**

* Update local-setup to include a means to create two kind clusters clusters with one configured as a "primary". 
* Re-evaluate [multicluster controller runtime package](https://github.com/kubernetes-sigs/multicluster-runtime) and if we can use it.
* Multi cluster secret controller logic added
* Watch maintained on DNSRecords on the remote clusters with appropriate rbac defined.
* Stub function for reconcile called on updates (just log the requests for now).

Note: Does not require DNSRecord reconciliation logic to be completed, follow on tasks will deal with that.

mikenairn:
closes #484 

Adds the ability for the dns controller to process DNSRecord resources on remote clusters as well as its own, makes use of [multicluster runtime](https://github.com/kubernetes-sigs/multicluster-runtime), and a [kubeconfig provider](https://github.com/kubernetes-sigs/multicluster-runtime/tree/main/examples/kubeconfig) to do this.

Adds a watch for Secrets in a configurable namespace (default: dns-operator-system) with a specific label (default: kuadrant.io/multicluster-kubeconfig).
Adds a new controller "multicluster_dnsrecord_controller"  that is intended to process all reconcile requests of DNSRecord resources coming from a remote cluster. This controller does any processing required specific to "remote" records and will eventually delegate the reconcile request on to the normal reconciler.

**Verification**

Manually create clusters and configure them, see [multi-cluster](https://github.com/Kuadrant/dns-operator/blob/cluster_aware_dns_controller/hack/multi-cluster.md) doc.

-- or --

Use the `multicluster-local-setup` make target to create two kind clusters with one added as a remote on the other.  
```
make multicluster-local-setup
```

Check the cluster secrets:
```
kubectl get secrets -A -l kuadrant.io/multicluster-kubeconfig=true --show-labels --context kind-kuadrant-dns-local-1
NAMESPACE             NAME                        TYPE     DATA   AGE     LABELS
dns-operator-system   kind-kuadrant-dns-local-2   Opaque   1      5m14s   kuadrant.io/multicluster-kubeconfig=true
```

Tail the first clusters logs:
```
kubectl stern deployment/dns-operator-controller-manager -n dns-operator-system --context kind-kuadrant-dns-local-1
```

Verify the second cluster is registered with the first by checking the logs:
```
{"level":"info","ts":"2025-07-23T14:41:14Z","logger":"kubeconfig-provider","msg":"Successfully added cluster","cluster":"kind-kuadrant-dns-local-2","secret":"dns-operator-system/kind-kuadrant-dns-local-2"}
{"level":"info","ts":"2025-07-23T14:41:14Z","logger":"kubeconfig-provider","msg":"Successfully engaged manager","cluster":"kind-kuadrant-dns-local-2","secret":"dns-operator-system/kind-kuadrant-dns-local-2"}
```

Create a dnsrecord on the second cluster (kind-kuadrant-dns-local-2):
```
kubectl apply -f coredns/examples/dnsrecord-api-k-example-com.yaml -n dnstest --context kind-kuadrant-dns-local-2
```
Verify you see requests made on the first (kind-kuadrant-dns-local-1):
```
{"level":"info","ts":"2025-07-23T14:48:57Z","msg":"Reconciling Remote DNSRecord","controller":"remote-dnsrecord-controller","controllerGroup":"kuadrant.io","controllerKind":"DNSRecord","reconcileID":"3f1d939f-cde4-4dee-a565-72d72c078219","cluster":"kind-kuadrant-dns-local-2"}
```

**Requires:**

* https://github.com/Kuadrant/dns-operator/pull/496
* https://github.com/Kuadrant/dns-operator/pull/497

**ToDo**

* Add some docs

### DIFF
--- a/hack/multi-cluster.md
+++ b/hack/multi-cluster.md
@@ -1,0 +1,39 @@
+## Multi cluster
+
+### Cluster Setup
+
+Create two clusters with the DNS Operator deployed
+```shell
+make local-setup CLUSTER_COUNT=2 DEPLOY=true
+```
+
+Tail the operator logs on Cluster 1 (Optional)
+```shell
+kubectl stern deployment/dns-operator-controller-manager -n dns-operator-system --context kind-kuadrant-dns-local-1
+```
+
+Tail the operator logs on Cluster 2 (Optional)
+```shell
+kubectl stern deployment/dns-operator-controller-manager -n dns-operator-system --context kind-kuadrant-dns-local-2
+```
+
+#### Add cluster 2 kubeconfig secret to cluster 1
+
+When testing locally with kind, depending on how you are running the dns controller, locally (make run), or on cluster (make deploy), you are going to need to generate the cluster secret with different `kubeconfig` data to allow correct communication between clusters.
+
+#####  Remote("make deploy")
+```shell
+make kubeconfig-secret-create-kind-internal NAMESPACE=dns-operator-system NAME=kind-kuadrant-dns-local-2 TARGET_CONTEXT=kind-kuadrant-dns-local-1 REMOTE_CONTEXT=kind-kuadrant-dns-local-2 SERVICE_ACCOUNT=dns-operator-remote-cluster
+```
+>Note: This is running in a container on the kind cluster in order for the script to properly communicate with the remote and primary servers and generate a config that will work from inside the kind cluster.
+
+##### Locally("make run")
+```shell
+make kubeconfig-secret-create NAMESPACE=dns-operator-system NAME=kind-kuadrant-dns-local-2 TARGET_CONTEXT=kind-kuadrant-dns-local-1 REMOTE_CONTEXT=kind-kuadrant-dns-local-2 SERVICE_ACCOUNT=dns-operator-remote-cluster
+```
+
+Verify cluster secret exists on cluster 1 only
+```shell
+kubectl get secrets -A -l kuadrant.io/multicluster-kubeconfig=true --show-labels --context kind-kuadrant-dns-local-1
+kubectl get secrets -A -l kuadrant.io/multicluster-kubeconfig=true --show-labels --context kind-kuadrant-dns-local-2
+```


cmd/main.go
```diff
@@ -108,9 +115,11 @@ func main() {
 
 	printControllerMetaInfo()
 
+	ctx := ctrl.SetupSignalHandler()
+
 	var watchNamespaces = "WATCH_NAMESPACES"
```
maleck13, line 120: does this do anything in a multi-cluster setup. I presume not given it is an operator sdk thing

cmd/main.go
```diff
@@ -100,6 +103,10 @@ func main() {
 		"The minimal timeout between calls to the DNS Provider"+
 			"Controls if we commit to the full reconcile loop")
 	flag.Var(&providers, "provider", "DNS Provider(s) to enable. Can be passed multiple times e.g. --provider aws --provider google, or as a comma separated list e.g. --provider aws,gcp")
+
+	flag.StringVar(&clusterSecretNamespace, "cluster-secret-namespace", "dns-operator-system", "The Namespace to look for cluster secrets.")
```
maleck13, line 107: when installed via Kuadrant we will need to set this via to be kuadrant-system. Is it possible to use the downward api to set this as an env var to the namespace we are installed in by default?
https://kubernetes.io/docs/concepts/workloads/pods/downward-api/#available-fields

mikenairn: Possibly, i was wondering about this a bit. Originally we said make the default "kuadrant-sytem", but that doesn't really make sense as a default in this repo.

If we can default to using the current namespace when running inside a pod that would be ideal, but we would still need to allow it be set manually, or overridden.

cmd/main.go
mikenairn: Good question, I'd say no, but i haven't really thought about what this would actually mean if you are reconciling remote records. Would seem odd to reconcile all namespaces on a remote but limit to a single one on the local cluster. :thinking: 

The multi cluster manager does use the same options, i haven't dug through the code, but might be worth trying just to know for sure what it actually does if you set this.

cmd/main.go
mikenairn: Updated https://github.com/Kuadrant/dns-operator/pull/489/commits/a38518245290c4780c001748f1ab7a1c4c0163c5, works well, should now automatically set that correctly when installed via kuadrant operator.

cmd/main.go
mikenairn: Need to think about this, currently this will impact on the behaviour. If `WATCH_NAMESPACES` is set, it is set on the local manager options which is used for any local watches, including the cluster secrets. This means that whatever  `CLUSTER_SECRET_NAMESPACE` is set to must also be in this list in order for it to discover local cluster kubeconfig secrets as expected.

maleck13 (approved):
This looks good to merge now 👍

Issue #487, opened 2025-07-15
Update finalizer removal logic to check status condition for reason "ProviderEndpointsRemoved"


mikenairn:
**What**

Update the DNSRecord API to add a new reason to the "Ready" status condition to indicate that it's not ready because all endpoints have been removed from the provider.
```
ConditionReasonProviderEndpointsRemoved ConditionReason = "ProviderEndpointsRemoved"
```
The decision to remove the finalizer from the record during deletion will be based on this new expected status existing after a successful cleanup on the provider. 

**Why**

In order to support multi cluster reconciliation of records we need to be able to signal that a remote record can be cleaned up by the remote clusters dns controller after a primary cluster has successfully dealt with the deletion and removed endpoints from the target provider. 

**How**

* When a record is deleting, the Ready condition should be set to false immediately with the reason endpoints deleting.
* When all the records are verified as removed from the provider, update the condition to: ProviderEndpointsRemoved.
* This should always be the last status update when reconciling a deleting record.
* Update deletion logic to only remove the finalizer when a `Ready=false` status condition exists with the reason `ProviderEndpointsRemoved`.

No e2e tests need to be updated for this.

Boomatang:
@mikenairn When `Ready=false` and reason is `ProviderEndpointRemoved` is set, is there any thing else that should be removed from the status. The below status is from a record that was marked for deletion, but also had a second finalizer blocking the deletion. To me is feels a bit strange to have the `relatedEndpoints`, `zoneDomainName` and `zoneID` in the status when the records have being removed from the DNS provider. 

Can you advise.

```yaml
status:
  conditions:
  - lastTransitionTime: "2025-07-21T09:53:19Z"
    message: DNS records removed from provider
    observedGeneration: 2
    reason: ProviderEndpointsRemoved
    status: "False"
    type: Ready
  observedGeneration: 2
  ownerID: 7ba4094k
  queuedAt: "2025-07-21T09:56:55Z"
  relatedEndpoints:
  - dnsName: echo.jf.hcpapps.net
    labels:
      owner: 7ba4094k
    recordTTL: 60
    recordType: A
    targets:
    - 172.31.201.1
    - 172.31.202.1
  validFor: 6.773s
  writeCounter: 41
  zoneDomainName: jf.hcpapps.net
  zoneID: /hostedzone/Z0676175C89CAO9HT1KJ

```

mikenairn:
[...]

PR #491, opened 2025-07-21
Update finalizer removal logic to check status condition for "ProviderEndpointsRemoved"

Boomatang:
Fixes: #487

# Summary
This PR updates the DNSRecord finalizer removal logic to check for the ProviderEndpointsRemoved status condition reason before removing the finalizer. This change supports multi-cluster reconciliation of DNS records by ensuring that finalizers are only removed when DNS endpoints have been successfully removed from the provider.

# Changes Made
## API Changes

Added new condition reasons:
- ConditionReasonProviderEndpointsRemoved: Indicates DNS records have been successfully removed from the provider
- ConditionReasonProviderEndpointsDeletion: Indicates DNS records are currently being deleted from the provider

Added status helper methods:
- DNSRecordStatus.ProviderEndpointsRemoved(): Returns true if the ready condition has reason "ProviderEndpointsRemoved"
- DNSRecordStatus.ProviderEndpointsDeletion(): Returns true if the ready condition has reason "ProviderEndpointsDeletion"

## Controller Logic Changes
Finalizer removal logic:
- Finalizer removal is now outside the deletion timestamp logic 
- Finalizer is now removed only when ProviderEndpointsRemoved() returns true
- Finalizer can be removed in a multi cluster setup by the primary updating the resource status
- Locic add to stop requeue loops when endpoints are already removed when there are secondary finalizers on the resource

Deletion workflow:
- Sets status to ProviderEndpointsDeletion when deletion begins
- Existing DNS record cleanup remains on changed
- Sets status to ProviderEndpointsRemoved when cleanup completes
- Clears zone-related status fields (endpoints, domain name, zone ID, owner ID)
- Finalizer removal is handled separately based on status condition

# Verification 

- Install this branch.
- Create a valid DNSrecord resource.
- Add a second finalizer, stops removal of record allowing the confirmation of status
e.g. `kubectl patch dnsrecord echo.apps.hcpapps.net -n dnstest --type='merge' -p '{"metadata":{"finalizers":["user-applied"]}}'`
- Set up a watch of the DNSrecord status
- Wait for condition to be in a ready state
- Delete the DNSrecord.
- **Expect:** During DNS provider records removal ready condition should be false with a reason of `ProviderEndpointsDeletion`
- Once DNS provider records have being removed
- **Expect:** ready condition to be `ProviderEndpointsRemoved`
- **Expect:** Finalizer `kuadrant.io/dns-record` to be removed from the resource

### DIFF
--- a/internal/controller/dnsrecord_controller.go
+++ b/internal/controller/dnsrecord_controller.go
@@ -120,8 +120,33 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	// Update the logger with appropriate record/zone metadata from the dnsRecord
 	ctx, logger = r.setLogger(ctx, baseLogger, dnsRecord)
 
+	if dnsRecord.Status.ProviderEndpointsRemoved() {
+		logger.V(1).Info("Status ProviderEndpointRemoved is true, finalizer can be removed")
+		logger.Info("Removing Finalizer", "finalizer_name", DNSRecordFinalizer)
+		controllerutil.RemoveFinalizer(dnsRecord, DNSRecordFinalizer)
+		if err = r.Update(ctx, dnsRecord); client.IgnoreNotFound(err) != nil {
+			if apierrors.IsConflict(err) {
+				return ctrl.Result{Requeue: true}, nil
+			}
+			return ctrl.Result{}, err
+		}
+	}
+
 	if dnsRecord.DeletionTimestamp != nil && !dnsRecord.DeletionTimestamp.IsZero() {
 		logger.Info("Deleting DNSRecord")
+		if dnsRecord.Status.ProviderEndpointsRemoved() {
+			// If the status is set there is no clean up work required.
+			// This stops a requeue loop if there are other finalizers add to the resource.
+			// i.e. user generated finalizers.
+			return ctrl.Result{Requeue: false}, nil
+		}
+
+		if !dnsRecord.Status.ProviderEndpointsDeletion() {
+			setDNSRecordCondition(dnsRecord, string(v1alpha1.ConditionTypeReady), metav1.ConditionFalse, string(v1alpha1.ConditionReasonProviderEndpointsDeletion), "DNS records are being deleted from provider")
+			result, err := r.updateStatus(ctx, previous, dnsRecord, probes, true, []string{}, err)
+			return result, err
+		}
+
 		if dnsRecord.HasDNSZoneAssigned() {
 			// Create a dns provider with config calculated for the current dns record status (Last successful)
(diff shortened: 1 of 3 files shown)


internal/controller/dnsrecord_controller.go
```diff
@@ -120,8 +120,33 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+		logger.V(1).Info("Status ProviderEndpointRemoved is true, finalizer can be removed")
+		logger.Info("Removing Finalizer", "finalizer_name", DNSRecordFinalizer)
+		controllerutil.RemoveFinalizer(dnsRecord, DNSRecordFinalizer)
+		if err = r.Update(ctx, dnsRecord); client.IgnoreNotFound(err) != nil {
+			if apierrors.IsConflict(err) {
+				return ctrl.Result{Requeue: true}, nil
+			}
+			return ctrl.Result{}, err
+		}
+	}
```
mikenairn, line 133: :thinking: I had expected this to be inside the check for a deletion, since we only ever want to remove the finalizer when we are deleting. I guess this is fine, and not really arguing about it, but looks like you could replace [this](https://github.com/Kuadrant/dns-operator/pull/491/files#diff-adacb6a4589c65c2d853c0f59ca02d6b8fb50bb388b41de1d13e99b0d128cfe1R137-R142), with the above no?

mikenairn (commented):
I think this looks good, just a question about where we are removing the finalizer.

Running the e2e test also https://github.com/Kuadrant/dns-operator/actions/runs/16420024159

internal/controller/dnsrecord_controller.go
Boomatang: I was thinking about this from the multi cluster point. If the DNSrecord is on remote, I can see a case where the primary cleans up the DNS entries with out the the remote DNSrecord being marked for deletion. In that case the remote dns-operator would need to remove the finalizer from the resource.  

The spot where you said it could be placed is valid depending on how this is meant to work in multi cluster setup. 

1. If an external entity updates the status to  `ProviderEndpointsRemoved`, the dns-operator should remove the finalizer.
2. If an external entity updates the status to  `ProviderEndpointsRemoved` and the DNSrecord is marked for deletion, the dns-operator should remove the finalizer.

Which of the two statements is the correct behavior?

mikenairn (commented):
Looks good to me :+1: 

Tests passing should be enough, can't really add additional tests to assert the status updates since it will remove the record.

Running the e2e [test ](https://github.com/Kuadrant/dns-operator/actions/runs/16516198526) for completeness. Can merge when that passes.

mikenairn (approved):

PR #500, opened 2025-07-23
fix: panic when using default secret

mikenairn:
The spec providerRef can be nil and the watch on secrets was incorrectly still using the spec value instead of the one set in the status.

Would cause a panic if you had a secret labelled as default and a record with no provider ref.


maksymvavilov (approved):
Missed this one. Thanks!

PR #501, opened 2025-07-24
override provider falgs with configMap

maksymvavilov:
Controller will look up a config map on startup and will attempt to override controller flags with values from it


mikenairn (changes requested):
The change requested was to have the env vars loaded from a configmap mounted in the deployment, and **not** have it loaded directly in code.

```
envFrom:
    - configMapRef:
        name: app-config
```

K8s already has ways to do this that will allow restarts of the controller on update (we talked about this in the refinement call)

mikenairn:
I'd also expect env vars to be uppercase with underscores as per normal convention. i.e. `--metrics-bind-address`  =  `METRICS_BIND_ADDRESS`.

Is there no generic parser you could use instead of listing them all in a switch statement?

maksymvavilov:
Deprioritized

PR #504, opened 2025-07-25
[FEATURE] add endpoints provider

philbrookes:
Sample files to test this provider:
https://gist.github.com/philbrookes/b428d78a86d92add58e6e2f8bdf1c9a4

The provider secret (and related GEO codes and hostnames) in the envvars can be swapped out (in between deleting and recreating the records) to test multiple providers.

Note on cleanup, it is not possible to delete the destination record and the source records at the same time, all source records must be successfully deleted before the destination record is delete.


api/v1alpha1/dnsrecord_types_test.go
```diff
@@ -15,15 +15,10 @@ func TestValidate(t *testing.T) {
 		dnsNames []string
 		wantErr  bool
 	}{
-		{
```
philbrookes, line 18: this was never really testing for invalid domain, the error was because there were no endpoints that matched the roothost.

internal/provider/endpoint/endpoint.go
```diff
@@ -0,0 +1,216 @@
...
+	return dnsOperatorProvider.DNSProviderEndpoint
+}
+
+// AdjustEndpoints nothing to do here
+func (e *EndpointProvider) AdjustEndpoints(endpoints []*endpoint.Endpoint) ([]*endpoint.Endpoint, error) {
+	return endpoints, nil
+}
+
+func NewProviderFromSecret(ctx context.Context, client dynamic.Interface, s *v1.Secret, providerConfig dnsOperatorProvider.Config) (dnsOperatorProvider.Provider, error) {
+	logger := log.FromContext(ctx).WithName("endpoint-dns")
```
mikenairn, line 175: Should this add checks to make sure the secret exists and has a namespace? This will be important for this provider:

```go
if s == nil {
    return nil, fmt.Errorf("secret cannot be nil")
}

if s.GetNamespace() == "" {
    return nil, fmt.Errorf("namespace not set")
}
```

config/local-setup/dns-provider/endpoint/kustomization.yaml
```diff
@@ -0,0 +1,13 @@
...
+  disableNameSuffixHash: true
+  labels:
+    app.kubernetes.io/part-of: dns-operator
+    app.kubernetes.io/managed-by: kustomize
+    kuadrant.io/endpoint_gvr: "kuadrant.io_v1alpha1.dnsrecords"
+    kuadrant.io/zone_record_label: "kuadrant_zone_record"
+
+secretGenerator:
+  - name: dns-provider-credentials
+    type: "kuadrant.io/endpoint"
```
mikenairn, line 13:
```suggestion
generatorOptions:
  disableNameSuffixHash: true
  labels:
    app.kubernetes.io/part-of: dns-operator
    app.kubernetes.io/managed-by: kustomize

secretGenerator:
  - name: dns-provider-credentials
    type: "kuadrant.io/endpoint"
    literals:
      - ENDPOINT_GVR=kuadrant.io_v1alpha1.dnsrecords
      - ENDPOINT_ZONE_RECORD_LABEL=kuadrant.io/endpoint-provider-zone-record=true
```

As per the design I'd expect these to be setting configuration in the secret

api/v1alpha1/shared_types.go
```diff
@@ -62,7 +62,11 @@ const (
 
 	SecretTypeKuadrantCoreDNS corev1.SecretType = "kuadrant.io/coredns"
 
-	DefaultProviderSecretLabel = "kuadrant.io/default-provider"
+	DefaultProviderSecretLabel                   = "kuadrant.io/default-provider"
+	EndpointGVRKey                               = "kuadrant.io/endpoint_gvr"
+	EndpointLabelSelectorKey                     = "kuadrant.io/zone_record_label"
+	EndpointLabelValueKey                        = "kuadrant.io/zone_record_label_value"
```
mikenairn, line 68:
```suggestion
	EndpointGVRKey           = "ENDPOINT_GVR"
	EndpointLabelSelectorKey = "ENDPOINT_ZONE_RECORD_LABEL"
```

Consistent with how all other providers work, these are not expected to be labels, but keys in the provider secret.

internal/provider/endpoint/endpoint.go
```diff
@@ -0,0 +1,216 @@
...
+
+	endpointProvider := &EndpointProvider{
+		logger:        logger,
+		config:        providerConfig,
+		client:        client,
+		gvr:           gvr,
+		labelSelector: labels.Set(metav1.LabelSelector{MatchLabels: map[string]string{labelSelector: labelValue}}.MatchLabels).String(),
+		providerNS:    providerNS,
+		object:        s,
+	}
```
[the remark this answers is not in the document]
philbrookes: good idea!

cmd/main.go
```diff
@@ -145,8 +149,29 @@ func main() {
...
+
+	dynamicClient, err := dynamic.NewForConfig(config)
+	if err != nil {
+		setupLog.Error(err, "unable to create dynamic client for cluster")
+		os.Exit(1)
+	}
+
 	setupLog.Info("init provider factory", "providers", providers)
-	providerFactory, err := provider.NewFactory(mgr.GetClient(), providers)
+	providerFactory, err := provider.NewFactory(mgr.GetClient(), dynamicClient, providers)
```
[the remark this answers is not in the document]
philbrookes: I did initially want to do it this way, but couldn't figure out how to get the rest config.

Where would the provider get the rest config from?

api/v1alpha1/shared_types.go
```diff
@@ -63,6 +63,25 @@ const (
...
+	EndpointGVRKey = "ENDPOINT_GVR"
+
+	// The GVR to use in the endpoint provider is none are provided
+	DefaultEndpointGVR = "kuadrant.io/v1alpha1.dnsrecords"
+
+	// This is a label selector for the endpoint provider to find available zones
+	EndpointLabelSelectorKey = "ENDPOINT_ZONE_RECORD_LABEL"
+
+	// Default label selector used by the endpoint provider
+	DefaultLabelSelector = "kuadrant.io/kuadrant_zone_record=true"
```
mikenairn, line 77: While I'm not sure I've seen any hard rules about label names, I've always followed the convention of using hyphens over underscores in them since that seemed to be what [kubernetes](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/) themselves did. Either way, we should probably decide what we want to do and stick to a convention, I'd vote for hyphens for the reason above, it appears to be more common.

Also, i don't think this needs "kuadrant in it twice", would this not do:
```
kuadrant.io/zone-record=true
```

When we  are done we also need to make sure we go back to the rfc and diagrams and update to reflect reality.

philbrookes:
passing e2e tests: https://github.com/Kuadrant/dns-operator/actions/runs/16627047800

api/v1alpha1/dnsrecord_types_test.go
```diff
@@ -58,17 +53,6 @@ func TestValidate(t *testing.T) {
...
-			name:     "valid wildcard domain no endpoint",
-			rootHost: "*.example.com",
-			dnsNames: []string{
-				"a.b.example.com",
-				"b.a.example.com",
-				"a.example.com",
-				"b.example.com",
-			},
-			wantErr: true,
-		},
```
mikenairn, line 71: This should error, it would expect an endpoint to have a "*.example.com" dnsName. I think you removed the check and added it back so this can probably be reverted?

philbrookes: I had to keep that check out [here](https://github.com/Kuadrant/dns-operator/pull/504/files#diff-6a70ad61b66d79fb40a3db3e374825803cbd470e8f37a79153c054aace8b5527L249) as the zone record might be e.g. `rootHost: pb.hcpapps.net`, but the records merged into it: `app.pb.hcpapps.net`, so this error would trigger then and block the reconcile.

philbrookes: as it stands, the endpoint provider uses the existing logic to find a matching zone for the record (i.e. looks for the closest matching zone name), so it's possible to match a DNS Record `app.pb.hcpapps.net` against a zone record of `pb.hcpapps.net`. Although it will also work if both are `app.pb.hcpapps.net`.

I suppose it depends on the plan for creating the authoritative record, if it'll always match, then I can change the "find the right zone" logic for the endpoints to always expect a direct match, and then reinstate the above check.

I'm happy to go either way, but I made these changes under the assumption that we wanted the existing zone finding logic used in AWS etc, to also apply to the endpoints provider.

mikenairn: Spoke offline and now see what the issue is, and why we want to remove this validation.

Thought it through and i don't think it should be an issue. The important thing with this is that all the endpoints in the slice are related, and are part of the same domain as the `rootHost`, which we are still validating for.

internal/common/group_version.go
```diff
@@ -0,0 +1,50 @@
...
+	resourceIndex := gvkRegex.SubexpIndex("resource")
+
+	matches := gvkRegex.FindStringSubmatch(gvkStr)
+
+	if len(matches) < 3 {
+		return schema.GroupVersionResource{}, errors.New("provided GVR is an invalid format")
+	}
+
+	return schema.GroupVersionResource{Group: matches[groupIndex], Version: matches[versionIndex], Resource: matches[resourceIndex]}, nil
+}
```
mikenairn, line 50: I'm surprised there isn't already some parser that can do this. Can be changed later if we figure out there is.

api/v1alpha1/dnsrecord_types_test.go
philbrookes: Discussed in person: as we want to allow source records to be a subdomain of the destination records  (e.g. source rootHost: app.pb.com and destination rootHost: pb.com), we will leave this validation step out.

PR #509, opened 2025-07-29
ci: Negate input values used to select providers

mikenairn:
Use skip<provider> with a default value of false instead of test<provider> with a default of true.

Inputs are only set when the job is triggered manually using the "workflow_dispatch" event. The default needs to be that it runs the step if this does not exist or exists and is false.


mikenairn:
Verified it works on "push" here https://github.com/Kuadrant/dns-operator/actions/runs/16595784583/job/46942149016

maksymvavilov (approved):

philbrookes (approved):

PR #514, opened 2025-07-30
fix: status update on delete events

mikenairn:
Fixes an issue where the status update on delete to indicate that the finalizer can be removed was potentially not triggering an immediate requeue of the record in order for it to remove the finalizer.

The current logic in `updateStatus` might prevent it from actually doing any update if it decides it shouldn't, which is not compatible with the needs of the status update for deletion events.


internal/controller/dnsrecord_controller.go
```diff
@@ -401,12 +401,17 @@ func (r *DNSRecordReconciler) updateStatus(ctx context.Context, previous, curren
...
+
+func (r *DNSRecordReconciler) updateStatusAndRequeue(ctx context.Context, previous, current *v1alpha1.DNSRecord, requeueTime time.Duration) (reconcile.Result, error) {
+	logger := log.FromContext(ctx)
 	// update the record after setting the status
 	if !equality.Semantic.DeepEqual(previous.Status, current.Status) {
 		logger.V(1).Info("Updating status of DNSRecord")
 		if updateError := r.Status().Update(ctx, current); updateError != nil {
 			if apierrors.IsConflict(updateError) {
-				return ctrl.Result{Requeue: true}, nil
+				return ctrl.Result{RequeueAfter: time.Second}, nil
```
mikenairn, line 414: `Requeue` is deprecated, so not really sure how you are supposed to signal an immediate re-queue other than just setting a really short time :man_shrugging:

philbrookes: is `time.Second * 0` allowed?

mikenairn: Docs say: 
```
// RequeueAfter if greater than 0, tells the Controller to requeue the reconcile key after the Duration.
// Implies that Requeue is true, there is no need to set Requeue to true at the same time as RequeueAfter.
RequeueAfter time.Duration
```

So .. does that mean you could just use this to "imply" Requeue is true, or does it need to be greater than 0:

```
ctrl.Result{RequeueAfter: 0}
```

mikenairn: I'm going to merge this as is, as i know it gets round the current issue that is causing  a few problems. I think we should look into how that is supposed to actually work though, if "0" works it would probably be better, but it's really difficult to know for sure what has triggered a reconcile, so it would probably involve digging through the controller-runtime code to determine what it actually does.

philbrookes: yeah, that reads like it has to be greater than 0 to infer that requeue is true.

PR #549, opened 2025-08-26
make delegate field immutable

philbrookes:
It turns out I could create a DNSRecord with `delegate: false` and change it to `delegate: true`, fixing that here.


mikenairn (commented):
Good catch, i do recall thinking about this but must have forgotten to come back to it. Can we add a test for this here https://github.com/Kuadrant/dns-operator/blob/main/internal/controller/dnsrecord_controller_test.go#L149-L182

philbrookes:
> Good catch, i do recall thinking about this but must have forgotten to come back to it. Can we add a test for this here https://github.com/Kuadrant/dns-operator/blob/main/internal/controller/dnsrecord_controller_test.go#L149-L182

it appears not. I can create a dnsrecord with spec.delegate false from yaml, but if I try to do it by the k8sclient, the omitempty removes it. So setting it to false and then trying to set it back to true gives me: `Delegate can't be set if it was previously unset`. Do you know  a way to force the writing of the value as false with omitempty?

I also cannot test it the other way around (i.e. false to true) as that's the same as [this test](https://github.com/Kuadrant/dns-operator/blob/main/internal/controller/dnsrecord_controller_test.go#L169):

philbrookes:
> > Good catch, i do recall thinking about this but must have forgotten to come back to it. Can we add a test for this here https://github.com/Kuadrant/dns-operator/blob/main/internal/controller/dnsrecord_controller_test.go#L149-L182
> 
> it appears not. I can create a dnsrecord with spec.delegate false from yaml, but if I try to do it by the k8sclient, the omitempty removes it. So setting it to false and then trying to set it back to true gives me: `Delegate can't be set if it was previously unset`. Do you know a way to force the writing of the value as false with omitempty?
> 
> I also cannot test it the other way around (i.e. false to true) as that's the same as [this test](https://github.com/Kuadrant/dns-operator/blob/main/internal/controller/dnsrecord_controller_test.go#L169):

According to this I need to change the bool to a pointer, which seems a little excessive: https://stackoverflow.com/questions/72206365/go-omitempty-bool-field-not-showing-false-type

PR #550, opened 2025-08-27
ADD: Metric

Boomatang:
Added metric is used to give a SHA of the authoritative records on the
cluster.

Signed-off-by: Jim Fitzpatrick <jfitzpat@redhat.com>


Boomatang:
From the conversation in [Zulip]([#DNS > Metrics to determine identical authoritative records](https://kuadrant.zulipchat.com/#narrow/channel/493518-DNS/topic/Metrics.20to.20determine.20identical.20authoritative.20records)) this needs a rework. Putting the PR back into draft to do that rework.

internal/metrics/metrics.go
```diff
@@ -46,9 +58,48 @@ var (
...
+	NameSpace string
+}
+
+func (m *authoritativeRecordSpecInfoMetric) calculateSha(dnsRecord *v1alpha1.DNSRecord) error {
+	if dnsRecord == nil {
+		return errors.New("dnsRecord nil pointer")
+	}
+
+	specString := hash.GetCanonicalString(dnsRecord.Spec)
+	m.Sha = hash.ToBase36HashLen(specString, 6)
```
mikenairn, line 91: In terms of metrics does the length of the value have a significant impact on the expense of this metric? I'm wondering would you be better to use `hash.ToBase36Hash`, or some other hashing algorithm,  to decrease the possibility of two different strings returning the same value.

internal/common/hash/hash.go
```diff
@@ -16,3 +20,91 @@ func ToBase36Hash(s string) string {
...
+
+	case reflect.Invalid, reflect.Chan, reflect.Func, reflect.UnsafePointer, reflect.Uintptr:
+		// Handle nil or unsupported types gracefully.
+		return "null"
+
+	default:
+		// Fallback for any other type.
+		return fmt.Sprintf("%v", v)
+	}
+}
```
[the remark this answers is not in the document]
Boomatang: I didn't spend a long time looking, but what I found were mostly focused on testing, with some saying not to use it in production. Or it was examples using the standard library features. The feature that seem to be missing was the sorting of the elements before generation the string.

Thinking about it now, you could marshal it into a json, and sort the characters in the json string. The string would not be human readable, but also it doesn't have to be. We only need a string with the same bit order for the hash. 

The more I think about this, the more I want to try the sorting of the string method.

internal/metrics/metrics.go
mikenairn: OK, i don't really mind what we use. The shortened version of the UID was being used because of limitations in the txt registry. This limitation no longer exist, and we probably should be using the full length hash for those instead.

mikenairn (approved):

PR #551, opened 2025-08-27
Update integration test suite for primary and secondary testing 

mikenairn:
Updates the integration test suite to make it possible to test both primary and secondary functionality in isolation or together in more comprehensive test cases.

The test suite creates two controller-runtime test environments (k8s api server, etcd etc..) on initialisation, one to act as a primary and one as a secondary. Each test environment runs the appropriate components for the given delegation role(based on the main application):

**Primary:**
   - create multicluster-controller-runtime manager
   - setup kubeconfig provider
   - setup DNSRecordReconciler (DelegationRole: primary)
   - setup RemoteDNSRecordReconciler

**Secondary:**
   - create controller-runtime manager
   - setup DNSRecordReconciler  (DelegationRole: secondary)

The test suite contains a global reference to the primary and secondary environments, managers, k8s clients and kubeconfigs for a dedicated user "kuadrant" allowing them to be used in specs as required.

**Add kubeconfig provider tests**

Adds basic tests to the integration test suite to validate the behaviour of the kubeconfig provider for multicluster secrets.

Makes use of `gbytes.Buffer` to capture log statements during the current test and the gomega `gbytes.Say` matcher see
https://onsi.github.io/gomega/#codegbytescode-testing-streaming-buffers since there is no way of really checking the status success/failure of these secrets from any particular resource. Logs in the test suite are prepended with the delegation role (primary or secondary) so it can be determined which environment is logging the statements.

### DIFF
--- a/internal/controller/kubeconfig_provider_test.go
+++ b/internal/controller/kubeconfig_provider_test.go
@@ -1,0 +1,129 @@
+//go:build integration
+
+package controller
+
+import (
+	"fmt"
+	"strings"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/gbytes"
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/kuadrant/dns-operator/api/v1alpha1"
+)
+
+var _ = Describe("Kubeconfig Provider", Labels{"multicluster"}, func() {
+
+	var (
+		secondaryClusterSecret *v1.Secret
+
+		// buffer containing all the log entries added during the current spec execution
+		// can be used in conjunction with `gbytes.Say` matcher to check log entries see https://onsi.github.io/gomega/#codegbytescode-testing-streaming-buffers
+		logBuffer *gbytes.Buffer
+	)
+
+	BeforeEach(func() {
+		logBuffer = gbytes.NewBuffer()
+		GinkgoWriter.TeeTo(logBuffer)
+
+		secondaryClusterSecret = &v1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "secondary-cluster-1",
+				Namespace: testDefaultClusterSecretNamespace,
+				Labels: map[string]string{
+					testDefaultClusterSecretLabel: "true",
+				},
+			},
+			StringData: map[string]string{
+				"kubeconfig": string(secondaryKubeconfig),
+			},
+		}
+	})
+
+	AfterEach(func() {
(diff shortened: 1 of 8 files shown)


internal/controller/kubeconfig_provider_test.go
```diff
@@ -0,0 +1,129 @@
...
+
+		testZoneDomainName := strings.Join([]string{GenerateName(), "example.com"}, ".")
+		testHostname := strings.Join([]string{"foo", testZoneDomainName}, ".")
+
+		secondaryRecord := &v1alpha1.DNSRecord{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      testHostname,
+				Namespace: testNamespace,
+			},
+			Spec: v1alpha1.DNSRecordSpec{
```
philbrookes, line 96: Should this have "Delegate: true"? I guess it still emits the log either way, so that's why this is passing?

internal/controller/kubeconfig_provider_test.go
```diff
@@ -0,0 +1,129 @@
...
+	})
+
+	AfterEach(func() {
+		if secondaryClusterSecret != nil {
+			Expect(client.IgnoreNotFound(primaryK8sClient.Delete(ctx, secondaryClusterSecret))).To(Succeed())
+		}
+		GinkgoWriter.ClearTeeWriters()
+	})
+
+	It("successfully creates a new cluster from valid kubeconfig secret", Labels{"primary"}, func(ctx SpecContext) {
```
philbrookes, line 56: "creates a new cluster"? Is this better phrased as "creates a new controller"

internal/controller/kubeconfig_provider_test.go
```diff
@@ -0,0 +1,129 @@
...
+		createClusterKubeconfigSecret(primaryK8sClient, secondaryClusterSecret, logBuffer)
+	})
+
+	It("successfully deletes cluster", Labels{"primary"}, func(ctx SpecContext) {
```
philbrookes, line 60: same here and line 65 too.

philbrookes (approved):
I think this looks fine. I just had a few questions that do not need to block merging this.

internal/controller/kubeconfig_provider_test.go
mikenairn: > I guess it still emits the log either way, so that's why this is passing?

yes

internal/controller/kubeconfig_provider_test.go
mikenairn: changing it to "successfully establishes a connection to a remote cluster from a valid kubeconfig secret"

internal/controller/kubeconfig_provider_test.go
mikenairn: changing it to "successfully removes connection to remote cluster"

philbrookes (approved):

Issue #342, opened 2025-01-15
DNS Record is not ready for more than N duration


philbrookes:

maksymvavilov:
The #537 introduces a ready metric that can be used in a query

PR #552, opened 2025-08-27
emit dns record metadata as metric

maksymvavilov:
Adding a new metric to report usefull metadata from DNSRecord CR. 
There will be a new metric per DNS Record and the metrics will stay there. Meaning if you create and delete 100 records - you will have 100 "dead" metrics (until they are [removed](https://prometheus.io/docs/prometheus/latest/storage/#right-sizing-retention-size)) 

Here is how it looks:
image

There are two ways to skin this cat: 
1. have the ready metric carry all the metadata <- current approach 
2. have a dedicated `_info` metric purely for the metadata and later on create a separate ready metric for #342

### DIFF
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -9,6 +9,8 @@ import (
 const (
 	dnsRecordNameLabel           = "dns_record_name"
 	dnsRecordNamespaceLabel      = "dns_record_namespace"
+	dnsRecordRootHost            = "dns_record_root_host"
+	dnsRecordDelegating          = "dns_record_is_delegating"
 	dnsHealthCheckNameLabel      = "dns_health_check_name"
 	dnsHealthCheckNamespaceLabel = "dns_health_check_namespace"
 	dnsHealthCheckHostLabel      = "dns_health_check_host"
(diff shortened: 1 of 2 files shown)


maksymvavilov:
There are two ways to skin this cat: 
1. have the ready metric carry all the metadata 
2. have a dedicated `_info` metric purely for the metadata and later on create a separate ready metric for #342

internal/metrics/metrics.go
```diff
@@ -36,10 +38,17 @@ var (
 			Help: "Emits one when provider secret is found to be absent, or zero when expected secrets exist",
 		},
 		[]string{mzRecordNameLabel, mzRecordNamespaceLabel, mzSecretNameLabel})
+	RecordReady = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "dns_provider_record_ready",
```
Boomatang, line 43: There was a question over using "_ready" or "_info". I think ready is fine, as it is what the value of the metric is indicating. From my understanding of the last few days, if the metric has the "_info" suffix the value of the metric would not be expected to change, and should always be set to 1.

Boomatang (approved):
I miss understood that the reconcile  is also on a timer. My concern over upgrades is not valid. Happy to approve this.

I do have one general question not related to this PR over the use of `var` within the project.

internal/metrics/metrics.go
maksymvavilov: This is the way I understood Dave as well. 
It just seems... wasteful to me to disregard a metric value  - this is a piece of information that could be used. 
That said, if we did not have a concept of record being ready, I'm not sure I could justify anything else as a "value" of the metric :thinking:

PR #520, opened 2025-08-01
feat: multicluster dnsrecord delegation

mikenairn:
closes #518

Updates the dns record reconciler to accommodate the required flow when using delegation in a multicluster (primary/secondary) configuration.

Related changes:
 - #519
 - #517 
- #489 

**Verification**

**Terminal 1 (primary)**
```
make multicluster-local-setup DEPLOY=false
kubectl config use-context kind-kuadrant-dns-local-1
make run
```

**Terminal 2 (secondary)**
```
export KUBECONFIG=tmp/kubeconfigs/kuadrant-local-all.kubeconfig
kubectl config use-context kind-kuadrant-dns-local-2
make run-secondary
```

**Terminal 3**

Verify cluster secrets
```
kubectl get secrets -A -l kuadrant.io/multicluster-kubeconfig=true --show-labels --context kind-kuadrant-dns-local-1
```
Expected output:
```
NAMESPACE             NAME                        TYPE     DATA   AGE     LABELS
dns-operator-system   kind-kuadrant-dns-local-2   Opaque   1      2m31s   kuadrant.io/multicluster-kubeconfig=true
kubectl get secrets -A -l kuadrant.io/multicluster-kubeconfig=true --show-labels --context kind-kuadrant-dns-local-2
No resources found
```

Label default provider
```
kubectl label secret/dns-provider-credentials-inmemory -n dnstest kuadrant.io/default-provider=true --context kind-kuadrant-dns-local-1
```

Apply record 1 on cluster 1
```
kubectl apply -f config/local-setup/dnsrecords/delegating/inmem/simple/dnsrecord-simple-inmem-cluster1.yaml -n dnstest --context kind-kuadrant-dns-local-1
```

Verify record
```
kubectl get dnsrecord -o wide -n dnstest --show-labels --context kind-kuadrant-dns-local-1
```
Expected output:
```
NAME                                    READY   HEALTHY   ROOT HOST               OWNER ID   ZONE DOMAIN             ZONE ID                                 LABELS
delegation-authoritative-record-c2ckr   True              simple.kuadrant.local   tp3vumh9   kuadrant.local          kuadrant.local                          kuadrant.io/delegation-authoritative-record=2km2e6b6,kuadrant.io/dns-provider-name=inmemory
dnsrecord-simple-inmem-cluster1         True              simple.kuadrant.local   2a9v835j   simple.kuadrant.local   delegation-authoritative-record-c2ckr   kuadrant.io/dns-provider-name=endpoint
```

[...]

PR #557, opened 2025-08-29
update delegation authoritative record naming

mikenairn:
Updates the naming of the authoritative dnsrecord created by the delegation feature to base the name on a hash of the root host rather than generating a random name. Naming the auth records in this way allows them to be deterministic, and in environments with more than one primary allows the names to match across clusters conforming to the sameness principals of k8s multi cluster that we want to achieve.

Updates the labels added to "authoritative" records during creation: 
```
kuadrant.io/authoritative-record=true
kuadrant.io/authoritative-record-hash=<hash of record.spec.rootHost>
```

Allows authoritative records to be more easily listed via the cli:
```
kubectl get dnsrecord -A -l kuadrant.io/authoritative-record=true
```

follow up to https://github.com/Kuadrant/dns-operator/pull/520

### DIFF
--- a/internal/controller/dnsrecord_delegation_helper.go
+++ b/internal/controller/dnsrecord_delegation_helper.go
@@ -33,7 +33,7 @@ func (r *DNSRecordDelegationHelper) EnsureAuthoritativeRecord(ctx context.Contex
 func (r *DNSRecordDelegationHelper) getAuthoritativeRecordFor(ctx context.Context, record v1alpha1.DNSRecord) (*v1alpha1.DNSRecord, error) {
 	aRecords := v1alpha1.DNSRecordList{}
 
-	labelSelector, err := labels.Parse(fmt.Sprintf("%s=%s", v1alpha1.DelegationAuthoritativeRecordLabel, common.HashRootHost(record.Spec.RootHost)))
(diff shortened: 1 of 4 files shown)


internal/controller/dnsrecord_delegation_helper.go
```diff
@@ -60,16 +60,21 @@ func (r *DNSRecordDelegationHelper) createAuthoritativeRecordFor(ctx context.Con
...
 func authoritativeRecordFor(rec v1alpha1.DNSRecord) *v1alpha1.DNSRecord {
+	rootHostHash := common.HashRootHost(rec.Spec.RootHost)
 	return &v1alpha1.DNSRecord{
 		ObjectMeta: metav1.ObjectMeta{
-			GenerateName: "delegation-authoritative-record-",
-			Namespace:    rec.Namespace,
+			Name:      toAuthoritativeRecordName(rootHostHash),
+			Namespace: rec.Namespace,
 			Labels: map[string]string{
-				v1alpha1.DelegationAuthoritativeRecordLabel: common.HashRootHost(rec.Spec.RootHost),
```
maleck13, line 68: do we want to label these records and authoratative: true so listing authoratitive records only is possible?

maleck13 (approved):

PR #561, opened 2025-09-05
refactor: delegation and authoritative record creation logic

mikenairn:
Moves the delegation code into the provider package. The provider itself is now adapted for delegating instances to ensure the existence of the required authoritative record when any requests to `DNSZoneForHost` are made. Also adds the ability for individual providers to require delegation implicitly if they required it, this will be the case for the CoreDNS provider in upcoming changes.

Requires:
- https://github.com/Kuadrant/dns-operator/pull/560


internal/provider/endpoint/authoritative_dnsrecord_test.go
```diff
@@ -0,0 +1,107 @@
...
+			},
+		}
+
+		actualProvider, err := NewAuthoritativeDNSRecordProvider(t.Context(), c, dc, pa, provider.Config{})
+		assert.NotNil(t, actualProvider)
+		assert.NoError(t, err)
+	})
+}
+
+func TestAuthoritativeDNSRecordProvider_DNSZoneForHost(t *testing.T) {
```
maleck13, line 52: did you want to add these now or post merge?

internal/provider/factory.go
```diff
@@ -68,76 +72,91 @@ type Factory interface {
...
-
-		return nil, fmt.Errorf("unrecognised contructor for provider '%s'", provider)
+		return nil, fmt.Errorf("unrecognised contructor for provider '%s'", providerName)
 	}
+	return nil, fmt.Errorf("providerName '%s' not registered", providerName)
+}
 
-	return nil, fmt.Errorf("provider '%s' not registered", provider)
+// requiresDelegation return true if the given provider requires delegation be enforced for the given resource(DNSRecord)
+func requiresDelegation(_ Provider, _ v1alpha1.ProviderAccessor) bool {
```
maleck13, line 158: is this how you expect as in always returning false?

internal/provider/factory.go
```diff
@@ -68,76 +72,91 @@ type Factory interface {
...
 				Namespace: pa.GetNamespace(),
 			}}
+		if err = f.Client.Get(ctx, client.ObjectKeyFromObject(pSecret), pSecret); err != nil {
+			return
+		}
 
-		if err := f.Client.Get(ctx, client.ObjectKeyFromObject(providerSecret), providerSecret); err != nil {
-			return nil, err
+		provider, err = f.providerForSecret(ctx, pSecret, c)
+		if err != nil {
```
maleck13, line 111: Slightly diff here with named return values. Took me a while to understand that was happening

internal/provider/factory_test.go
```diff
@@ -0,0 +1,239 @@
...
+
+		p, err := f.ProviderFor(t.Context(), pa, Config{})
+		assert.NotNil(t, p)
+		assert.NoError(t, err)
+		assert.Equal(t, string(p.Name()), "fake-delegation-provider")
+	})
+
+	t.Run("returns configured delegation provider if provider requires delegation", func(t *testing.T) {
+		t.Run("when accessor requests delegation ", func(t *testing.T) {
+			//Todo
```
maleck13, line 135: do we want to leave these here?

maleck13 (commented):
Couple of comments but generally looks ok to me

internal/provider/factory.go
mikenairn: When CoreDNS is updated this be changed, for now it just returns false.

internal/provider/endpoint/authoritative_dnsrecord.go
```diff
@@ -0,0 +1,139 @@
...
+	labelSelector, err := labels.Parse(authoritativeRecordSelectorFor(record))
+	if err != nil {
+		return nil, err
+	}
+
+	if err = p.Client.List(ctx, &aRecords, &client.ListOptions{LabelSelector: labelSelector}); err != nil {
+		return nil, fmt.Errorf("failed to get authoritative record: %w", err)
+	}
+
+	if len(aRecords.Items) > 0 {
```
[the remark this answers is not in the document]
mikenairn: Possibly, i haven't updated it though since this PR is about moving code without modification to the logic.

Previously we would match on a more generic label, but now we expect a label that matches exactly based on the root host. Since we are also now creating a record with a name based on the root host I'm thinking this might be better changed to just look for the expected resource by name in the current namespace. 

I would prefer to make this change in a different PR though.

internal/provider/endpoint/authoritative_dnsrecord.go
```diff
@@ -0,0 +1,139 @@
...
+	"github.com/kuadrant/dns-operator/internal/provider"
+)
+
+var (
+	errIncompatibleAccessorType = fmt.Errorf("authoritative DNSRecords can only be created for resources of the same type (DNSRecord)")
+)
+
+var _ provider.Provider = &AuthoritativeDNSRecordProvider{}
+
+type AuthoritativeDNSRecordProvider struct {
```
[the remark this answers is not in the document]
mikenairn: Added some docs, probably could do with more, but it's a start.

PR #571, opened 2025-09-12
feat: Add support for multiple primary controllers

### LINKED ISSUE #562: Add support for multiple primary clusters
**What**

Add support for multiple primary clusters.

**How**

Add a new field to the DNSRecord status `remoteRecordStatuses` that is used to store multiple DNSRecordStatuses keyed by a UID relevant only to the controller that added it. During remote record reconciliation this status is loaded and updated by each primary.

```yaml
{
  "conditions": [
    {
      "lastTransitionTime": "2025-09-10T09:23:02Z",
      "message": "",
      "reason": "FinalizersSet",
      "status": "True",
      "type": "ReadyForDelegation"
    }
  ],
  "ownerID": "2w5ajt0s",
  "providerRef": {
    "name": ""
  },
  "remoteRecordStatuses": {
    "111-222-333-444-555-666": {
      "conditions": [
        {
          "lastTransitionTime": "2025-09-10T09:23:03Z",
          "message": "Provider ensured the dns record",
          "observedGeneration": 1,
          "reason": "ProviderSuccess",
          "status": "True",
          "type": "Ready"
        }
      ],
      "domainOwners": [
        "2svwsfo4",
        "2w5ajt0s"
      ],
      "endpoints": [
        {
          "dnsName": "simple.k.example.com",
          "labels": {
            "owner": "2svwsfo4&&2w5ajt0s"
          },
          "recordTTL": 60,
          "recordType": "A",
          "targets": [
            "172.18.200.2"
          ]
        }
      ],
      "providerRef": {
        "name": ""
      },
      "queuedAt": null,
      "zoneDomainName": "simple.k.example.com",
      "zoneID": "authoritative-record-1c8db9d0"
    }
  }
}
```

* Use a UID for the zone.

Requires:
 - https://github.com/Kuadrant/dns-operator/issues/563

mikenairn:
closes #562 

Adds a new field to the status of the DNSRecord resource `remoteRecordStatuses` which is intended to hold primary cluster specific statuses, one for each primary, that is reconciling it as a "remote" and adding it's endpoints to a local authoritative record. A cluster UID value is generated from the kube-system namespace for a repeatable cluster specific ID, and used as the key in the `remoteRecordStatuses` map, with each entry being a `DNSRecordStatus` for that cluster.

Note: It was mentioned we would use a "Zone ID" or something here instead of a "Cluster ID" but that wasn't really possible since this id needed to be different for every cluster, and is really a status for a cluster rather than a record on it. 

The RemoteDNSRecordReconciler is updated to reconcile the remote record by loading and updating the correct cluster specific status in `record.Status.remoteRecordStatuses` instead of the normal `record.Status`. A `RemoteDNSRecord` wrapper and `DNSRecordAccessor` interface were added to help achieve the required loading and setting of the status, with the intention of making this a pattern used in other reconcilers at a later date to reduce the amount of duplicated code.

Most changes are confined to the remote record reconciliation logic, but some changes were made to the checks in `ProviderEndpointsRemoved` to ensure during deletion events that all remote clusters have the chance to
clean up before the dnsrecord finalizer is removed.

### Verification

Create clusters (2 primary)
```
make multicluster-local-setup CLUSTER_COUNT=2 PRIMARY_CLUSTER_COUNT=2 DEPLOY=false
```

**Terminal 1**
```
kubectl config use-context kind-kuadrant-dns-local-1 && make run RUN_METRICS_ADDR=:8080 RUN_HEALTH_ADDR=:8081 RUN_DELEGATION_ROLE=primary
```

**Terminal 2**
```
kubectl config use-context kind-kuadrant-dns-local-2 && make run RUN_METRICS_ADDR=:8082 RUN_HEALTH_ADDR=:8083 RUN_DELEGATION_ROLE=primary
```

**Terminal 3**

Label coredns as the default provider on both clusters:
```
kubectl label secret/dns-provider-credentials-coredns -n dnstest kuadrant.io/default-provider=true --context kind-kuadrant-dns-local-1
kubectl label secret/dns-provider-credentials-coredns -n dnstest kuadrant.io/default-provider=true --context kind-kuadrant-dns-local-2
```

Create delegating records on both clusters:
```
kubectl apply -f config/local-setup/dnsrecords/delegating/coredns/simple/dnsrecord-simple-coredns-cluster1.yaml --context kind-kuadrant-dns-local-1 -n dnstest
kubectl apply -f config/local-setup/dnsrecords/delegating/coredns/simple/dnsrecord-simple-coredns-cluster2.yaml --context kind-kuadrant-dns-local-2 -n dnstest
```

Verify the dnsrecord are created as expected
```
kubectl get dnsrecord -n dnstest -o wide --show-labels --context kind-kuadrant-dns-local-1
```
Expected output
```
NAME                                READY   HEALTHY   ROOT HOST              OWNER ID   ZONE DOMAIN            ZONE ID                         LABELS
authoritative-record-1c8db9d0       True              simple.k.example.com   2qaqohjj   k.example.com          k.example.com                   kuadrant.io/authoritative-record-hash=1c8db9d0,kuadrant.io/authoritative-record=true,kuadrant.io/coredns-zone-name=k.example.com,kuadrant.io/dns-provider-name=coredns
dnsrecord-simple-coredns-cluster1   True              simple.k.example.com   217kxadk   simple.k.example.com   authoritative-record-1c8db9d0   <none>
```

```
kubectl get dnsrecord -n dnstest -o wide --show-labels --context kind-kuadrant-dns-local-2
```
Expected output
```
NAME                                READY   HEALTHY   ROOT HOST              OWNER ID   ZONE DOMAIN            ZONE ID                         LABELS
authoritative-record-1c8db9d0       True              simple.k.example.com   283zngr1   k.example.com          k.example.com                   kuadrant.io/authoritative-record-hash=1c8db9d0,kuadrant.io/authoritative-record=true,kuadrant.io/coredns-zone-name=k.example.com,kuadrant.io/dns-provider-name=coredns
dnsrecord-simple-coredns-cluster2   True              simple.k.example.com   yfy2hheg   simple.k.example.com   authoritative-record-1c8db9d0   <none>
```

### DIFF
--- a/api/v1alpha1/dnsrecord_types.go
+++ b/api/v1alpha1/dnsrecord_types.go
@@ -17,9 +17,11 @@ limitations under the License.
 package v1alpha1
 
 import (
+	"encoding/json"
 	"fmt"
 	"strings"
 
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	externaldns "sigs.k8s.io/external-dns/endpoint"
@@ -170,6 +172,47 @@ type DNSRecordStatus struct {
 
 	// zoneDomainName is the domain name of the zone that the dns record is publishing endpoints
 	ZoneDomainName string `json:"zoneDomainName,omitempty"`
+
+	// remoteRecordStatuses is a map of cluster IDs and their unique DNSRecordStatus as raw JSON.
+	//
+	// A CRD can't reference a type within itself so the `apiextensionsv1.JSON` type is used.
+	// Use GetRemoteRecordStatuses to get the converted type.
+	RemoteRecordStatuses map[string]apiextensionsv1.JSON `json:"remoteRecordStatuses,omitempty"`
+}
+
+// GetRemoteRecordStatuses returns any remote record statuses in the current status.
+//
+// Converts the raw json type `map[string]apiextensionsv1.JSON` to the desired map[string]DNSRecordStatus` type.
+func (s *DNSRecordStatus) GetRemoteRecordStatuses() map[string]DNSRecordStatus {
+	out := map[string]DNSRecordStatus{}
+
+	for i, v := range s.RemoteRecordStatuses {
+		drs := &DNSRecordStatus{}
+		if err := json.Unmarshal(v.Raw, drs); err != nil {
+			return nil
+		}
+		out[i] = *drs
+	}
+
+	return out
+}
+
+// GetRemoteRecordStatus returns a remote record status for the given cluster ID.
+func (s *DNSRecordStatus) GetRemoteRecordStatus(clusterID string) DNSRecordStatus {
+	status, ok := s.GetRemoteRecordStatuses()[clusterID]
+	if !ok {
+		status = DNSRecordStatus{}
+	}
+	return status
+}
+
+// SetRemoteRecordStatus sets a remote record status for the given cluster ID.
+func (s *DNSRecordStatus) SetRemoteRecordStatus(clusterID string, rStatus DNSRecordStatus) {
+	if s.RemoteRecordStatuses == nil {
+		s.RemoteRecordStatuses = map[string]apiextensionsv1.JSON{}
+	}
+	js, _ := json.Marshal(rStatus)
+	s.RemoteRecordStatuses[clusterID] = apiextensionsv1.JSON{Raw: js}
 }
 
 func (s *DNSRecordStatus) ReadyForDelegation() bool {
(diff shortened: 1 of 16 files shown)


internal/controller/dnsrecord_accessor.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+func (s *RemoteDNSRecord) GetSpec() *v1alpha1.DNSRecordSpec {
+	return &s.Spec
+}
+
+// GetStatus returns the status set for the current cluster ID.
+// If none is set an empty DNSRecordStatus is returned.
+func (s *RemoteDNSRecord) GetStatus() *v1alpha1.DNSRecordStatus {
+	if s.status == nil {
```
maleck13, line 80: this reads a bit odd. So the cluster status is only set once ?

api/v1alpha1/dnsrecord_types.go
```diff
@@ -170,6 +172,47 @@ type DNSRecordStatus struct {
 
 	// zoneDomainName is the domain name of the zone that the dns record is publishing endpoints
 	ZoneDomainName string `json:"zoneDomainName,omitempty"`
+
+	// remoteRecordStatuses is a map of cluster IDs and their unique DNSRecordStatus as raw JSON.
```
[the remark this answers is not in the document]
mikenairn: These would be statuses from primary clusters yes. Did you think a rename is needed, more specific?

internal/controller/dnsrecord_accessor.go
mikenairn: It currently means that `setStatus()` must be called before doing an `Update` of the resource through the k8s API which I'm not totally happy with, and is the reason that all the `SetStatus*` methods are added to the accessor interface as they currently make the call to `setStatus()` on all updates to make sure it is actually set in the raw json.

maleck13 (approved):
verfied locally. with deletion, modifying 👍

PR #580, opened 2025-09-18
equate not set delegate field to false in CR validation

maksymvavilov:
The issue we have now is that `false` for a boolean is treated like a `default` value and is omitted. Se [here](https://github.com/golang/go/issues/13284#issuecomment-157288692).  
The typical resolution is to make a primitive into a pointer that gives you 3 values to work with. In our case, it is not optimal as the default value and the absence of a value are functionally the same. Right now, to solve this, we are forbidding setting `delegate` if it was unset and unsetting if it was set. 

I'm changing the CEL validation rules to only forbid an update from `nil` to `true` and from `true` to `nil`. The rest of the cases are covered by the immutability rule. 

Here is a breakdown.

Rule: `oldSelf == self` 
| old   | new   | got   | want  |
|-------|-------|-------|-------|
| false | nil   | true  | true  |
| true  | nil   | true  | false |
| nil   | true  | true  | false |
| false | true  | false | false |
| nil   | false | true  | true  |
| true  | false | false | false |

Rule: `has(oldSelf.delegate) || !has(self.delegate) || self.delegate == false`
| old   | new   | got   | want  | covered above |
|-------|-------|-------|-------|:-------------:|
| false | nil   | true  | true  |               |
| true  | nil   | true  | false |               |
| nil   | true  | false | false |               |
| false | true  | true  | false |       *       |
| nil   | false | true  | true  |               |
| true  | false | true  | false |       *       |

Rule: `!has(oldSelf.delegate) || oldSelf.delegate == false || has(self.delegate)`
| old   | new   | got   | want  | covered above |
|-------|-------|-------|-------|:-------------:|
| false | nil   | true  | true  |               |
| true  | nil   | false | false |               |
| nil   | true  | true  | false |       *       |
| false | true  | true  | false |       *       |
| nil   | false | true  | true  |               |
| true  | false | true  | false |       *       |

All combinations:
| old   | new   | all rules | want  |
|-------|-------|-----------|-------|
| nil   | nil   | n/a       | n/a   |
| false | nil   | true      | true  |
| true  | nil   | false     | false |
| nil   | true  | false     | false |
| false | true  | false     | false |
| true  | true  | n/a       | n/a   |
| nil   | false | true      | true  |
| false | false | n/a       | n/a   |
| true  | false | false     | false |

### DIFF
--- a/api/v1alpha1/dnsrecord_types.go
+++ b/api/v1alpha1/dnsrecord_types.go
@@ -90,8 +90,8 @@ type HealthCheckStatusProbe struct {
 // DNSRecordSpec defines the desired state of DNSRecord
 // +kubebuilder:validation:XValidation:rule="!has(oldSelf.ownerID) || has(self.ownerID)", message="OwnerID can't be unset if it was previously set"
 // +kubebuilder:validation:XValidation:rule="has(oldSelf.ownerID) || !has(self.ownerID)", message="OwnerID can't be set if it was previously unset"
-// +kubebuilder:validation:XValidation:rule="!has(oldSelf.delegate) || has(self.delegate)", message="Delegate can't be unset if it was previously set"
-// +kubebuilder:validation:XValidation:rule="has(oldSelf.delegate) || !has(self.delegate)", message="Delegate can't be set if it was previously unset"
+// +kubebuilder:validation:XValidation:rule="has(oldSelf.delegate) || !has(self.delegate) || self.delegate == false", message="delegate can't be set to true if unset"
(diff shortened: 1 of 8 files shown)


Boomatang (approved):
All seems to be working as expected. I played around with updating the resource to trigger the validation. I think I checked all the different version
- nil -> false (no providerRef)
- false -> nil (no providerRef)
- nil -> false (with providerRef)
- false -> nil (with providerRef)
- nil -> true (no providerRef)
- true -> nil (no providerRef)
- nil -> true (with providerRef)
- When true adding providerRef

mikenairn (approved):
Good to see the extra tests :+1: . I haven't tried it, but sounds like others have so think we should merge. Do we need to check the kuadrant operator validations for the same?

Issue #578, opened 2025-09-17
Metrics, on label change not removed from cache


Boomatang:
# What

The client_golang/prometheus never removes metrics from its cache. This causes the `dns_provider_authoritative_record_spec_info` to remain when the sha label is changed. 

# Solutions
## Approach 1
There is a [delete](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#MetricVec.Delete) function that should allow removing the older metric

## Approach 2
Configure a prometheus [collector](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#Collector) which would only emit the metric on scrapes.

PR #582, opened 2025-09-19
Introduce metric collector

Boomatang:
Closes: #578 

The metric being published by the operator was never removed from the cache in with the operator.
This meant with the `_info` metric the older version of the metric, where the sha was different, was still exported.
With in Prometheus this made it impossible to compare the latest sha across clusters.

The collector pattern also allowed the removal of the reconcile that was publishing the count of connected clusters.
There is no external change to how this metric behaves. 

## Verification

To verify these change, deploy a multi cluster setup.
Add a number of delegated records.
Check the metrics.
For each authoritative record there should only be one `dns_provider_authoritative_record_spec_info` entry, no matter how much change is happening.
Only the `sha` label should change.

For the refactor on the connected remote clusters.
This should not change the count, add should be equal to the number of clusters configured.
Metric is `dns_provider_active_multi_cluster_count`


Boomatang:
While this was very much focused on the `_info` bug and the refactor to remove the reconcile. The other existing metrics should be reviewed, and refactered to match this format. Refacting metrics that include any label based on user created resources could be a memory leak, #579 and doing this refacter I think solves that problem without having to investigate if there was any OOM problem to being with.

internal/metrics/metrics.go
```diff
@@ -108,6 +107,65 @@ func (m *recordReadyMetric) Publish() {
...
+
+	listOptions := &client.ListOptions{}
+	dnsRecordList := &v1alpha1.DNSRecordList{}
+
+	err := cl.List(c.Ctx, dnsRecordList, listOptions)
+	if err != nil {
+		logger.Info("failed to list dns records", "status", "error")
+	}
+
+	for _, record := range dnsRecordList.Items {
```
maksymvavilov, line 148: So this is where other metrics will be registered once we "move" the rest of the metrics to this pattern?

cmd/main.go
```diff
@@ -207,8 +209,11 @@ func main() {
 		}
 
 		mgr = mcmgr.GetLocalManager()
+		metrics.Registry.MustRegister(&dnsMetrics.RemoteClusterCollector{Provider: clusterProvider})
 	}
 
+	metrics.Registry.MustRegister(&dnsMetrics.LocalCollector{Ctx: ctx, Mgr: mgr})
```
maksymvavilov, line 215: Do I get this correctly:
All we need to do is register a collector, and it will `Collect()` on each scrape? 
Meaning that we don't need to do anything else inside the code. How often will the `Collect()` happen (my concern is that it is a lot of `.List()` calls)?

internal/metrics/metrics.go
```diff
@@ -108,6 +107,65 @@ func (m *recordReadyMetric) Publish() {
...
+		if record.IsAuthoritativeRecord() {
+			specString, err := hash.GetCanonicalString(record.Spec)
+			if err != nil {
+				logger.Info("failed to list dns records", "status", "error")
+				continue
+			}
+			value := hash.ToBase36HashLen(specString, 6)
+			ch <- prometheus.MustNewConstMetric(
```
maksymvavilov, line 156: the `NewConstMetric` and `Must...` variant seems to be [throw-away](https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#bound-service-account-tokens) metric. Does it mean that once the next `Collect()` event happens, all previous metrics are being removed? 
Meaning it will make sure we have metrics only for "live" records

maksymvavilov (approved):
Looks good. We should consider moving all the metrics to collectors at some point. This seems to be the best approach considering that we might be dealing with a lot of records. 

Just a few questions to clarify my understanding

internal/metrics/metrics.go
Boomatang: Not sure what you are meant to be linking to there. Yes, they are a "throw-away" metric that is not stored in the prometheus cache with in the operator. This means all previous metrics are removed. I have only looked at using this with gauge metric, and am not sure how it would have counter metric. 

Remember it is when we add labels based on the user generated content (resource name), we also need a way to remove of the metric.

mikenairn (commented):
I ran through a few tests as i was unsure about what was happening just looking at it. Seems to work as i would expect:

```bash
make multicluster-local-setup CLUSTER_COUNT=3 PRIMARY_CLUSTER_COUNT=2 DEPLOY=false
```

Run in separate terminals:
```bash
kubectl config use-context kind-kuadrant-dns-local-1 && make run RUN_METRICS_ADDR=:8080 RUN_HEALTH_ADDR=:8081 RUN_DELEGATION_ROLE=primary
kubectl config use-context kind-kuadrant-dns-local-2 && make run RUN_METRICS_ADDR=:8082 RUN_HEALTH_ADDR=:8083 RUN_DELEGATION_ROLE=primary
kubectl config use-context kind-kuadrant-dns-local-3 && make run RUN_METRICS_ADDR=:8084 RUN_HEALTH_ADDR=:8085 RUN_DELEGATION_ROLE=secondary
```

## Verify dns_provider_active_multi_cluster_count

```
curl -s 127.0.0.1:8080/metrics | grep ^dns_provider_active_multi_cluster_count
dns_provider_active_multi_cluster_count 2
curl -s 127.0.0.1:8082/metrics | grep ^dns_provider_active_multi_cluster_count
dns_provider_active_multi_cluster_count 2
curl -s 127.0.0.1:8084/metrics | grep ^dns_provider_active_multi_cluster_count
```

Delete one cluster secret:
```
kubectl delete secret/kind-kuadrant-dns-local-3 -n dns-operator-system --context kind-kuadrant-dns-local-1
secret "kind-kuadrant-dns-local-3" deleted
curl -s 127.0.0.1:8080/metrics | grep ^dns_provider_active_multi_cluster_count
dns_provider_active_multi_cluster_count 1
curl -s 127.0.0.1:8082/metrics | grep ^dns_provider_active_multi_cluster_count
dns_provider_active_multi_cluster_count 2
```

Delete all cluster secrets:
```
make multicluster-local-setup-delete-kubeconfig-secrets CLUSTER_COUNT=3 PRIMARY_CLUSTER_COUNT=2 DEPLOY=false
Switched to context "kind-kuadrant-dns-local-1".
secret "kind-kuadrant-dns-local-2" deleted
Switched to context "kind-kuadrant-dns-local-2".
secret "kind-kuadrant-dns-local-1" deleted
secret "kind-kuadrant-dns-local-3" deleted
Switched to context "kind-kuadrant-dns-local-3".
No resources found
curl -s 127.0.0.1:8080/metrics | grep ^dns_provider_active_multi_cluster_count
dns_provider_active_multi_cluster_count 0
curl -s 127.0.0.1:8082/metrics | grep ^dns_provider_active_multi_cluster_count
dns_provider_active_multi_cluster_count 0
```

Create all cluster secrets:
```
make multicluster-local-setup-create-kubeconfig-secrets CLUSTER_COUNT=3 PRIMARY_CLUSTER_COUNT=2 DEPLOY=false
curl -s 127.0.0.1:8080/metrics | grep ^dns_provider_active_multi_cluster_count
dns_provider_active_multi_cluster_count 2
curl -s 127.0.0.1:8082/metrics | grep ^dns_provider_active_multi_cluster_count
dns_provider_active_multi_cluster_count 2
```

## Verify dns_provider_authoritative_record_spec_info

```
kubectl apply -f config/local-setup/dnsrecords/delegating/coredns/simple/dnsrecord-simple-coredns-cluster1.yaml --context kind-kuadrant-dns-local-1 -n dnstest
curl -s 127.0.0.1:8080/metrics | grep ^dns_provider_authoritative_record_spec_info
dns_provider_authoritative_record_spec_info{dns_record_name="authoritative-record-1c8db9d0",dns_record_namespace="dnstest",root_host="simple.k.example.com",sha="1hqj2q"} 1
curl -s 127.0.0.1:8082/metrics | grep ^dns_provider_authoritative_record_spec_info
dns_provider_authoritative_record_spec_info{dns_record_name="authoritative-record-1c8db9d0",dns_record_namespace="dnstest",root_host="simple.k.example.com",sha="1hqj2q"} 1
curl -s 127.0.0.1:8084/metrics | grep ^dns_provider_authoritative_record_spec_info
```

Edit the record and update:
```
kubectl apply -f config/local-setup/dnsrecords/delegating/coredns/simple/dnsrecord-simple-coredns-cluster1.yaml --context kind-kuadrant-dns-local-1 -n dnstest
dnsrecord.kuadrant.io/dnsrecord-simple-coredns-cluster1 configured
curl -s 127.0.0.1:8080/metrics | grep ^dns_provider_authoritative_record_spec_info
dns_provider_authoritative_record_spec_info{dns_record_name="authoritative-record-1c8db9d0",dns_record_namespace="dnstest",root_host="simple.k.example.com",sha="v9woqc"} 1
curl -s 127.0.0.1:8082/metrics | grep ^dns_provider_authoritative_record_spec_info
dns_provider_authoritative_record_spec_info{dns_record_name="authoritative-record-1c8db9d0",dns_record_namespace="dnstest",root_host="simple.k.example.com",sha="v9woqc"} 1
```

internal/metrics/metrics.go
maksymvavilov: Oh, the buffer betrayed me. This was the intended link https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#NewConstMetric but you answered it regardless

PR #595, opened 2025-10-01
dep: go 1.24.6 

KevFan:
# Description
Closes: https://github.com/Kuadrant/limitador-operator/issues/193

- Use go version from `go.mod` in github actions
  - `setup-go` supports a `go-version-file` field 
  -  `go-release-action` supports a path to.mod - https://github.com/wangyoucao577/go-release-action
- Bump go version to 1.24.6

Note: CoreDNS plugin go version is **not** updated


mikenairn:
@KevFan This looks fine to me. I'm goign to create a release of the dns operator today. Not seeing an urgent need to merge this before hand so might just wait until that release is done and then approve.

PR #597, opened 2025-10-03
add clenup cmd for old TXTs

maksymvavilov:
I'm also modifying existing commands to make everything more... friendly. 
To verify 
1. `make build-cli` 
2. `make cp-cli`
3. Create a provider secret for aws 
4. `./bin/kubectl-dns delete-old-txt [provider-secret-name] -n  [namespace]` (no domain filter to see all existing records)
5. checkout an old version of dns-operator by `git checkout 3e5ec5bbcf4401cdd58448632fa1311acc2ea6ba`
6. `make run`
7. deploy a DNSRecord 
8.  `./bin/kubectl-dns delete-old-txt [provider-secret-name] -n [namespace] -d [your-domain]` 
9. agree to a clenup
10. Check with the provider to see old records are gone

### DIFF
--- a/cmd/plugin/delete_old_txt.go
+++ b/cmd/plugin/delete_old_txt.go
@@ -1,0 +1,203 @@
+package main
+
+import (
+	"bufio"
+	"context"
+	"fmt"
+	"os"
+	"strings"
+	"time"
+
+	"github.com/spf13/cobra"
+
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/kubernetes/scheme"
+	controllerruntime "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	externaldns "sigs.k8s.io/external-dns/endpoint"
+	"sigs.k8s.io/external-dns/plan"
+
+	"github.com/kuadrant/dns-operator/api/v1alpha1"
+	"github.com/kuadrant/dns-operator/internal/common/slice"
+	"github.com/kuadrant/dns-operator/internal/external-dns/registry"
+	"github.com/kuadrant/dns-operator/internal/provider"
+	_ "github.com/kuadrant/dns-operator/internal/provider/aws"
+	_ "github.com/kuadrant/dns-operator/internal/provider/azure"
+	_ "github.com/kuadrant/dns-operator/internal/provider/coredns"
+	"github.com/kuadrant/dns-operator/internal/provider/endpoint"
+	_ "github.com/kuadrant/dns-operator/internal/provider/google"
+)
+
+var deleteOldTXTCMD = &cobra.Command{
+	Use:  "delete-old-txt [provider-secret-name]",
+	RunE: deleteOldTXT,
+}
+
+type DeleteOldTXTCMDFlags struct {
+	ns     string
+	domain string
+}
+
+var deleteOldTXTCMDFlags DeleteOldTXTCMDFlags
+
+func init() {
+	deleteOldTXTCMD.Flags().StringVarP(&deleteOldTXTCMDFlags.ns, "namespace", "n", "default", "namespace of a provider secret")
+	deleteOldTXTCMD.Flags().StringVarP(&deleteOldTXTCMDFlags.domain, "domain", "d", "", "domain filter to appy to endpoints")
+
+}
+
+func deleteOldTXT(_ *cobra.Command, args []string) error {
+	log = logf.Log.WithName("delete-old-txt")
+
+	d := time.Now().Add(time.Minute * 5)
+	ctx, cancel := context.WithDeadline(context.Background(), d)
+	defer cancel()
+
+	if len(args) > 1 {
+		return fmt.Errorf("too many arguments (expected one secret name)")
+	}
+
+	// setup client
+	runtime.Must(v1alpha1.AddToScheme(scheme.Scheme))
+
+	config := controllerruntime.GetConfigOrDie()
+	k8sClient, err := client.New(config, client.Options{Scheme: scheme.Scheme})
+	if err != nil {
+		fmt.Printf("failed to create client: %v\n", err)
+		return nil
+	}
+
+	secretList := &v1.SecretList{}
+	secret := &v1.Secret{}
+
+	// looking for a default secret
+	if len(args) == 0 {
+		err = k8sClient.List(ctx, secretList, &client.ListOptions{
+			LabelSelector: labels.SelectorFromSet(map[string]string{
+				v1alpha1.DefaultProviderSecretLabel: "true",
+			}),
+			Namespace: deleteOldTXTCMDFlags.ns,
+		})
+		if err != nil {
+			fmt.Printf("failed to list secrets: %v\n", err)
+			return nil
+		}
+
+		if len(secretList.Items) != 1 {
+			fmt.Printf("unexpected nmber of secrets: %d; expected 1 default secret\n", len(secretList.Items))
+			return nil
+		}
+		secret = &secretList.Items[0]
+
+	} else {
+		err = k8sClient.Get(ctx, client.ObjectKey{Name: args[0], Namespace: deleteOldTXTCMDFlags.ns}, secret)
+		if err != nil {
+			fmt.Printf("failed to get secret: %v\n", err)
+			return nil
+		}
+	}
+	// factory to get a list of all endpoints
+	dynClient, err := dynamic.NewForConfig(config)
+	if err != nil {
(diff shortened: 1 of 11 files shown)


cmd/plugin/delete_old_txt.go
```diff
@@ -0,0 +1,203 @@
...
+	for _, ep := range endpoints {
+		fmt.Printf("ep: %s\n", ep.DNSName)
+	}
+
+	// display what is about to be deleted
+	fmt.Printf("TXT records (%d) to be deleted:\n", len(endpoints))
+	for _, txtRecord := range endpoints {
+		fmt.Printf("%s\t IN\t %s\t %s\n", txtRecord.DNSName, txtRecord.RecordType, txtRecord.Targets)
+	}
+	fmt.Printf("Do you want to proceed? [Y/N]\n")
```
Boomatang, line 180: Its a bit odd to put the user input on a new line.
```suggestion
	fmt.Printf("Do you want to proceed? [Y/N]: ")
```

cmd/plugin/delete_old_txt.go
```diff
@@ -0,0 +1,203 @@
...
+	reader := bufio.NewReader(os.Stdin)
+	answer, _ := reader.ReadString('\n')
```
Boomatang, line 182: Why is the error being ignore?

cmd/plugin/delete_old_txt.go
```diff
@@ -0,0 +1,203 @@
...
+			fmt.Printf("failed to delete old TXT records: %v\n", err)
+		}
+		return nil
+	case "n":
+		// do nothing
+		return nil
+	default:
+		fmt.Printf("unknown answer %s\n", answer)
+		return nil
+	}
```
Boomatang, line 202: It is common for a cli to that asks a "can we proceed" question to default to "no". In this case they would get an unknown answer response, over we are cancelling. 

For the no case it might be good to have some feedback to the user saying the process is being cancelled.

cmd/plugin/plugin.go
```diff
@@ -18,39 +18,36 @@ var (
...
-	root.AddCommand(secretGenerationCommand())
+func init() {
+	rootCMD.SetArgs(os.Args[1:])
+	rootCMD.PersistentFlags().BoolVarP(&verbose, "verbose", "v", true, "verbose output")
 
-	if err := root.Execute(); err != nil {
-		os.Exit(1)
-	}
+	rootCMD.AddCommand(versionCMD, deleteOldTXTCMD, getZoneRecordsCMD)
+	rootCMD.AddCommand(generateSecretCMD, deleteOwnerCMD)
```
Boomatang, line 46: I feel a bit odd about this one. Why not add them all in same call? Do the grouping mean something?

internal/external-dns/registry/txt.go
```diff
@@ -480,7 +480,7 @@ func (im *TXTRegistry) removeFromCache(ep *endpoint.Endpoint) {
 	}
 }
 
-func (im *TXTRegistry) NewLabelsFromString(labelText string, aesKey []byte) (owner, version string, labels endpoint.Labels, err error) {
```
Boomatang, line 483: Why was this change need?

cmd/plugin/delete_old_txt.go
Boomatang: Had a though here. Can we add a confirm flag to over ride this prompt check. This user required input means this cli command can never be used in a ci/cd pipeline.

cmd/plugin/delete_old_txt.go
```diff
@@ -0,0 +1,203 @@
...
+	"github.com/kuadrant/dns-operator/internal/provider"
+	_ "github.com/kuadrant/dns-operator/internal/provider/aws"
+	_ "github.com/kuadrant/dns-operator/internal/provider/azure"
+	_ "github.com/kuadrant/dns-operator/internal/provider/coredns"
+	"github.com/kuadrant/dns-operator/internal/provider/endpoint"
+	_ "github.com/kuadrant/dns-operator/internal/provider/google"
+)
+
+var deleteOldTXTCMD = &cobra.Command{
+	Use:  "delete-old-txt [provider-secret-name]",
```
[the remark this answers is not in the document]
maksymvavilov: Added a bit of descriptions and renamed the command

cmd/plugin/delete_old_txt.go
```diff
@@ -0,0 +1,203 @@
...
+type DeleteOldTXTCMDFlags struct {
+	ns     string
+	domain string
+}
+
+var deleteOldTXTCMDFlags DeleteOldTXTCMDFlags
+
+func init() {
+	deleteOldTXTCMD.Flags().StringVarP(&deleteOldTXTCMDFlags.ns, "namespace", "n", "default", "namespace of a provider secret")
+	deleteOldTXTCMD.Flags().StringVarP(&deleteOldTXTCMDFlags.domain, "domain", "d", "", "domain filter to appy to endpoints")
```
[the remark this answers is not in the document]
maksymvavilov: if the ns is empty, we will do the same that `kubectl` does - look in the `default` namespace 
if the domain filter is empty, we will not have any applied: i.e. act on all the records we have access to

cmd/plugin/delete_old_txt.go
```diff
@@ -0,0 +1,203 @@
...
+	if err != nil {
+		fmt.Printf("failed to create client: %v\n", err)
+		return nil
+	}
+
+	secretList := &v1.SecretList{}
+	secret := &v1.Secret{}
+
+	// looking for a default secret
+	if len(args) == 0 {
```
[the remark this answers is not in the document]
maksymvavilov: There are no logs at all, aren't there? I'll add a few.

cmd/plugin/delete_old_txt.go
```diff
@@ -0,0 +1,203 @@
...
+	fmt.Printf("filtered endpoints len %d\n", len(endpoints))
+	for _, ep := range endpoints {
+		fmt.Printf("ep: %s\n", ep.DNSName)
+	}
+
+	// display what is about to be deleted
+	fmt.Printf("TXT records (%d) to be deleted:\n", len(endpoints))
+	for _, txtRecord := range endpoints {
+		fmt.Printf("%s\t IN\t %s\t %s\n", txtRecord.DNSName, txtRecord.RecordType, txtRecord.Targets)
+	}
```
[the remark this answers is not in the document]
maksymvavilov: oh, the one above was a debug I forgot to remove. This one is meant to be for the user :facepalm:

cmd/plugin/delete_old_txt.go
maksymvavilov: hmm, I'd be extremely concerned if anyone would try to have a tool that deletes records somewhere in a pipeline. 
The reason I go for the input is to make sure the user is aware of what is happening and not just fat-fingered their way with autocomplete here and nuked their whole TXT registry of something. 

That said I see no harm in having a flag (like [dnf](https://dnf.readthedocs.io/en/latest/command_ref.html) has, for example)

cmd/plugin/delete_old_txt.go
maksymvavilov: fair. Also added a message for a "yes" to make them know we are done

cmd/plugin/plugin.go
maksymvavilov: grouping means nothing. One command it is

internal/external-dns/registry/txt.go
maksymvavilov: This is not to initialise a registry to translate targets into labels. 
I have no idea why it was a part of the registry to begin with - it uses nothing from it for its logic :thinking:

cmd/plugin/cleanup_old_txt.go
```diff
@@ -0,0 +1,203 @@
...
+
+	// setup client
+	runtime.Must(v1alpha1.AddToScheme(scheme.Scheme))
+
+	config := controllerruntime.GetConfigOrDie()
+	k8sClient, err := client.New(config, client.Options{Scheme: scheme.Scheme})
+	if err != nil {
+		fmt.Printf("failed to create client: %v\n", err)
+		return nil
+	}
```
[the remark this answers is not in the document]
Boomatang: I see the commands success as follows 
- if we successfully delete the TXT records, or 
- we could not find any records to delete
Anything other than that the tool failed to do what it was meant to do, and should return an error to the user. Even if it was the users fault that we couldn't do something. The error status is not about being on fire, it is about we couldn't do what we were asked to do.

Side note: our make targets don't report a phase failed, and number of time this has caught me out. The only way to know that it failed was to look over the logs. I have my shell prompt to show me the error status of the last command that I ran.

cmd/plugin/delete_old_txt.go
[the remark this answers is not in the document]
maksymvavilov: added a bit of info to the flag description

cmd/plugin/delete_old_txt.go
maksymvavilov: Added the check

PR #598, opened 2025-10-06
chore: Update kind (0.30.0) and node image (1.32.0)

mikenairn:
* Updates the kind version to latest (0.30.0)
* Update the kind node image to (1.32.0) this is inline with the kubernetes version used by the latest version of OpenShift (4.19)
* Remove erroneous arg to image archive command (kind-load-image)


Makefile
```diff
@@ -421,7 +421,7 @@ KUBECTL_DNS ?= $(LOCALBIN)/kubectl-dns
 KUSTOMIZE_VERSION ?= v5.5.0
 CONTROLLER_TOOLS_VERSION ?= v0.14.0
 OPENSHIFT_GOIMPORTS_VERSION ?= c70783e636f2213cac683f6865d88c5edace3157
-KIND_VERSION = v0.20.0
+KIND_VERSION = v0.30.0
```
thomasmaas, line 424: Would be good to require the same version in all pieces of Kuadrant https://github.com/search?q=org%3AKuadrant%20KIND_VERSION%20%3D&type=code

PR #615, opened 2025-10-22
add slash command to prepare current branch for a PR

philbrookes:
To use this from claude:
- make whatever changes you want to make
- commit the changes
- run `/prepare-for-pr` in claude.


philbrookes:
I agree with @Boomatang comment, this would be better automated into a pre-commit hook.

Issue #204, opened 2024-08-01
Create kubectl plugin for commonly required debugging information


philbrookes:

maksymvavilov:
The [kuadrantctl](https://github.com/Kuadrant/kuadrantctl) is the place to start. Creating our own "dns-operator based" cli makes no sense

PR #617, opened 2025-10-30
remove relatedEndpoints

maksymvavilov:
result of the #204 being done. 

Removing the code to put endpoints and a test associated with it. 
Haven't found any reference for the in the docs, which seems odd.


mikenairn (approved):

philbrookes:
What happens if a record has related endpoints in it's status and the operator changes to this code?

maksymvavilov:
> What happens if a record has related endpoints in it's status and the operator changes to this code?

They will get removed at the next reconciliation

PR #629, opened 2025-11-07
refactor: Add base DNSRecord reconciler

### LINKED ISSUE #591: refactor: Add base controller and reconcile local and remote common functionality
**What**

Add a base dnsrecord controller that all other controllers(DNSRecordReconciler and RemoteDNSRecordReconciler) extend and remove duplication of common methods.

Ideal opportunity to increase unit test coverage of the controllers.

Possible common functionality:

* setLogger()
* updateStatus()
* publishRecord()
* deleteRecord()
* applyChanges()
   * Should be updated to remove all references to health checks and status updates and return to what it was originally intended to be for, applying changes. The other stuff it's now doing should be done outside of it as part of the `Reconcile` flow for the local record.

mikenairn:
closes #591 

Adds BaseDNSRecordReconciler that is intended to be embedded by concrete DNSRecord reconciler implementations that adds common helper functions:
* setLogger() - adapt the current logger with common values, rootHost, ownerID etc..
* getDNSProvider() - returns a concrete provider implementation for the given DNSRecordAccessor
* deleteRecord() - delete the given record from the provider
* publishRecord() - publish the given record to the provider
* applyChanges() - applies changes using the registry and plan for the current record and provider

Update DNSRecordReconciler and RemoteDNSRecordReconciler to use embedded BaseDNSRecordReconciler.

Updates HealthChecks to use healthCheckAdapter that wraps an accessor resource and manipulates the endpoints based on the configured health checks and updates the health related status on status update.

### DIFF
--- a/internal/controller/base_dnsrecord_reconciler.go
+++ b/internal/controller/base_dnsrecord_reconciler.go
@@ -113,7 +113,8 @@ func publishRecord(ctx context.Context, dnsRecord DNSRecordAccessor, dnsProvider
 // The error is nil only if the changes were successfully applied or there were no changes to be made.
 func applyChanges(ctx context.Context, dnsRecord DNSRecordAccessor, dnsProvider provider.Provider, isDelete bool) (bool, error) {
 	logger := log.FromContext(ctx)
-	rootDomainName := dnsRecord.GetRootHost()
+	//ToDo We can't use GetRootHost() here as it currently removes any wildcard prefix which needs to be maintained in this scenario.
+	rootDomainName := dnsRecord.GetSpec().RootHost
 	zoneDomainFilter := externaldnsendpoint.NewDomainFilter([]string{dnsRecord.GetZoneDomainName()})
 	managedDNSRecordTypes := []string{externaldnsendpoint.RecordTypeA, externaldnsendpoint.RecordTypeAAAA, externaldnsendpoint.RecordTypeCNAME}
 	var excludeDNSRecordTypes []string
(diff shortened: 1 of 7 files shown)


philbrookes:
What I see in this refactor, is a lot of the common functions are in the base reconciler and used from the implementations, which reduces the duplicated code.

I was wondering if you see any value in having a `reconcile` that is passed a reconciler that can call out to various stage functions from the base and have those extended in the implementations?

e.g. (in metacode)
```
type reconciler interface
   func postReconcile
   func ensureFinalizer
   func ensureOwner
   func processDelete
   ... other stuff
```
```
base implements reconciler
r *base -> ensureFinalizer()
   some logic
r *base -> ensureOwner()
   some logic
r *base -> processDelete(x, y, z)
   generic delete logic
... generic implementations of the other stuff
```

```
dnsrecord_reconciler extends base
r *dnsrecord_reconcilier->processDelete(x, y, z)
   r.base->processDelete(x, y, z)
   specific deletion logic that differs for local records.
```

```
remote_dnsrecord_reconciler extends base
r *remote_dnsrecord_reconcilier->processDelete(x, y, z)
   r.base->processDelete(x, y, z)
   specific deletion logic that differs for remote records.
```

```
reconcile(r reconciler)
   //amongst other stuff
   defer(r.postReconcile)
   r.ensureFinalizer()
   r.ensureOwner()
   if isDeleting() r.processDelete(x, y, z)
```

This would also allow us to put the general "flow" of reconcile into the reconcile function, but override it if needed in an implementation, so the contents of the implementations would be only the stuff that differs from the... generic base flow.

mikenairn:
> I was wondering if you see any value in having a `reconcile` that is passed a reconciler that can call out to various stage functions from the base and have those extended in the implementations?

This might be possible, but not what i was trying to achieve with #591. I do however see enough differences in the flows of the two current types (local and remote) that this might lead to a lot of methods on the interface that will just be empty implementations in one or the other. It might be worth doing a spike to try it.

philbrookes (approved):

Issue #627, opened 2025-11-04
DNS Operator add/update group and targets to registry entries.


philbrookes:
We want 2 new entries in each registry entry:
- The runtime groupID, if present.
- The targets this ownerID wants present (remove the rootHost) only when groupID present.

For example, if a DNS Record with groupID: `group1` has an endpoint of "host in CNAME 60 target1,target2". The resulting registry entry should be (alongside whatever current information is put into the registry) something like:
```
version=1;ownerID=sdfsdf;groupID=group1;targets=target1&&target2
```

In terms of implementation, before publish is called: 
- https://github.com/Kuadrant/dns-operator/blob/main/internal/controller/dnsrecord_controller.go#L354
- https://github.com/Kuadrant/dns-operator/blob/main/internal/controller/remote_dnsrecord_controller.go#L211

We want something like
```go
if groupID != "" {
   e.labels[group] = groupID
   e.labels[targets] = e.targets
} else {
   // clean up logic
}
```

unit test of usage: https://github.com/Kuadrant/dns-operator/blob/main/internal/external-dns/registry/txt_test.go#L190-L196

This will enable us to work out:
1) Which groups own a particular endpoint
2) Which groups own any particular target

PR #632, opened 2025-11-11
Add group and targets to txt registry entries

Boomatang:
Closes: #627 

- **ADD: Group to TXT Record**

### DIFF
--- a/test/e2e/group_setting_test.go
+++ b/test/e2e/group_setting_test.go
@@ -1,0 +1,419 @@
+//go:build e2e
+
+package e2e
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	. "github.com/onsi/gomega/gstruct"
+
+	appsv1 "k8s.io/api/apps/v1"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	externaldnsendpoint "sigs.k8s.io/external-dns/endpoint"
+
+	"github.com/kuadrant/dns-operator/api/v1alpha1"
+	"github.com/kuadrant/dns-operator/internal/common/hash"
+	. "github.com/kuadrant/dns-operator/test/e2e/helpers"
+)
+
+const (
+	controllerNamespace      = "dns-operator-system"
+	controllerDeploymentName = "dns-operator-controller-manager"
+	deploymentReadyTimeout   = 3 * time.Minute
+)
+
+// Group tests are Serial and Ordered to minimize deployment restarts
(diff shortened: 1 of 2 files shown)


internal/external-dns/plan/plan.go
```diff
@@ -635,6 +635,15 @@ func shouldUpdateProviderSpecific(desired, current *endpoint.Endpoint) bool {
...
 
+func shouldUpdateGroup(desired, current *endpoint.Endpoint) bool {
+	currentGroup, hasCurrentGroup := current.Labels["group"]
+	desiredGroup, hasDesiredGroup := desired.Labels["group"]
+	if hasCurrentGroup && hasDesiredGroup {
+		return currentGroup != desiredGroup
+	}
+	return hasCurrentGroup != hasDesiredGroup
+}
+
```
mikenairn: Hmm OK, i guess we need this so that changes to the group result in an update. I have kind of looked into this before, wondering if it would be better to make this more generic and update on any label changes instead of hard coding specific labels like this.

Nothing to do here, this will work, just commenting.

mikenairn (commented):
In a primary/secondary scenario what controller(local/remote) is deciding what the group that goes into the DNSRecord status should be?

test/e2e/group_setting_test.go
```diff
@@ -0,0 +1,419 @@
...
+			g.Expect(dnsRecord.Status.Conditions).To(
+				ContainElement(MatchFields(IgnoreExtras, Fields{
+					"Type":   Equal(string(v1alpha1.ConditionTypeReady)),
+					"Status": Equal(metav1.ConditionTrue),
+				})),
+			)
+			g.Expect(string(dnsRecord.Status.Group)).To(BeEmpty())
+		}, recordsReadyMaxDuration, 10*time.Second, ctx).Should(Succeed())
+
+		DeferCleanup(func(ctx SpecContext) {
```
philbrookes: I think this DeferCleanup should be earlier in the test, what's the point of deferring it, when it's the last thing done anyway. And won't cleanup not run if the tests fail before this defer is defined?

test/e2e/group_setting_test.go
```diff
@@ -0,0 +1,419 @@
...
+			g.Expect(string(dnsRecord.Status.Group)).To(Equal(updatedGroup))
+		}, recordsReadyMaxDuration, 10*time.Second, ctx).Should(Succeed())
+
+		By("verifying updated group is present in provider TXT records")
+		if txtRegistryEnabled {
+			err := verifyGroupInTXTRecords(ctx, k8sClient, dnsRecord, updatedGroup)
+			Expect(err).NotTo(HaveOccurred())
+		}
+
+		DeferCleanup(func(ctx SpecContext) {
```
philbrookes: and here.

internal/external-dns/plan/plan.go
mikenairn: Undo these changes in favour of https://github.com/Kuadrant/dns-operator/pull/643 once merged.

mikenairn:
Closing, replaced by #652

PR #635, opened 2025-11-12
Bug: with getting dnsrecord list

Boomatang:
When the using a dnsRecord the name should equal the roothost when getting the endpoints.


philbrookes:
I think it'd be preferable to pass the rootHost to getEndpoints as an arg, rather than changing the value of a global.

philbrookes (approved):

Issue #639, opened 2025-11-13
DNS-Operator doesn't have logLevel overridable with env var


crstrn13:
We are in the middle of enabling debug level across all components in QE environments. As the DNS-Operator is not managed by a CR and directly by the subscription, there is no way to override the default `--zap-log-level` through an environment variable. I checked the code and I can see there is such a mechanism and adding the same behavior to the logLevel shouldn't be a problem.

PR #642, opened 2025-11-14
fix: wrapped zap log level and mode with custom flags.

crstrn13:
Tries to fix: #639

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -388,10 +388,10 @@ func overrideControllerFlags() {
 			setupLog.Info(fmt.Sprintf("overriding %s flag with \"%s\" value", groupKey.Flag(), v))
 		case logModeKey.Envar():
 			logMode = v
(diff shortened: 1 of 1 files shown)


maksymvavilov (commented):
Pushed changes that I've suggested before. We will need another soul to review this PR

cmd/main.go
```diff
@@ -377,6 +386,12 @@ func overrideControllerFlags() {
 				os.Exit(1)
 			}
 			setupLog.Info(fmt.Sprintf("overriding %s flag with \"%s\" value", groupKey.Flag(), v))
+		case logModeKey.Envar():
+			logMode = v
+			setupLog.Info(fmt.Sprintf("overriding %s flag with \"%s\" value", groupKey.Flag(), v))
```
philbrookes: should this be logModeKey?

PR #652, opened 2025-11-20
Add group and targets to registry entries

### LINKED ISSUE #627: DNS Operator add/update group and targets to registry entries.

mikenairn:
closes #627 
replaces https://github.com/Kuadrant/dns-operator/pull/632:
* Brings it up to date after refactor of reconcilers 
* Removes TXT registry changes
* Removes e2e tests (Will be added separately as part of #665, currently blocked by #664)

### DIFF
--- a/internal/controller/dnsrecord_accessor.go
+++ b/internal/controller/dnsrecord_accessor.go
@@ -32,6 +32,7 @@ type DNSRecordAccessor interface {
 	v1alpha1.ProviderAccessor
 	GetDNSRecord() *v1alpha1.DNSRecord
 	GetOwnerID() string
+	GetGroup() types.Group
 	GetRootHost() string
 	GetZoneDomainName() string
 	GetZoneID() string
@@ -67,6 +68,10 @@ func (s *DNSRecord) GetDNSRecord() *v1alpha1.DNSRecord {
 
 func (s *DNSRecord) GetOwnerID() string {
 	return s.GetStatus().OwnerID
+}
+
+func (s *DNSRecord) GetGroup() types.Group {
+	return s.GetStatus().Group
 }
 
 func (s *DNSRecord) GetZoneDomainName() string {
@@ -147,6 +152,10 @@ func (s *RemoteDNSRecord) GetDNSRecord() *v1alpha1.DNSRecord {
 
 func (s *RemoteDNSRecord) GetOwnerID() string {
 	return s.DNSRecord.Status.OwnerID
+}
+
+func (s *RemoteDNSRecord) GetGroup() types.Group {
+	return s.DNSRecord.Status.Group
 }
 
 func (s *RemoteDNSRecord) GetZoneDomainName() string {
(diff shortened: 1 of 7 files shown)


internal/controller/dnsrecord_accessor.go
```diff
@@ -56,6 +59,20 @@ type DNSRecord struct {
...
 func (s *DNSRecord) GetEndpoints() []*externaldns.Endpoint {
+	//Modify endpoints to include records current group if set or ensure it is removed if unset
+	recordGroup := s.GetGroup()
+	for _, ep := range s.GetSpec().Endpoints {
+		if recordGroup.IsSet() {
+			if ep.Labels == nil {
+				ep.Labels = externaldns.NewLabels()
+			}
+			ep.Labels[types.GroupLabelKey] = recordGroup.String()
+			ep.Labels[types.TargetsLabelKey] = ep.Targets.String()
```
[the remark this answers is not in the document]
mikenairn: We said we would only add these if we had a group. Adding unbounded labels like this has a possibility of breaking the TXT record length limit so IMO i would rather it was only happening if groups were set to avoid it becoming an issue for normal operation where we are not using groups.

internal/controller/dnsrecord_controller.go
```diff
@@ -225,6 +225,8 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		return r.updateStatus(ctx, previous, dnsRecord, false, err)
 	}
 
+	dnsRecord.SetStatusGroup(r.Group)
```
philbrookes: This is causing me a bit of a headache, if this is empty in the remote reconciler, I don't know if that's because the remote has no group set, or because it hasn't set it in the status yet.

mikenairn: The remote reconcile won't proceed until the local one has added the owner and group and set the "ReadyForDelegation" status condition. Not really seeing the problem, if it's empty then it's because it has no group set by the local controller. I asked a question about this on here https://github.com/Kuadrant/dns-operator/pull/632#pullrequestreview-3453814300 but got no reply. Only makes sense to me that the local controller is setting the group though, so that's what this is doing.

PR #666, opened 2025-12-04
Gh 663

maksymvavilov:


philbrookes (changes requested):
It'd be easier to read this if you opened the PR against my branch for now, until it merges. 

2 small changes, plus the failing CI needs addressing (unless that's due to my branch?)

PR #669, opened 2025-12-09
DNS Groups.

philbrookes:
This PR introduces DNS Groups functionality to enable active-passive failover for DNS records across multiple clusters. Groups allow operators to control which cluster's DNS records are published based on a configurable "active groups" list, enabling scenarios like:

  - Multi-cluster disaster recovery with controlled failover
  - Geographic failover between regions
  - Blue-green deployments at the DNS level
  - Staged rollouts with DNS-based traffic shifting

###  API Changes:
  - Added ActiveGroups field to DNSRecordStatus to track the current list of active groups
  - Added IsActive() method to DNSRecord type (interface method for group-aware behavior)
  - New condition type ConditionTypeActive with reasons:
    - ConditionReasonInActiveGroup - Record's group is active and will be published
    - ConditionReasonNotInActiveGroup - Record's group is inactive and won't be published

###  Core Implementation (internal/controller/dnsrecord_groups.go - new file):
  - TXTResolver interface for querying active groups from DNS
  - DefaultTXTResolver implementation with support for custom nameservers
  - GroupAdapter wraps DNSRecordAccessor to provide group-aware IsActive() behavior
  - getActiveGroups() queries the special TXT record kuadrant-active-groups.<zone> to determine which groups are active
  - unpublishInactiveGroups() cleans up DNS records from previously active groups during failover

###  Controller Changes (internal/controller/dnsrecord_controller.go):
  - Integrated group adapter into reconciliation flow
  - Inactive group records exit early with 15s requeue time
  - Active group records unpublish inactive group DNS records after successful publish
  - Authoritative records are exempt from all of these changes

 ### Test Coverage:
  - Added integration tests
  - Tests cover:
    - Multiple groups with active/inactive switching
    - Ungrouped records (always active)
    - Inactive group cleanup
    - Status condition updates
    - Multi-cluster coordination

 ## How It Works

  1. Group Assignment

  Each DNS operator instance is started with a group identifier:
  --group=us-east
  or
  GROUP=us-east

  Records managed by that operator inherit the group assignment in their status.

  2. Active Groups Declaration

  The active groups list is stored as a TXT record in DNS:
  kuadrant-active-groups.example.com  TXT  "groups=us-east&&us-west;version=1"

  3. Reconciliation Flow

  Before publishing DNS records, each controller:
  1. Queries the active groups TXT record
  2. Compares its group against the active groups list
  3. If inactive: Updates status condition and requeues (15s)
  4. If active: Publishes its records AND cleans up records from inactive groups

  4. Ungrouped Records

  Records without a group assignment (group="") are always active and published alongside whichever groups are currently active. They will never process unpublishing of records.

##  Example Scenario:

  Setup:
  - Cluster A (group="us-east") has DNSRecord → 1.2.3.4
  - Cluster B (group="us-west") has DNSRecord → 5.6.7.8
  - Cluster C (ungrouped) has DNSRecord → 9.9.9.9

  Active groups = ["us-east"]:
    Published: 1.2.3.4, 9.9.9.9

  Switch active groups to ["us-west"]:
    Published: 5.6.7.8, 9.9.9.9
    (Cluster B unpublishes stale 1.2.3.4)

 ## Manual Verification Instructions

  Prerequisites

  - DNS domain managed by a provider (AWS Route53, GCP, Azure, or CoreDNS)
  - Two clusters with DNS operator installed
  - DNS provider credentials configured

  Option 1: Verification with AWS Route53

 ### Setup:

  1. make local-setup with 2 clusters, and deploy true. Then edit the deployments to set the group runtime argument (e.g. us-east and us-west).

  3. Create test DNSRecords:

 #### In cluster-1
```bash
  cat <<EOF | kubectl apply -f -
  kind: DNSRecord
  metadata:
    name: test-record
  spec:
    rootHost: test-groups.example.com
    endpoints:
    - dnsName: test-groups.example.com
      recordType: A
      targets:
      - 1.2.3.4
  EOF
```

#### In cluster-2
```bash
  cat <<EOF | kubectl apply -f -
  kind: DNSRecord
  metadata:
    name: test-record
  spec:
    rootHost: test-groups.example.com
    endpoints:
    - dnsName: test-groups.example.com
      recordType: A
      targets:
      - 5.6.7.8
  EOF
```
  4. Set initial active groups (us-east):
  Using kuadrant-dns-cli, set the active group to us-east

  5. Verify initial state:
  Query DNS (should return 1.2.3.4)
  dig test-groups.example.com +short

  Check Route53 for published records
  aws route53 list-resource-record-sets --hosted-zone-id <zone-id> \
    --query "ResourceRecordSets[?Name=='test-groups.example.com.']"

  6. Test failover to us-west:

  Use dns cli to update the TXT record in Route53 to us-west

  Wait 15-30 seconds for reconciliation, then verify:

  Query DNS (should now return 5.6.7.8)
  dig test-groups.example.com +short

 ### Verification Checklist

  - DNSRecord in active group shows Active: True condition
  - DNSRecord in inactive group shows Active: False condition
  - DNS queries return only targets from active groups and ungrouped clusters
  - Switching active groups triggers cleanup of old records from inactive groups
  - Ungrouped records are always published


config/local-setup/dnsrecords/delegating/coredns/loadbalanced/dnsrecord-loadbalanced-coredns-cluster1.yaml
```diff
@@ -3,7 +3,7 @@ kind: DNSRecord
 metadata:
   name: dnsrecord-loadbalanced-coredns-cluster1
 spec:
-  rootHost: loadbalanced.k.example.com
+  rootHost: k.example.com
```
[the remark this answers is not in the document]
philbrookes: ok, got it. PR here: https://github.com/Kuadrant/dns-operator/pull/677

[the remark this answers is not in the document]
philbrookes: when that merges, I'll undo these changes to the sample files.

config/coredns/Corefile
```diff
@@ -1,11 +1,15 @@
 k.example.com {
     debug
     errors
+    log
+    
+    rewrite name regex kuadrant-active-groups\.(.*)k.example\.com kuadrant-active-groups-coredns.pb.hcpapps.net
+    forward kuadrant-active-groups-coredns.pb.hcpapps.net /etc/resolv.conf
```
philbrookes: this will need to be updated to allow this custom host to be passed in at set up time, somehow, that's for another ticket: https://github.com/Kuadrant/dns-operator/issues/670

types/group.go
```diff
@@ -47,3 +49,17 @@ func (g *Group) IsSet() bool {
 func (g *Group) Labels() map[string]string {
 	return map[string]string{GroupLabelKey: g.String()}
 }
+
+type Groups []Group
+
+func (g Groups) HasGroup(group Group) bool {
+	return slice.Contains(g, func(gElem Group) bool { return gElem == group })
```
Boomatang: Why not use the slices.Contains from the standard library?

internal/controller/dnsrecord_groups.go
```diff
@@ -0,0 +1,318 @@
...
+			logger.V(1).Info("using nameserver", "nameserver", ns, "resolved", nsAddr)
+
+			resolver := &net.Resolver{
+				PreferGo: true,
+				Dial: func(ctx context.Context, network, address string) (net.Conn, error) {
+					dialer := net.Dialer{
+						Timeout: time.Second * 3,
+					}
+					// Always use our specified nameserver, ignoring the address parameter
+					return dialer.DialContext(ctx, network, nsAddr)
```
Boomatang: early returns

Boomatang (commented):
I have a number of questions. I know it is still a draft and subject to change. For the reason I didn't look at any test changes.

api/v1alpha1/dnsrecord_types.go
```diff
@@ -340,6 +343,10 @@ func (s *DNSRecord) IsDeleting() bool {
 	return s.DeletionTimestamp != nil && !s.DeletionTimestamp.IsZero()
 }
 
+func (s *DNSRecord) IsActive() bool {
+	return true
+}
+
```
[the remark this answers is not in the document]
philbrookes: this is always true if the adapter hasn't been adapted into a group adapter  (i.e. no group logic == always active), this method is overridden in the `GroupAdapter`

internal/controller/dnsrecord_groups.go
```diff
@@ -0,0 +1,318 @@
...
+	}
+
+	return ga
+}
+
+func (s *GroupAdapter) IsActive() bool {
+	//this controller is ungrouped
+	return s.GetGroup() == "" ||
+		//no active groups specified or
+		len(s.activeGroups) == 0 ||
```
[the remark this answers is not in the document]
philbrookes: No active groups matches the current set up, so also needs to match current default behaviour. 

The only safe action when no active groups are set, is to have every group be active, in what scenario would a user require all groups inactive? (if that requirement did exist, they could still configure it with a non-existant group name).

internal/controller/dnsrecord_groups.go
```diff
@@ -0,0 +1,318 @@
...
+func (r *BaseDNSRecordReconciler) getActiveGroups(ctx context.Context, c client.Client, dnsRecord DNSRecordAccessor) types.Groups {
+	logger := log.FromContext(ctx).WithName("active-groups")
+	activeGroups := types.Groups{}
+	activeGroupsHost := activeGroupsTXTRecordName + "." + dnsRecord.GetZoneDomainName()
+
+	nameservers := r.getNameserversFromProvider(ctx, c, dnsRecord)
+	values, err := r.TXTResolver.LookupTXT(ctx, activeGroupsHost, nameservers)
+	if err != nil {
+		logger.Error(err, "error looking up active groups")
+		return activeGroups
```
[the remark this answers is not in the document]
philbrookes: in the case that this returns empty all groups would be activated.

internal/controller/dnsrecord_groups.go
```diff
@@ -0,0 +1,318 @@
...
+	for _, e := range allZoneEndpoints {
+		//not a TXT record type, skip it
+		if e.RecordType != endpoint.RecordTypeTXT {
+			continue
+		}
+		labels := make(map[string]string)
+		for _, target := range e.Targets {
+			var labelsFromTarget endpoint.Labels
+			_, _, labelsFromTarget, err = registry.NewLabelsFromString(target, []byte(txtRegistryEncryptAESKey))
+			if errors.Is(err, endpoint.ErrInvalidHeritage) {
```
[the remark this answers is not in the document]
philbrookes: good point, I think that's the only error it can ever be, but why limit the code like that.

internal/controller/dnsrecord_groups.go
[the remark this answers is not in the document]
Boomatang: I was trying to follow all the logic of the LookupTXT call, and I think it is making a network call to the provider. If my understanding is wrong please explain what is happening. 

With this logic if there is any blip in the communication with a third party provider we assume all groups are active. This seems like it could be very problematic. The edge case that comes to mind is: The current cluster is not in the list of active groups, but gets rate limited when trying to get the list of active groups. So now an error happens and the empty list is returned. With the empty list the cluster assumes there is no active groups and tries to reconcile all its records, but due to the initial rate limiting the problem gets compounded.

internal/controller/dnsrecord_groups.go
```diff
@@ -0,0 +1,318 @@
...
+
+	// if this record does not have a group, do not process unpublish
+	if dnsRecord.GetGroup() == "" {
+		return nil
+	}
+
+	activeGroups := r.getActiveGroups(ctx, c, dnsRecord)
+
+	// only process unpublish when there are active groups and we are reconciling a record from an active group
+	if len(activeGroups) == 0 || !dnsRecord.IsActive() {
```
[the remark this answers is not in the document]
Boomatang: I think this now can be simplified.
```suggestion
	if !dnsRecord.IsActive() {
```

internal/controller/dnsrecord_groups.go
```diff
@@ -0,0 +1,327 @@
...
+		return nil
+	}
+
+	if prematurely, _ := recordReceivedPrematurely(dnsRecord); prematurely {
+		logger.V(1).Info("Skipping DNSRecord - is still valid")
+		return nil
+	}
+
```
Boomatang: Can this check ever be true? I am looking at how the `recordReceivedPrematurely` is handling the is active.

https://github.com/philbrookes/dns-operator/blob/gh-620/internal/controller/dnsrecord_controller.go#L584-L587

philbrookes: good point, this does seem redundant now. I'll remove it

internal/controller/remote_dnsrecord_controller.go
```diff
@@ -229,6 +221,19 @@ func (r *RemoteDNSRecordReconciler) Reconcile(ctx context.Context, req mcreconci
 
 	dnsRecord.SetStatusCondition(string(v1alpha1.ConditionTypeReady), metav1.ConditionTrue, string(v1alpha1.ConditionReasonProviderSuccess), "Provider ensured the dns record")
 	dnsRecord.SetStatusObservedGeneration(dnsRecord.GetDNSRecord().GetGeneration())
+
+	logger.Info("unpublish section start")
+	// process unpublish of inactive groups once active cluster has no changes to publish
+	if !hadChanges && dnsRecord.GetGroup() != "" {
```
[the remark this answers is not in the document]
philbrookes: 1. The reason we read from the record here is because it is in the remote record reconciler.
2. answers below
  - Not necessarily, the remote may be ungrouped
  - Also no necessarily, local cluster could be group-1 and the remote group-2.
  - No, we'd expect an empty group (or absent)

internal/controller/dnsrecord_controller.go
```diff
@@ -178,6 +179,7 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		dnsRecord.SetStatusZoneDomainName("")
 		dnsRecord.SetStatusZoneID("")
 		dnsRecord.SetStatusDomainOwners(nil)
+		controllerutil.RemoveFinalizer(dnsRecord.GetDNSRecord(), DNSRecordFinalizer)
```
maksymvavilov: This doesn't seem right. With this change, we will not wait for the ` dnsRecord.GetStatus().ProviderEndpointsRemoved()` to be true

internal/controller/dnsrecord_groups.go
```diff
@@ -0,0 +1,479 @@
...
+				for g := range strings.SplitSeq(sections[1], "&&") {
+					group := types.Group(g)
+					if len(g) > 0 && !activeGroups.HasGroup(group) {
+						activeGroups = append(activeGroups, group)
+					}
+				}
+			}
+
+		}
+	}
```
maksymvavilov: This will work with all TXT records in the zone. If it happens to hold a random TXT record with `groups=cat` in it, this might cause a problem. I don't see that happening, to be honest, but it is a theoretical possibility. If we can't rely on the name of the TXT record here, why don't we take the route ExternalDNS [took](https://github.com/kubernetes-sigs/external-dns/blob/master/endpoint/labels.go#L60-L88) with the `heritage` prefix?  
I would not expect it to be done in this PR tho - as it is a "theoretical edge case that will, probably, never happen"

internal/controller/dnsrecord_groups.go
```diff
@@ -0,0 +1,479 @@
...
+//   - DNS providers host the active groups record on specific nameservers
+//   - You need to bypass public DNS for the active groups query
+//
+// The function looks for the NAMESERVERS data key in either:
+//  1. The provider secret referenced by dnsRecord.GetProviderRef(), OR
+//  2. The default provider secret (labeled kuadrant.io/default-provider=true)
+//
+// Returns a list of nameserver addresses (e.g., ["10.96.0.10:53", "10.96.0.11"])
+// Returns an empty list if no NAMESERVERS field is found or the secret doesn't exist.
+func (r *BaseDNSRecordReconciler) getNameserversFromProvider(ctx context.Context, c client.Client, dnsRecord DNSRecordAccessor) ([]string, error) {
```
maksymvavilov: These are the `baseReconciler` methods. Why aren't they in the `base_dnsrecord_reconciler.go`?

internal/controller/dnsrecord_groups.go
```diff
@@ -0,0 +1,479 @@
...
+		}
+		// no group, or active group, do not delete
+		if v, ok := labels[externaldnsregistry.GroupLabelKey]; !ok || activeGroups.HasGroup(types.Group(v)) {
+			continue
+		}
+		changes.Delete = append(changes.Delete, e)
+	}
+
+	if changes.HasChanges() {
+		err = dnsProvider.ApplyChanges(ctx, changes)
```
maksymvavilov: I've panicked here for a second. Why delete endpoints and TXTs in two separate calls? The first thing that comes to mind is that an error in the deletion of the TXTRecord will result in that TXT staying in the provider.

internal/controller/helper_test.go
```diff
@@ -84,3 +106,178 @@ func getTestHealthCheckSpec() *v1alpha1.HealthCheckSpec {
...
+		labels = map[string]string{}
+	}
+	labels[v1alpha1.DefaultProviderSecretLabel] = "true"
+	secret.SetLabels(labels)
+	Expect(k8sClient.Create(ctx, secret)).To(Succeed())
+	return secret
+}
+
+// createWeightedCNAMEDNSRecord creates a delegated DNSRecord with weighted CNAME routing
+func createWeightedCNAMEDNSRecord(name, namespace, hostname, clusterTarget string, weight int) *v1alpha1.DNSRecord {
```
maksymvavilov: We will now have two ways we create records for the tests. [With](https://github.com/philbrookes/dns-operator/blob/gh-620/internal/controller/dnsrecord_controller_test.go#L74-L88) the `TestEndpoints` from the helper and using this method to get a record itself. Don't know if this matters, but would prefer us to be consistent.

internal/controller/remote_dnsrecord_controller.go
```diff
@@ -229,6 +241,17 @@ func (r *RemoteDNSRecordReconciler) Reconcile(ctx context.Context, req mcreconci
 
 	dnsRecord.SetStatusCondition(string(v1alpha1.ConditionTypeReady), metav1.ConditionTrue, string(v1alpha1.ConditionReasonProviderSuccess), "Provider ensured the dns record")
 	dnsRecord.SetStatusObservedGeneration(dnsRecord.GetDNSRecord().GetGeneration())
+
+	logger.Info("unpublish section start")
```
maksymvavilov: This feels like a debug call that was left behind

internal/external-dns/registry/txt.go
```diff
@@ -39,6 +42,8 @@ const (
 
 	txtFormatVersion         = "1"
 	externalDNSMapperVersion = ""
+
+	GroupLabelKey = "group"
```
maksymvavilov: We [have](https://github.com/philbrookes/dns-operator/blob/gh-620/types/group.go#L10) this key in the `types/group.go`

internal/controller/dnsrecord_groups.go
```diff
@@ -0,0 +1,479 @@
...
+		logger.Error(err, "error getting custom nameservers from provider")
+		return activeGroups
+	}
+	values, err := r.TXTResolver.LookupTXT(ctx, activeGroupsHost, nameservers)
+	if err != nil {
+		logger.Error(err, "error looking up active groups")
+		return activeGroups
+	}
+	// Parse the TXT record to extract active groups
+	// Expected format: "groups=group1&&group2&&group3;version=1;other=value"
```
maksymvavilov: I expected most parts of this format to reside in [constants](https://github.com/philbrookes/dns-operator/blob/gh-620/cmd/plugin/failover/shared.go#L20-L24) and be reused in places we want to reuse them.

internal/controller/dnsrecord_controller.go
philbrookes: yes, you're right. I've already undone this locally but not pushed it up yet.

internal/controller/dnsrecord_accessor.go
```diff
@@ -41,17 +41,21 @@ type DNSRecordAccessor interface {
 	GetStatus() *v1alpha1.DNSRecordStatus
 	SetStatusConditions(hadChanges bool)
 	SetStatusCondition(conditionType string, status metav1.ConditionStatus, reason, message string)
+	ClearStatusCondition(conditionType string)
+	GetStatusCondition(conditionType string) *metav1.Condition
```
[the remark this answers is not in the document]
philbrookes: It just felt like a natural method to accompany the SetStatusCondition which also wraps meta.SetStatusCondition.

internal/controller/dnsrecord_groups.go
philbrookes: I thought it made more sense to keep all the group related logic in the one file. Like we have done with the healthchecks.

internal/controller/dnsrecord_groups.go
philbrookes: That would still be picked up in a following reconcile. It's done in stages intentionally to ensure the cleanup of DNS is done before cleaning out the metadata required to work out what to delete. 

If the metadata was deleted and the records not (which would be possible if we did these changes all in one plan, thanks to Azures lack of batching), it would be impossible to recover from that later on.

internal/controller/suite_test.go
```diff
@@ -300,6 +347,7 @@ func setupEnv(delegationRole string, count int) (*envtest.Environment, ctrl.Mana
 			Scheme:          mgr.GetScheme(),
 			ProviderFactory: providerFactory,
 			DelegationRole:  delegationRole,
+			TXTResolver:     &MockTXTResolver{},
```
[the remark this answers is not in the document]
philbrookes: yes, exactly. This allows us to mock a TXT lookup response without waiting for propagation.

internal/controller/base_dnsrecord_reconciler.go
```diff
@@ -192,7 +195,15 @@ func (r *BaseDNSRecordReconciler) applyChanges(ctx context.Context, dnsRecord DN
 }
 
 func (r *BaseDNSRecordReconciler) updateStatus(ctx context.Context, client client.Client, previous, current DNSRecordAccessor, err error) (reconcile.Result, error) {
-	result, uErr := r.updateStatusAndRequeue(ctx, client, previous, current, 0)
+	_, requeueTime := recordReceivedPrematurely(current)
+	if !current.IsActive() {
+		requeueTime = InactiveGroupRequeueTime
+	}
+	if current.IsDeleting() {
+		requeueTime = time.Millisecond * 100
```
maksymvavilov: This feels arbitrary. 
Given that we will be changing the whole requeue logic, would it be a good idea to stay consistent with [other](https://github.com/philbrookes/dns-operator/blob/gh-620/internal/controller/dnsrecord_controller.go#L61-L70) requeue times? 
Or at least make it a constant somewhere around that place

internal/controller/dnsrecord_groups.go
maksymvavilov: Thanks for the extra comment in the code. With my goldfish memory, you saved me from a few heart attacks later on

maksymvavilov (dismissed):
Looks good. Two comments that ask for attention: adding a test case and making a value in a constant. Neither of them is a reason enough to hold the merge

internal/controller/helper_test.go
maksymvavilov: What matters for those tests are endpoints. And as loong as we have only one place/way of doing them all should be good

config/local-setup/dns-provider/coredns/kustomization.yaml
```diff
@@ -5,6 +5,7 @@ generatorOptions:
   labels:
     app.kubernetes.io/part-of: dns-operator
     app.kubernetes.io/managed-by: kustomize
+    kuadrant.io/default-provider: "true"
```
mikenairn: I don't think we should making any of these secrets created by local-setup a default. Presumably this is only this way to make local testing easier for you? One less step (labelling the secret)? Note, you can still just set the coredns provider secret directly in the record providerRef like any other if you know you want to use it.

internal/controller/base_dnsrecord_reconciler.go
```diff
@@ -106,7 +107,7 @@ func (r *BaseDNSRecordReconciler) publishRecord(ctx context.Context, dnsRecord D
 	if err != nil {
 		return hadChanges, err
 	}
-	logger.Info("Published DNSRecord to zone")
+	logger.Info("Published DNSRecord to zone", "hadChanges?", hadChanges)
```
mikenairn: nit: Usually we just use the variable name for the key i.e. `hadChanges` not `hadChanges?`

internal/controller/dnsrecord_controller.go
```diff
@@ -341,9 +362,31 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+		}
+	}
+
 	return r.updateStatus(ctx, previous, dnsRecord, hadChanges, nil)
 }
 
+func (r *DNSRecordReconciler) applyGroupAdapter(ctx context.Context, dnsRecord DNSRecordAccessor) DNSRecordAccessor {
+	if dnsRecord.GetDNSRecord().IsAuthoritativeRecord() || r.Group == "" {
+		return dnsRecord
+	}
```
mikenairn: If you make `dnsRecord.SetStatusGroup(r.Group)` ensure the group is set correctly (Only has a group when IsAuthoritativeRecord() is false ) then you won't need to do this and can just adapt it based on the status of the record (dnsRecord.GetGroup()) .

config/local-setup/dns-provider/coredns/kustomization.yaml
philbrookes: agreed, I was just speeding up my local-dev.

coredns/examples/Corefile
```diff
@@ -1,11 +1,22 @@
...
-     to *
-   }
-   kuadrant
-}
+    debug
+    errors
+    log
+    
+    rewrite name regex kuadrant-active-groups\.(.*)k.example\.com kuadrant-active-groups-coredns.pb.hcpapps.net
+    forward kuadrant-active-groups-coredns.pb.hcpapps.net /etc/resolv.conf
```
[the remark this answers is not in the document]
philbrookes: I have a configurable option created for this now, that I was going to do in a separate PR, I'll add it to this one.

README.md
```diff
@@ -56,24 +56,53 @@ az network dns zone show --name <my domain name> --resource-group <my resource g
...
+
+Once the external record has been created, find the local IP of the coredns instance:
+```sh
+NS=$(kubectl get secrets -n dnstest dns-provider-credentials-coredns -o yaml | yq ".data.NAMESERVERS" | base64 -d | cut -f1 -d":")
+```
+Then dig the local TXT record from that IP:
+```sh
+dig @$NS kuadrant-active-groups.k.example.com TXT +short
+"version=1;groups=group-2"
+```
```
mikenairn: Presumably this is being read from a TXT record that expected to be retrievable from  `EXTERNAL_GROUPS_HOST`. It's not clear here where `version=1;groups=group-2` is coming from, mayb e you should add a part that explain that? What the expected contents of the TXT record are supposed to be?

config/coredns/.gitignore
```diff
@@ -1 +1,2 @@
 charts
+Corefile
```
mikenairn: ?

internal/controller/dnsrecord_controller.go
```diff
@@ -341,9 +360,33 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+
+	// Do not look up active groups or set status if this reconciler has no group set
+	if r.Group != "" {
+		activeGroups = r.getActiveGroups(ctx, r.Client, dnsRecord)
+		dnsRecord.SetStatusActiveGroups(activeGroups)
+	}
+
+	dnsRecord = newGroupAdapter(dnsRecord, activeGroups)
+
+	return dnsRecord
```
mikenairn: I think i know why we are doing it (removal of old group status), but i do wonder if the group adapter shouldn't be applied all the time but only when a group is set.

internal/controller/remote_dnsrecord_controller.go
```diff
@@ -229,7 +241,18 @@ func (r *RemoteDNSRecordReconciler) Reconcile(ctx context.Context, req mcreconci
 
 	dnsRecord.SetStatusCondition(string(v1alpha1.ConditionTypeReady), metav1.ConditionTrue, string(v1alpha1.ConditionReasonProviderSuccess), "Provider ensured the dns record")
 	dnsRecord.SetStatusObservedGeneration(dnsRecord.GetDNSRecord().GetGeneration())
-	return r.updateStatus(ctx, cl.GetClient(), previous, dnsRecord, nil)
+
+	// process unpublish of inactive groups once active cluster has no changes to publish
+	if !hadChanges && dnsRecord.IsActive() && dnsRecord.GetGroup() != "" {
```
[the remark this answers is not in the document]
philbrookes: Testing for GetGroup() != "" is still required though, to ensure ungrouped cluster don't process unpublishing.

internal/controller/dnsrecord_controller.go
philbrookes: well, "easier"... maybe better to say: less paths through the logic.

internal/controller/remote_dnsrecord_controller.go
```diff
@@ -218,8 +218,20 @@ func (r *RemoteDNSRecordReconciler) Reconcile(ctx context.Context, req mcreconci
...
+	activeGroups := r.getActiveGroups(ctx, r.mgr.GetLocalManager().GetClient(), dnsRecord)
+	if dnsRecord.GetGroup() != "" {
+		dnsRecord.SetStatusActiveGroups(activeGroups)
+		dnsRecord = newGroupAdapter(dnsRecord, activeGroups)
+	}
+
+	if !dnsRecord.IsActive() {
+		logger.V(1).Info("remote record is from an inactive group, exiting reconcile")
+		_, err := r.updateStatus(ctx, r.mgr.GetLocalManager().GetClient(), previous, dnsRecord, nil)
+		return reconcile.Result{RequeueAfter: InactiveGroupRequeueTime}, err
```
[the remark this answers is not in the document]
mikenairn: This is still the wrong client. Do we have tests that cover this scenario? What is the status supposed to look like here? Have we updated it?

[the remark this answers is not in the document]
philbrookes: removed.

internal/controller/dnsrecord_groups.go
```diff
@@ -0,0 +1,484 @@
...
+func (r *BaseDNSRecordReconciler) unpublishInactiveGroups(ctx context.Context, c client.Client, dnsRecord DNSRecordAccessor, dnsProvider provider.Provider) error {
+	logger := log.FromContext(ctx).WithName("active-groups").WithName("unpublish")
+	managedDNSRecordTypes := []string{externaldnsendpoint.RecordTypeA, externaldnsendpoint.RecordTypeAAAA, externaldnsendpoint.RecordTypeCNAME}
+
+	// if this record does not have a group, do not process unpublish
+	if dnsRecord.GetGroup() == "" {
+		return nil
+	}
+
+	activeGroups := r.getActiveGroups(ctx, c, dnsRecord)
```
mikenairn: This is called when the group adapter is setup and in here, this could potentially lead to different results? Do we really need to get them again? Can it not just use the active groups set on the record being passed in?

internal/controller/remote_dnsrecord_controller.go
```diff
@@ -218,8 +218,15 @@ func (r *RemoteDNSRecordReconciler) Reconcile(ctx context.Context, req mcreconci
 		return r.updateStatus(ctx, cl.GetClient(), previous, dnsRecord, err)
 	}
 
+	dnsRecord = r.applyGroupAdapter(ctx, r.mgr.GetLocalManager().GetClient(), dnsRecord)
+
+	if !dnsRecord.IsActive() {
+		logger.V(1).Info("remote record is from an inactive group, exiting reconcile")
+		return reconcile.Result{RequeueAfter: InactiveGroupRequeueTime}, nil
+	}
```
mikenairn: Why is not behaving the same as the local record reconciler which is setting the "InInactiveGroup" condition? Seems to me they should both do the same thing no?

mikenairn (commented):
Left a few comments, think this is close, but couple of things could be updated before merge.

Some other things have been discussed around improvements that could/should be made before GA, but are acceptable for dev preview.

Claude generated summary of dev preview status which may be of interest:
https://gist.github.com/mikenairn/82724d392211e848d5c6fa9732e0bc65

internal/controller/dnsrecord_groups.go
mikenairn: Highlighted as something to investigate after merge and pre GA.

PR #673, opened 2025-12-10
add output formatter

maksymvavilov:
adding text formatter and aligning logger levels with the zap implementation


cmd/plugin/failover/get-active-groups.go
```diff
@@ -121,14 +122,14 @@ func getActiveGroups(_ *cobra.Command, _ []string) error {
 	}
 
 	if len(allActiveGroups) == 0 {
-		log.Info(fmt.Sprintf("No active groups found for domain %s", domain))
+		output.Formatter.Print(fmt.Sprintf("No active groups found for domain %s", domain))
```
philbrookes: stuff like this is fine for print, though.

philbrookes (changes requested):

cmd/plugin/output/formatter.go
```diff
@@ -0,0 +1,33 @@
+package output
+
+import (
+	externaldns "sigs.k8s.io/external-dns/endpoint"
+)
+
+type OutputFormatter interface {
+	Print(message string)
+	Error(err error, message string)
+	RenderEndpoints(endpoints []*externaldns.Endpoint)
```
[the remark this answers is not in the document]
maksymvavilov: I've marked all of them as resolved - latest changes address those. 
For transparency, we had discussions in Zulip and on the call about the requested changes.

cmd/plugin/output/logging.go
```diff
@@ -0,0 +1,50 @@
...
+	case WarnLevel:
+		loggerLevel = zapcore.WarnLevel
+	case ErrorLevel:
+		loggerLevel = zapcore.ErrorLevel
+	case PanicLevel:
+		loggerLevel = zapcore.PanicLevel
+	default:
+		loggerLevel = zapcore.InfoLevel
+	}
+	return zap.New(zap.UseDevMode(false), zap.WriteTo(os.Stderr), zap.Level(loggerLevel))
```
maksymvavilov: I'm setting `devMode` here to false. It will be less human-readable this way. The DevMode uses a different config, and it invalidates any log levels passed in (it always will act as if it is in a debug mode)

cmd/plugin/common/pretty.go
```diff
@@ -1,36 +0,0 @@
...
-import (
-	"os"
-	"strings"
-
-	"github.com/jedib0t/go-pretty/v6/table"
-
-	externaldns "sigs.k8s.io/external-dns/endpoint"
-)
-
-func RenderEndpoints(endpoints []*externaldns.Endpoint) {
```
maksymvavilov: The `RenderEndpoints` became a more generic `PrintTable` in the formatter.

Issue #600, opened 2025-10-08
move dns operator metrics to collector pattern


philbrookes:
We had some of the [metrics](https://github.com/Kuadrant/dns-operator/blob/main/internal/metrics/metrics.go) moved to a Collector pattern. 
We should consider moving the rest of them. 
Also, consider how long the request would take with all the metrics using this pattern.

PR #676, opened 2025-12-12
Refactor metrics to the collector pattern

Boomatang:
Closes: https://github.com/Kuadrant/dns-operator/issues/600

Move the existing metrics to the collector pattern.

Removed the dead code around `SecretMissing` metric.

### DIFF
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -10,7 +10,6 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
-	"sigs.k8s.io/controller-runtime/pkg/metrics"
 	kubeconfigprovider "sigs.k8s.io/multicluster-runtime/providers/kubeconfig"
 
 	"github.com/kuadrant/dns-operator/api/v1alpha1"
@@ -25,9 +24,6 @@ const (
 	dnsHealthCheckNameLabel      = "dns_health_check_name"
 	dnsHealthCheckNamespaceLabel = "dns_health_check_namespace"
 	dnsHealthCheckHostLabel      = "dns_health_check_host"
-	mzRecordNameLabel            = "managed_zone_name"
-	mzRecordNamespaceLabel       = "managed_zone_namespace"
-	mzSecretNameLabel            = "managed_zone_secret_name"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


internal/controller/dnsrecord_controller.go
```diff
@@ -391,9 +381,9 @@ func (r *DNSRecordReconciler) updateStatus(ctx context.Context, previous, curren
 		// generation has not changed but there are changes.
 		// implies that they were overridden - bump write counter
 		requeueTime = randomizedValidationRequeue
+		// BUG: https://github.com/Kuadrant/dns-operator/issues/664 The premature checks stops this from running.
```
maksymvavilov: I don't think it is a bug. 
If we are here, it means we did write to the provider on the same generation. Premature logic, indeed, prevents us from getting here, but we need this specifically for the moments when premature logic allows us to do a full loop.

maksymvavilov (commented):
Looks good. Only one question about the pattern and a comment about a reference to the bug

internal/controller/dnsrecord_controller.go
Boomatang: The problem is the premature logic stops this reconciling if the status has being updated. When the generation changes the metric gets reset back to zero. 

That issue is also marked as a bug.

internal/metrics/metrics.go
```diff
@@ -121,6 +137,135 @@ func (c *RemoteClusterCollector) Collect(ch chan<- prometheus.Metric) {
...
+func (c *RemoteClusterCollector) RemoveRemoteRecordReconcile(cluster, namespace, name string) {
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
+
+	key := cluster + ":" + namespace + ":" + name
+	delete(c.remoteReconcileCounters, key)
+}
+
+// SetRemoteRecordsCount sets the count of delegated remote records for a specific cluster
+func (c *RemoteClusterCollector) SetRemoteRecordsCount(cluster string, count float64) {
```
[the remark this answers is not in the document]
Boomatang: I see your reasoning. 

The collector does not publish the metric till it is asked for. It can still collect the data that is need to publish before hand. In this case I looked at the cost of collecting that data. If this logic was to only happen at metric collection it would mean making a request to every remote cluster to get a list of dns records. While there would the cache would help I think in general that would be a very expensive request which would slow the metric collection.

On the probe counter. I don't like the idea of counting the number of defined in the probe CRs. That metric would be this is how many probe could be running, but not actually how many are running.

PR #687, opened 2026-01-12
add in pprof option for the controller

maleck13:
will expose the pprof http server endpoints at :8082 by default.

rh-pre-commit.version: 2.2.0
rh-pre-commit.check-secrets: DISABLED

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -69,6 +69,7 @@ var (
 	metricsAddr            string
 	enableLeaderElection   bool
 	probeAddr              string
+	pprofAddr              string
 	minRequeueTime         time.Duration
 	validFor               time.Duration
 	maxRequeueTime         time.Duration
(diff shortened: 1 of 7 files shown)


config/manager/metrics_service.yaml
```diff
@@ -8,8 +8,11 @@ metadata:
...
   ports:
-  - name: metrics
-    port: 8080
-    targetPort: metrics
+    - name: metrics
+      port: 8080
+      targetPort: metrics
+    - name: pprof
+      port: 8082
+      targetPort: pprof
```
mikenairn: Should this be included in the metrics service or have it's own? Presumably this should not be exposed externally, where metrics could be?

maleck13: This should be fine. It is observability. It is on a different port than metrics endpoints, so to expose it externally you would need a HTTPRoute pointing at the metrics backend with a target of the pprof port.

cmd/main.go
```diff
@@ -169,6 +172,13 @@ func main() {
...
 		LeaderElection:         enableLeaderElection,
 		LeaderElectionID:       "a3f98d6c.kuadrant.io",
+		// make pprof info available at /debug/pprof/${type} where type is one of (goroutine, heap, threadcreate block mutex, profile, trace)
+		// https://pkg.go.dev/runtime/pprof
+		PprofBindAddress: pprofAddr,
+	}
+
+	if pprofAddr != "" {
+		setupLog.Info("Pprof configured", "listening", pprofAddr)
 	}
```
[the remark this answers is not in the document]
maleck13: This doesn't turn on profiling. It just makes the profiling endpoints available to use. I think having them available as long as not publicly is useful rather than having to restart the server and have it be in potentially a different state?

Issue #686, opened 2026-01-09
Review resource limits on DNS -Operator


philbrookes:
In the case where there are an enormous number of secrets / configmaps on a cluster, the DNS Operator will spike in memory consumption when restarted due to caching all secrets on the cluster at startup and result in a loop where it is OOMKilled repeatedly.

We need to investigate permitting much higher limits, to avoid this for now, until a more permanent solution can be offered to avoid the memory spike in all scenarios.

Related Jira: https://issues.redhat.com/browse/CONNLINK-678#

Boomatang:
## Current understanding

After the investigation that has being done, bumping the limits on the operator is band-aid, and not the root cause of the issue.

The root cause of the issue is how the caching layer works in the controller-runtime. Once a watch is configured a reconciler, the controller-runtime opens a connection with the api server, and starts caching all the resources from the cluster. If the watch on the reconciler has a predicate, this will only filter the resource that can trigger the reconciler. It DOES NOT affect what resources the controller-runtime will cache. 

As the dns-operator runs cluster scope this means it will cache resources from all namespaces. We are affected by the secrets, even tho the reconciler watch filters them by type. The controller-runtime will cache every secret on the cluster.

It get even worse that this for how the controller-runtime is doing its caching. But luckily the dns-operator is not directly affected with this issue, and there are steps to stop us becoming accidentally affect in the future. If the a reconcile function does a `get` or `list` on a resource in cluster which had no watch previously configured, the controller-runtime sets up a informer to be notified of all changes to resource type on the cluster. The limitador-operator is affected by this once a user configures persisted storage.

## What can we do

For the dns-operator there is two caching object we need to protect against. 
- Resources we watch
- Resources that in the future maybe requested from the cluster.

To guard against resources that may be requested from the cluster in the future, the `ReaderFailOnMissingInformer: true` flag can be set in the cache options. This will cause an error when a `get` or `list` is called on any resource that does not directly have a watch configured. 

For resources we watch the solution is simple, but the implications are far reaching. We configure the cache to only cache resources that have a known label applied to them. There is two way this can be done, 1) at a default level, or 2) at a resource level. There is a few options when it comes to filtering by label values, but the key needs to stay the same.
1) an default label can be filter on all resource. This would require every resource required by the operator to have at least the same label key. 
2) but resource object. It is possible to target a resource type. In the dns-operator case the secrets could targeted and require a label. The controller-runtime would filter the cache for the secrets, but not on any other resource.

[...]

PR #696, opened 2026-01-15
UPDATE: adjust controller-runtime cache

Boomatang:
Closes: #686 

By changing how the controller-runtime is configured the memory usage of the operator can be greatly reduced.

The use of the multicluster-runtime does introduce some nuances as that creates watches that also need to be handled.

## Verifying

**warning:** These steps assumes you only have one kind cluster that was created during these steps, and didn't change default names.

1. From main create a reference cluster using `make local-setup DEPLOY=true`.
2. Wait for operator to be come ready.
3. In a second terminal set up monitoring on the deployment. This one liner can be used. Change the CONTAINER_TOOL if you are not using podman.
```bash
watch -n 1 'CONTAINER_TOOL=podman; CID=$(kubectl get pods -n dns-operator-system --field-selector=status.phase=Running --no-headers -o custom-columns=":status.containerStatuses[0].containerID" | head -1 | sed "s/.*:\/\///" | cut -c1-12); BYTES=$($CONTAINER_TOOL exec kuadrant-dns-local-1-control-plane sh -c "cat \$(find /sys/fs/cgroup -name memory.current -path "*"$CID"*" 2>/dev/null | head -1)"); awk "BEGIN {printf \"%.1f MiB\n\", $BYTES/1024/1024}"'
```
4. Back in the initial terminal create a testing namespace.
```sh
kubectl create namespace testing
```
5. Add a large number of secrets to the testing namespace.
```sh
 seq 1 12000 | xargs -P 20 -I {} sh -c 'kubectl create secret generic test-secret-$(cat /proc/sys/kernel/random/uuid | cut -c1-8) -n testing --from-literal=data=$(tr -dc a-zA-Z0-9</dev/urandom | head -c 5000) 2>/dev/null'
 ```

6. During the adding of the secrets the memory usage of the operator will raise. If the adding of secrets finish without crashing the operator, restart the operator. It will crash on restart.
7. Check out this breach, and rebuild the operator image.
```sh
make docker-build && kind get clusters | xargs -I {} bash -c 'make kind-load-image KIND_CLUSTER_NAME={} && kubectl -n dns-operator-system rollout restart deployment dns-operator-controller-manager'
```
8. Watching the monitoring terminal, when the new deployment starts the memory should become stable. In my testing this was around 12~14 MiB. No number of restarts will cause the operator to crash due to memory spikes.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -29,12 +29,15 @@ import (
 
 	"go.uber.org/zap/zapcore"
 
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes/scheme"
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	zap "sigs.k8s.io/controller-runtime/pkg/log/zap"
 	"sigs.k8s.io/controller-runtime/pkg/metrics"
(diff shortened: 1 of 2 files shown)


philbrookes (approved):

Boomatang:
@mikenairn yes, I totally agree. If we want to get back to the immediate reactive behavior the secrets we require would need to have a common label across all the secrets that we depend on. That would include the multi cluster kubeconfig cluster. 

There is two main points that would need to be address.
1. How we handle the migration of existing clusters. As we would not be able to fetch secrets that are not labelled correctly. 
2. The label used should be standardize across the whole of kuadrant, or at least the format. If you have thoughts on that format please add it to https://github.com/Kuadrant/kuadrant-operator/issues/1731

PR #712, opened 2026-01-29
improve doc-verification skill

maksymvavilov:
Improving on the doc-verification skill 

During the wotrk on the CoreDNS docs i needed to "train" Claude a lot to generate and review docs. I have him update the skill with "everything you learned". This seems to work better :shrug:


philbrookes:
Do these files get picked up by Claude automatically, or how are they activated?

maksymvavilov:
@philbrookes 

> Do these files get picked up by Claude automatically, or how are they activated?

Once you start Claude it loads `Skills` and uses them from the context 
image
 [This](https://github.com/maksymvavilov/dns-operator/blob/claude-improvements/.claude/skills/doc-verification/SKILL.md?plain=1#L3) line tells claude when to use it

Issue #694, opened 2026-01-14
Document migrating away from DNS Groups


philbrookes:
How to remove groups from all controllers:

- Remove the group value from all envvar configmaps
- Restart controllers
- Confirm DNS Record statuses 
  - No Group or activeGroups entries
  - No Active condition
- Confirm zone as expected
- Delete active-groups TXT record

PR #718, opened 2026-02-04
Add documentation for migrating away from DNS Groups

Boomatang:
Closes: #694

- Add step-by-step guide for removing DNS Groups configuration from dns-operator deployments
- Document verification steps to confirm successful migration
- Include optional cleanup instructions for active-groups TXT records

### DIFF
--- a/docs/migrating_away_from_dns_groups.md
+++ b/docs/migrating_away_from_dns_groups.md
@@ -1,0 +1,60 @@
+# Migrating away from DNS Groups
+
+## How
+
+### Remove the group value from the operator
+Modify the `dns-operator-controller-env` configmap in the dns-operator namespace and remove the `GROUP` entry.
+
+If the dns-operator is configured to use a group but the configmap does not have the group configured, the configuration was added directly to the dns-operator deployment.
+Edit the deployment.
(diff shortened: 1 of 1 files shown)


docs/migrating_away_from_DNS_groups.md
```diff
@@ -0,0 +1,60 @@
+# Migrating away from DNS Groups
```
mikenairn: Maybe some overview of what this guide is trying to achieve would be useful?

docs/migrating_away_from_DNS_groups.md
mikenairn: I'd avoid capitals in the file names:
migrating_away_from_DNS_groups.md -> migrating_away_from_dns_groups.md

Doubt it actually causes any issues, just looks odd to me and we don't do it for any other file that i can see.

We seem to have a mix of - and _  in file names so i guess that's fine, but probably would be nice if we went with one over the other generally.

philbrookes:
This PR looks like it has outstanding comments, so I'll move it to "In Progress".

mikenairn (approved):
Looks fine to me.

PR #720, opened 2026-02-05
Add version back to release.mk

Boomatang:
Version field was removed on mistake from the prepare-release command,
causing tag image builds to fail.


Boomatang:
@mikenairn Here is that PR we spoke  about.

philbrookes (approved):

PR #731, opened 2026-02-16
fix: correct endpoint provider name in readme variable

averevki:
Change endpoint provider from `endpoints` to `endpoint` in the readme. Controller won't start with endpoint**s** provider


mikenairn:
Thanks @averevki. Rolled this into the larger docs change https://github.com/Kuadrant/dns-operator/pull/711.

PR #762, opened 2026-04-19
internal/provider: gate the TLD early-exit on the ICANN flag

### LINKED ISSUE #761: Zone matching fails for domains in the public suffix list
## Summary

`FindDNSZoneForHost()` in `internal/provider/provider.go` rejects any host whose domain is listed in the [public suffix list](https://publicsuffix.org/), even when a matching zone is explicitly configured in the provider secret.

## Root Cause

Lines 150–155 of `internal/provider/provider.go`:

```go
tld, _ := publicsuffix.PublicSuffix(host)
if host == tld {
    return nil, "", fmt.Errorf("%w: %s", ErrNoZoneForHost, originalHost)
}
```

`golang.org/x/net/publicsuffix.PublicSuffix()` returns the **public suffix** of a domain, not just the ICANN TLD. Many real-world domains appear in the public suffix list as privately managed domains (e.g., `httpbin.org`, `github.io`, `amazonaws.com`, `azurewebsites.net`). For these domains, `PublicSuffix(host)` returns the host itself, causing the early-exit check `host == tld` to fire before zone matching is attempted.

Demonstration:

```go
ps, icann := publicsuffix.PublicSuffix("httpbin.org")
// ps = "httpbin.org", icann = false, host == ps → true → rejected

ps, icann = publicsuffix.PublicSuffix("example.com")
// ps = "com", icann = true, host == ps → false → zone matching proceeds
```

## Impact

When using the CoreDNS provider with `ZONES=httpbin.org` and a DNSRecord with `rootHost: httpbin.org`, the authoritative record fails reconciliation:

```
error: "no zone for host: httpbin.org"
```

The individual DNSRecord succeeds (via the endpoint/delegation provider path), but the authoritative record — which goes through the CoreDNS provider's `DNSZoneForHost()` directly — hits the public suffix check and fails. This prevents the `kuadrant.io/coredns-zone-name` label from being added, so the CoreDNS plugin never discovers the record.

This affects any domain in the public suffix list used as both the rootHost and the zone. Subdomains work fine (e.g., `simple.k.example.com` with zone `k.example.com`).

## Reproduction

```bash
# Create provider secret with the domain as the zone
kubectl create secret generic dns-provider-credentials-coredns \
  --namespace=dnstest \
  --type=kuadrant.io/coredns \
  --from-literal=ZONES="httpbin.org"

# Create DNSRecord
kubectl apply -f - <<EOF
kind: DNSRecord
metadata:
  name: httpbin-test
  namespace: dnstest
spec:
  rootHost: httpbin.org
  providerRef:
    name: dns-provider-credentials-coredns
  endpoints:
  - dnsName: httpbin.org
    recordType: A
    recordTTL: 60
    targets:
    - "10.89.0.17"
EOF

# Wait and check — authoritative record will show DNSProviderError
kubectl get dnsrecord -n dnstest
```

## Suggested Fix

The public suffix check was added to prevent recursion past a useful boundary, but it's too aggressive for privately managed public suffixes. Consider:

1. Only checking `icann == true` public suffixes (skip privately managed ones)
2. Checking the zone list before applying the TLD rejection — if a zone explicitly matches the host, allow it
3. Removing the early exit entirely and relying on the recursive subdomain stripping to terminate naturally

Option 2 seems safest — the user has explicitly configured the zone, so it should be respected.

## Context

Discovered while investigating DNS-based egress gateway routing for [Kuadrant/kuadrant-operator#1847](https://github.com/Kuadrant/kuadrant-operator/issues/1847). The use case is mapping external hostnames to an egress gateway's ClusterIP via kuadrant CoreDNS + DNSRecord, where the rootHost IS the external hostname (e.g., `httpbin.org`).

SAY-5:
## What

`findDNSZoneForHost` in `internal/provider/provider.go` asks `publicsuffix.PublicSuffix(host)` for the "TLD" and then bails out of zone matching when the host equals that value:

```go
tld, _ := publicsuffix.PublicSuffix(host)

// The host is a TLD, so we now know `originalHost` can't possibly have a valid `DNSZone` available.
if host == tld {
    return nil, "", fmt.Errorf("%w: %s", ErrNoZoneForHost, originalHost)
}
```

The Go public suffix list is not just the ICANN TLD set — it also contains privately-managed domains like `httpbin.org`, `github.io`, `amazonaws.com`, `azurewebsites.net`. For those, `PublicSuffix` returns the host itself and the *second* return value (`icann`) is `false`, so the early-exit fires before zone matching is attempted.

Concretely, with

```yaml
ZONES=httpbin.org

# DNSRecord
rootHost: httpbin.org
```

the CoreDNS provider path rejects the authoritative record with `no zone for host: httpbin.org`, the `kuadrant.io/coredns-zone-name` label is never added, and the CoreDNS plugin never discovers the record. Subdomains like `simple.k.example.com` are unaffected because `PublicSuffix("simple.k.example.com")` returns `com`, which doesn't equal the host, so zone matching proceeds.

Per #761.

## Fix

Read the `icann` return value the check actually meant to rely on, and only bail out when the host is an ICANN-managed TLD:

```go
tld, icann := publicsuffix.PublicSuffix(host)
if icann && host == tld {
    return nil, "", fmt.Errorf("%w: %s", ErrNoZoneForHost, originalHost)
}
```

- `example.com` → `tld="com"`, `icann=true`, `host != tld` → zone matching proceeds (unchanged).
- `com` → `tld="com"`, `icann=true`, `host == tld` → reject with `ErrNoZoneForHost` (unchanged).
- `httpbin.org` → `tld="httpbin.org"`, `icann=false`, check skipped → zone matching proceeds, the CoreDNS zone in the provider secret matches, authoritative record succeeds (previously rejected).

## Scope

One-line logic change; no new imports (`icann` is already returned by `publicsuffix.PublicSuffix`, we were just throwing it away). No DNSRecord that currently resolves stops resolving.

Fixes #761

## Summary by CodeRabbit

* **Bug Fixes**
  * DNS zone resolution now recognises privately managed public suffixes (e.g. subdomains under private suffixes), avoiding unnecessary "no zone" failures while still rejecting disallowed ICANN apex hosts.
* **Tests**
  * Expanded tests to cover private-suffix resolution, apex-deny behaviour, and a regression case ensuring ICANN suffix rejection remains enforced.

### DIFF
--- a/internal/provider/provider.go
+++ b/internal/provider/provider.go
@@ -146,14 +146,8 @@ func findDNSZoneForHost(originalHost, host string, zones []DNSZone, denyApex boo
 		return nil, "", fmt.Errorf("%w: %s", ErrNoZoneForHost, host)
 	}
 	host = strings.ToLower(host)
-	// Bail out only if the host equals an ICANN TLD (e.g. `com`, `org`). The
-	// Go public suffix list also contains privately-managed domains such as
-	// `httpbin.org`, `github.io`, `amazonaws.com`, `azurewebsites.net` - for
-	// those, PublicSuffix returns the host itself but the second return
-	// value (`icann`) is false. Ignoring that flag rejected any DNSRecord
-	// whose rootHost happened to appear in the private suffix list, even
-	// when a matching zone was explicitly provisioned in the provider
-	// secret.
+	// Only reject if the host is an ICANN TLD. Private suffixes (e.g.
+	// `s3.amazonaws.com`, `github.io`) should proceed to zone matching.
 	tld, icann := publicsuffix.PublicSuffix(host)
 	if icann && host == tld {
 		return nil, "", fmt.Errorf("%w: %s", ErrNoZoneForHost, originalHost)
--- a/internal/provider/provider.go
+++ b/internal/provider/provider.go
@@ -146,11 +146,16 @@ func findDNSZoneForHost(originalHost, host string, zones []DNSZone, denyApex boo
 		return nil, "", fmt.Errorf("%w: %s", ErrNoZoneForHost, host)
 	}
 	host = strings.ToLower(host)
-	//get the TLD from this host
-	tld, _ := publicsuffix.PublicSuffix(host)
-
-	//The host is a TLD, so we now know `originalHost` can't possibly have a valid `DNSZone` available.
-	if host == tld {
+	// Bail out only if the host equals an ICANN TLD (e.g. `com`, `org`). The
+	// Go public suffix list also contains privately-managed domains such as
+	// `httpbin.org`, `github.io`, `amazonaws.com`, `azurewebsites.net` - for
+	// those, PublicSuffix returns the host itself but the second return
+	// value (`icann`) is false. Ignoring that flag rejected any DNSRecord
+	// whose rootHost happened to appear in the private suffix list, even
+	// when a matching zone was explicitly provisioned in the provider
+	// secret.
+	tld, icann := publicsuffix.PublicSuffix(host)
+	if icann && host == tld {
 		return nil, "", fmt.Errorf("%w: %s", ErrNoZoneForHost, originalHost)
 	}
 
--- a/internal/provider/provider_test.go
+++ b/internal/provider/provider_test.go
@@ -210,6 +210,74 @@ func Test_findDNSZoneForHost(t *testing.T) {
 			wantErr:       true,
 			expectedErrIs: ErrMultipleZonesFound,
 		},
+
+		// Private suffix coverage (per #762 review). The Go public suffix list
+		// also contains privately-managed suffixes such as `httpbin.org`,
+		// `github.io`, and `s3.amazonaws.com`. These should NOT be rejected as
+		// TLDs — only ICANN suffixes (e.g. `com`, `co.uk`) are rejected.
+		{
+			name: "private suffix as host with matching zone (httpbin.org)",
+			host: "httpbin.org",
+			zones: []DNSZone{
+				{DNSName: "httpbin.org", ID: "zone-1"},
+			},
+			denyApex:      false,
+			wantZone:      "httpbin.org",
+			wantSubdomain: "",
+			wantErr:       false,
+		},
+		{
+			name: "subdomain of a private suffix matches its parent zone",
+			host: "sub.httpbin.org",
+			zones: []DNSZone{
+				{DNSName: "httpbin.org", ID: "zone-1"},
+			},
+			denyApex:      false,
+			wantZone:      "httpbin.org",
+			wantSubdomain: "sub",
+			wantErr:       false,
+		},
+		{
+			name: "multi-label private suffix as host with matching zone (s3.amazonaws.com)",
+			host: "s3.amazonaws.com",
+			zones: []DNSZone{
+				{DNSName: "s3.amazonaws.com", ID: "zone-1"},
+			},
+			denyApex:      false,
+			wantZone:      "s3.amazonaws.com",
+			wantSubdomain: "",
+			wantErr:       false,
+		},
+		{
+			name: "private suffix host with no matching zone returns ErrNoZoneForHost",
+			host: "httpbin.org",
+			zones: []DNSZone{
+				{DNSName: "example.com"},
+			},
+			denyApex:      false,
+			wantErr:       true,
+			expectedErrIs: ErrNoZoneForHost,
+		},
+		{
+			name: "private suffix as host with denyApex=true returns ErrApexDomainNotAllowed",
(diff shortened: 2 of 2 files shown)


maksymvavilov (changes requested):
Thanks for the contribution! The core approach — gating on the `icann` flag — is the right fix for this problem.

A few things before we can merge:

**Factual error in the PR description**

The description lists `amazonaws.com` as a private suffix. It's not — `publicsuffix.PublicSuffix("amazonaws.com")` returns `("com", true)`. The private suffix in that family is `s3.amazonaws.com`.

**Tests are required**

This PR needs test cases added to `internal/provider/provider_test.go`. The existing suite has no coverage for private suffix domains. At minimum we'd need:

- Private suffix as host with matching zone (e.g. `httpbin.org` with zone `httpbin.org`, `denyApex=false`) — should succeed
- Private suffix as host with no matching zone — should fail with `ErrNoZoneForHost`
- ICANN TLD still rejected (`com`, `co.uk`) — confirming no regression
- Subdomain of a private suffix (`sub.httpbin.org` with zone `httpbin.org`) — should succeed
- Private suffix with `denyApex=true` — should return `ErrApexDomainNotAllowed`
- Multi-label private suffix (`s3.amazonaws.com` with matching zone) — should succeed

**Comment is too verbose**

The 7-line inline comment is more than needed. Something like this would be sufficient:

```go
// Only reject if the host is an ICANN TLD. Private suffixes (e.g. httpbin.org)
// should proceed to zone matching.
tld, icann := publicsuffix.PublicSuffix(host)
if icann && host == tld {
```

SAY-5:
@maksymvavilov applied:

1. **Comment trimmed** to your 2-line suggested form. Updated the example to `s3.amazonaws.com` (private suffix) and `github.io` since you correctly noted `amazonaws.com` itself is ICANN — `PublicSuffix("amazonaws.com")` → `("com", true)`. The private-suffix family is `s3.amazonaws.com`.

2. **Tests added** in `internal/provider/provider_test.go` covering all six cases you listed:
   - Private suffix host with matching zone (`httpbin.org`) — success
   - Subdomain of private suffix (`sub.httpbin.org`) matches parent zone — success
   - Multi-label private suffix host (`s3.amazonaws.com`) with matching zone — success
   - Private suffix host with no matching zone — `ErrNoZoneForHost`
   - Private suffix host with `denyApex=true` — `ErrApexDomainNotAllowed`
   - ICANN TLD `co.uk` still rejected (no regression)

All 18 cases in `Test_findDNSZoneForHost` pass under `go test -tags unit ./internal/provider/`.

maksymvavilov (approved):
Looks good. Thank you for your work!

maksymvavilov:
@SAY-5, I think you must have a verified signature for your commit to have it merged. Can you, please, address it?

SAY-5:
@maksymvavilov pushed re-signed commits at `2d4949f` and `4d0e7c3`. Locally `git log --show-signature` reports

```
Good "git" signature for SAY-5@users.noreply.github.com with ED25519 key
SHA256:6YbCdfFU1NAyFJh/ZTCFtdJ2USLbaoNncrTy5uGOVPg
```

The PR API still reports the commits as unsigned on the GitHub side though, because that signing key hasn't yet been uploaded to my GitHub account's *Signing keys* list (the auth token I have only carries `repo`/`workflow` scopes, not `admin:ssh_signing_key`). I'll get the public key registered on the account out-of-band so the *Verified* badge picks up the existing commits. If you want me to use a different signing identity in the meantime, happy to swap.

maksymvavilov:
> @maksymvavilov pushed re-signed commits at `2d4949f` and `4d0e7c3`. Locally `git log --show-signature` reports
> 
> ```
> Good "git" signature for SAY-5@users.noreply.github.com with ED25519 key
> SHA256:6YbCdfFU1NAyFJh/ZTCFtdJ2USLbaoNncrTy5uGOVPg
> ```
> 
> The PR API still reports the commits as unsigned on the GitHub side though, because that signing key hasn't yet been uploaded to my GitHub account's _Signing keys_ list (the auth token I have only carries `repo`/`workflow` scopes, not `admin:ssh_signing_key`). I'll get the public key registered on the account out-of-band so the _Verified_ badge picks up the existing commits. If you want me to use a different signing identity in the meantime, happy to swap.

@SAY-5, the branch protection rules we have in place would not allow this to be merged otherwise. Can you, please, add the public key manually? There is a [guide](https://docs.github.com/en/authentication/connecting-to-github-with-ssh/adding-a-new-ssh-key-to-your-github-account) if you need it or something like this will also do if you have a gh cli: 
```
gh ssh-key add ~/.ssh/id_ed25519.pub --type signing    
```

PR #771, opened 2026-05-21
fix: suppress yq install stdout in Makefile

thomasmaas:
The yq install output leaks into `make bundle-operator-image-url` stdout, causing the tag-release bundle build to fail. Adds `@` prefix to suppress.

## Chores
  * Refined build command output display settings to enhance build process clarity.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Kuadrant/dns-operator/pull/771?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -496,7 +496,7 @@ $(ACT): $(LOCALBIN)
 .PHONY: yq
 yq: $(YQ) ## Download yq locally if necessary.
 $(YQ): $(LOCALBIN)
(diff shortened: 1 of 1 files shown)


mikenairn:
Replacing with https://github.com/Kuadrant/dns-operator/pull/772

This probably would work, but the above PR was the fix we added to main for this issue. As long as the `make yq` command is run as a separate target before running a target that depends on it (the builds tags job does this) then it won't output the details of the install of yq.

PR #786, opened 2026-06-22
Bump golangci-lint to v2.12.2

thomasmaas:
Upgrade golangci-lint from v2.7.2 to v2.12.2, pin install script to version tag, and update CI workflow version.


mikenairn (dismissed):

mikenairn:
@thomasmaas The linter is picking up a new issue that will need to be fixed and committed for this to merge. The PR doesn't run lint on all files, only files changed which is why it passed on the PR itself.

```
Error: cmd/plugin/output/text.go:43:23: inline: Constant reflect.Ptr should be inlined (govet)
     case reflect.Struct, reflect.Ptr:
```

mikenairn (approved):

PR #773, opened 2026-05-27
feat: add USE_IMAGE_DIGESTS flag and file-based catalog

mikenairn:
## Summary

Adds optional USE_IMAGE_DIGESTS flag and file-based catalog (FBC) generation to bring dns-operator in line with kuadrant-operator, authorino-operator, and limitador-operator for consistency across the Kuadrant ecosystem.

### Digest Support
- Added `USE_IMAGE_DIGESTS` flag (default: false) to enable SHA256 digest references in bundles
- Created `BUNDLE_GEN_FLAGS` variable with conditional `--use-image-digests`
- Added bundle post-generation step to deduplicate relatedImages (operator-sdk creates duplicates when using digests)
- Added `IMAGE_TAG` variable support throughout Makefile for flexible image tagging

### Bundle Expiry Support
- Added Quay image expiry label support for bundle images (matching other operators)
- Inject `QUAY_EXPIRY_TIME_LABEL` into bundle.Dockerfile during bundle generation
- Pass `QUAY_IMAGE_EXPIRY` build-arg when building bundle images
- Set bundle expiry to `never` for tag releases in CI workflow

### File-Based Catalog (FBC)
- Added catalog generation support via `make/catalog.mk` (matching other Kuadrant operators)
- Added `utils/generate-catalog.sh` for automated FBC creation
- Added `catalog/dns-operator-channel-entry.yaml` template
- Removed duplicate/old index-based catalog targets from main Makefile
- Excluded generated `catalog/` directory in .gitignore

### Consistency Improvements
These changes bring dns-operator's build and release infrastructure in line with kuadrant-operator, authorino-operator, and limitador-operator, ensuring consistent patterns across all Kuadrant components.

**Note:** CoreDNS image (`coredns-kuadrant`) is not included in relatedImages. Users requiring CoreDNS should mirror it manually. See kuadrant-operator documentation for details.

## Benefits

Digest-based references enable ImageDigestMirrorSet to redirect image pulls to internal registries, supporting disconnected/air-gapped cluster scenarios.

## Related

Related to https://github.com/Kuadrant/kuadrant-operator/issues/1894

PR #787, opened 2026-06-24
fix: restore catalog prerequisite to catalog-build target

Boomatang:
# Summary

Restores the `catalog` prerequisite on the `catalog-build` Makefile target, which was accidentally dropped in #773.

Without this dependency, `make catalog-build` skips catalog generation (dockerfile, operator.yaml, and OPM validation) and attempts to build a container image from potentially missing or stale catalog artifacts. This broke both local `make catalog-build` invocations and CI catalog image builds.

# Changes

- Added `catalog` as a prerequisite to the `catalog-build` target in `make/catalog.mk`, restoring the dependency chain: `catalog-build` -> `catalog` -> `opm` / `$(CATALOG_DOCKERFILE)` / `$(CATALOG_FILE)`.

# Test plan

- [ ] Run `make catalog-build` from a clean checkout and confirm it generates catalog content before building the image
- [ ] Verify CI `build-catalog` job passes on this branch

## Summary by CodeRabbit

* **Chores**
  * Updated the catalog image build process so the catalog content is generated first, helping ensure builds use the latest generated assets.
  * Kept the existing image build and tagging behaviour unchanged.

### DIFF
--- a/make/catalog.mk
+++ b/make/catalog.mk
@@ -52,7 +52,7 @@ catalog: opm ## Generate catalog content and validate.
 # Build a catalog image by adding bundle images to an empty catalog using the operator package manager tool, 'opm'.
 # Ref https://olm.operatorframework.io/docs/tasks/creating-a-catalog/#catalog-creation-with-raw-file-based-catalogs
 .PHONY: catalog-build
-catalog-build: ## Build a catalog image.
+catalog-build: catalog ## Build a catalog image.
 	# Build the Catalog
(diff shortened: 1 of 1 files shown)


mikenairn (approved):
This file based catalog make file was added to be consistent with the other repos and should be the same across them all. These other repos appear to use a different workflow though(missed that when i added it), which explicitly calls the "catalog" target first. This fix however seems more appropriate, the make targets should work as expected without requiring others to be called beforehand, and arguably the other repos processes are wrong here.

PR #795, opened 2026-07-15
fix: persist DEFAULT_CHANNEL in prepare-release

thomasmaas:
## Summary
- Add `DEFAULT_CHANNEL` and `BUNDLE_DEFAULT_CHANNEL` to the values written to `make/release.mk` by the `prepare-release` target

## Why
`prepare-release` already persisted `CHANNELS`/`BUNDLE_CHANNELS` into `make/release.mk`, but never did the same for `DEFAULT_CHANNEL`/`BUNDLE_DEFAULT_CHANNEL`. This meant any subsequent `make bundle` invocation on a release branch silently fell back to the Makefile default (unset), producing bundles without an explicit `channel.default.v1` annotation.

Same class of bug fixed for limitador-operator in Kuadrant/limitador-operator#268 / Kuadrant/limitador-operator#269.

## Test plan
- [ ] `make prepare-release VERSION=0.0.0-test CHANNELS=stable DEFAULT_CHANNEL=stable` writes both channel values to `make/release.mk`
- [ ] Generated `bundle.Dockerfile` and `bundle/metadata/annotations.yaml` contain `channel.default.v1=stable`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

- **Bug Fixes**
  - Release preparation now preserves the configured default update channel in generated release settings.
  - Bundled releases are configured to use the selected default channel automatically.
  - This ensures release packages and update settings remain aligned with the chosen distribution channel, reducing the risk of releases being published or delivered through an unintended channel.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -612,6 +612,9 @@ prepare-release: ## Generates a makefile that will override environment variable
 	echo -e "#Release default values\\nIMG=$(IMAGE_TAG_BASE):$(IMG_TAG)\nBUNDLE_IMG=$(IMAGE_TAG_BASE)-bundle:$(IMG_TAG)\n\
 	CATALOG_IMG=$(IMAGE_TAG_BASE)-catalog:$(IMG_TAG)\nCOREDNS_IMG=$(COREDNS_IMAGE_TAG_BASE):$(IMG_TAG)\nCHANNELS=$(CHANNELS)\n\
 	BUNDLE_CHANNELS=--channels=$(CHANNELS)\nVERSION=$(VERSION)" > $(RELEASE_FILE)
+ifneq ($(origin DEFAULT_CHANNEL), undefined)
+	echo -e "DEFAULT_CHANNEL=$(DEFAULT_CHANNEL)\nBUNDLE_DEFAULT_CHANNEL=--default-channel=$(DEFAULT_CHANNEL)" >> $(RELEASE_FILE)
+endif
 	$(MAKE) set-image-refs
 	$(MAKE) bundle
 	$(MAKE) helm-build VERSION=$(VERSION)


Makefile
```diff
@@ -626,6 +626,9 @@ prepare-release: ## Generates a makefile that will override environment variable
 	echo -e "#Release default values\\nIMG=$(IMAGE_TAG_BASE):$(IMG_TAG)\nBUNDLE_IMG=$(IMAGE_TAG_BASE)-bundle:$(IMG_TAG)\n\
 	CATALOG_IMG=$(IMAGE_TAG_BASE)-catalog:$(IMG_TAG)\nCOREDNS_IMG=$(COREDNS_IMAGE_TAG_BASE):$(IMG_TAG)\nCHANNELS=$(CHANNELS)\n\
 	BUNDLE_CHANNELS=--channels=$(CHANNELS)\nVERSION=$(VERSION)" > $(RELEASE_FILE)
+ifneq ($(origin DEFAULT_CHANNEL), undefined)
+	echo -e "DEFAULT_CHANNEL=$(DEFAULT_CHANNEL)\nBUNDLE_DEFAULT_CHANNEL=--default-channel=$(DEFAULT_CHANNEL)" >> $(RELEASE_FILE)
+endif
```
mikenairn, line 617: I don't think you need to include `BUNDLE_DEFAULT_CHANNEL` here just adding `DEFAULT_CHANNEL` should be sufficient, the Makefile will set the correct value for `BUNDLE_DEFAULT_CHANNEL` based on that being set or not.  

Would the idea not also be that it needs to always be set to a value? Maybe this command needs to fail if its not passed in and the [docs](https://github.com/Kuadrant/dns-operator/blob/main/docs/RELEASE.md#new-majorminor-version) updated?

Makefile
```diff
@@ -612,6 +612,9 @@ prepare-release: ## Generates a makefile that will override environment variable
 	echo -e "#Release default values\\nIMG=$(IMAGE_TAG_BASE):$(IMG_TAG)\nBUNDLE_IMG=$(IMAGE_TAG_BASE)-bundle:$(IMG_TAG)\n\
 	CATALOG_IMG=$(IMAGE_TAG_BASE)-catalog:$(IMG_TAG)\nCOREDNS_IMG=$(COREDNS_IMAGE_TAG_BASE):$(IMG_TAG)\nCHANNELS=$(CHANNELS)\n\
 	BUNDLE_CHANNELS=--channels=$(CHANNELS)\nVERSION=$(VERSION)" > $(RELEASE_FILE)
```
mikenairn, line 614:
```suggestion
	echo -e "#Release default values\\nIMG=$(IMAGE_TAG_BASE):$(IMG_TAG)\nBUNDLE_IMG=$(IMAGE_TAG_BASE)-bundle:$(IMG_TAG)\n\
	CATALOG_IMG=$(IMAGE_TAG_BASE)-catalog:$(IMG_TAG)\nCOREDNS_IMG=$(COREDNS_IMAGE_TAG_BASE):$(IMG_TAG)\nCHANNELS=$(CHANNELS)\n\
	DEFAULT_CHANNEL=$(DEFAULT_CHANNEL)\nVERSION=$(VERSION)" > $(RELEASE_FILE)
```

mikenairn: What you are doing is essentially the same as this. Note i also removed `BUNDLE_CHANNELS` here since that doesn't actually need to be here either. These values for the bundle command are set [here](https://github.com/Kuadrant/dns-operator/blob/main/Makefile#L16-L33) based on `CHANNELS` and `DEFAULT_CHANNEL` being set.

mikenairn: Note also, you are not setting a value for `DEFAULT_CHANNEL` anywhere, the only way this works (as you have it), is you pass it in: 

```
make prepare-release IMG_TAG=release-0.2 VERSION=0.2.0-dev CHANNELS=alpha DEFAULT_CHANNEL=alpha
```

Issue #778, opened 2026-06-17
Rework premature reconcile check to allow group/status propagation


philbrookes:
Part of #776 — DNS Failover via Groups — Tech Preview

Rework or remove the premature reconcile check so that spec and status changes (group assignment, delegation role) trigger immediate reconciliation rather than waiting for the next full reconcile window (#664).

Issue #776, opened 2026-06-17
DNS Failover via Groups — Tech Preview


philbrookes:
## User Experience

Today in developer preview, the user can failover between clusters by running CLI commands to update the active groups TXT record. However, group transitions can stall silently due to a premature reconcile bug, registry entries may be inaccurate after transitions, there is no way to preview the impact of removing a group before executing it, and no metrics exist to confirm failover completed — leaving the user reliant on polling DNSRecord statuses and inspecting DNS directly.

In tech preview, these gaps are resolved: group status changes propagate immediately, registry entries stay accurate throughout transitions, and the CLI previews affected endpoints before a group removal is confirmed. Prometheus metrics expose group state and cleanup operations, giving the user clear observability into failover progress. The result is that when a cluster goes down, the user runs a single CLI command to activate the standby group, watches metrics and record statuses converge, and has confidence the failover completed cleanly.

### 1. Reliable group transitions with E2E validation

As a cluster operator, I want group status changes to propagate immediately, registry entries to stay accurate across active/inactive transitions, and the CLI to preview impact before removing a group, so that failover is reliable, predictable, and safe — validated by E2E tests against real DNS providers.

Acceptance Criteria: Premature reconcile check no longer blocks group/status updates (#664). Inactive controllers always write accurate groupID to provider registry (#637). `remove-active-group` CLI shows affected endpoints before execution (#663, #620). E2E tests cover group assignment, transitions, and cleanup against real providers (#665).

Tasks:
- [ ] #777 — Fix delegation bug hidden by premature reconcile check (5 points)
- [ ] #778 — Rework premature reconcile check to allow group/status propagation (3 points)
- [ ] #779 — Always write groupID to provider registry for inactive controllers (3 points)
- [ ] #780 — Integration tests for groupID registry write behaviour (2 points)
- [ ] #781 — Parse zone TXT registry records into group-aware data structure (3 points)
- [ ] #782 — CLI dry-run mode for remove-active-group (3 points)

### 2. Group-aware observability

As an SRE, I want Prometheus metrics exposing each record's group assignment, active/inactive state, and inactive group cleanup operations, so that I can observe failover progress and build alerts.

Acceptance Criteria: Metric for current group assignment per DNSRecord. Metric for active/inactive state per DNSRecord. Metric for inactive group cleanup operations. Metrics documented with example Grafana queries or recording rules.

Tasks:
[...]

PR #803, opened 2026-07-27
fix: prevent deletion status flip when awaiting remote controller

### LINKED ISSUE #777: Fix delegation bug hidden by premature reconcile check

Investigate and fix the delegation bug exposed by removing the premature reconcile check (#664). The `recordReceivedPrematurely` function blocks status/group changes from propagating, but removing it causes delegation tests to fail — that underlying bug needs to be identified and fixed first.

philbrookes:
## Summary
- `ProviderEndpointsDeletion()` now recognizes `ProviderEndpointsRemoved` as a state that has already progressed past the deletion stage, preventing a status flip loop during delegating record deletion
- This fixes the underlying delegation bug that `recordReceivedPrematurely` was masking, unblocking #778 (rework premature reconcile check)

## Problem
When deleting a delegating record with pending remote record statuses, the deletion flow would flip between `ProviderEndpointsDeletion` and `ProviderEndpointsRemoved` indefinitely:

1. Local controller sets `ProviderEndpointsRemoved` after cleanup
2. `ProviderEndpointsRemoved()` returns false (remote controller hasn't confirmed yet)
3. `ProviderEndpointsDeletion()` sees reason `ProviderEndpointsRemoved` which is not `ProviderEndpointsDeletion` → returns false
4. Deletion flow overwrites status back to `ProviderEndpointsDeletion`
5. Cycle repeats

The `recordReceivedPrematurely` check masked this by throttling reconciliation, giving remote controllers time to catch up.

## Tested locally by:
- [x] `make test-unit` — all passing
- [x] `make test-integration` — all 44 specs passing (including all delegation tests)
- [x] Verified delegation tests also pass with premature check temporarily removed, confirming this fix eliminates the masked bug

Closes #777
Part of #776

### DIFF
--- a/api/v1alpha1/dnsrecord_types.go
+++ b/api/v1alpha1/dnsrecord_types.go
@@ -246,9 +246,12 @@ func (s *DNSRecordStatus) ProviderEndpointsRemoved() bool {
 }
 
 // ProviderEndpointsDeletion return true if the ready status condition has the reason set to "ProviderEndpointsDeletion"
+// or "ProviderEndpointsRemoved" (which is a progression past the deletion stage).
 func (s *DNSRecordStatus) ProviderEndpointsDeletion() bool {
 	readyCond := meta.FindStatusCondition(s.Conditions, string(ConditionTypeReady))
-	if readyCond == nil || readyCond.Reason == string(ConditionReasonProviderEndpointsDeletion) {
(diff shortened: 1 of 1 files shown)


maksymvavilov (approved):
Looks good and seems to be working 🤔

PR #808, opened 2026-08-04
Use controller SDK backoff and write groupID to provider registry for inactive controllers

### LINKED ISSUE #779: Always write groupID to provider registry for inactive controllers

Update inactive group controllers to always write accurate groupID to provider registry entries (#637). Currently when a controller is in an inactive group it exits reconciliation early and skips registry writes, leaving stale group values. The controller should continue far enough into reconciliation to update its registry entries even when inactive.

philbrookes:
## Summary
- Adds a `registryOnly` mode to `applyChanges` that strips Creates, Deletes, and target-changing Updates from the plan, allowing only TXT registry label updates to be applied
- Inactive group controllers now call `applyChanges(registryOnly=true)` to write accurate groupID labels to TXT registry entries without modifying DNS records
- Fixes stale group values in TXT records that prevented the active controller's `unpublishInactiveGroups` from correctly identifying and cleaning up inactive group records

Fixes #779

## Test plan
- [x] Unit tests pass (`make test-unit`)
- [x] Integration tests pass (`make test-integration`) — all 45 specs
- [x] New integration test verifies inactive controllers write groupID to registry without modifying zone DNS records


philbrookes:
replaced by: https://github.com/Kuadrant/dns-operator/pull/809

PR #809, opened 2026-08-04
Controllers in inactive groups still update group ID and replace premature check with controller SDK backoff.

### LINKED ISSUE #778: Rework premature reconcile check to allow group/status propagation

### LINKED ISSUE #779: Always write groupID to provider registry for inactive controllers

### LINKED ISSUE #780: Integration tests for groupID registry write behaviour

Add integration tests for the groupID registry write behaviour (#637), covering: inactive controller writes correct group, group change from one value to another updates registry, reactivated group has accurate registry entries.

philbrookes:
## Summary

This PR combines three related changes for the DNS groups feature:

### 1. Refactor: replace manual reconcile backoff with SDK rate limiter (#778)
- Replaces manual `nextReconcileAt`/`lastReconcileAt` tracking with the SDK's built-in `ItemExponentialFailureRateLimiter`
- Removes `RequeueTime` and `WriteCounter` from `DNSRecordStatus`, simplifying the CRD
- Adds `specOrDeletionChangedPredicate` to filter watch events to spec changes (generation changed) and deleting objects, preventing status-only updates from causing tight reconciliation loops
- Uses `defaultValidationRequeue` consistently for pre-deletion status transitions

### 2. Feat: always write groupID to provider registry for inactive controllers (#779)
- Adds a `registryOnly` mode to `applyChanges` that strips Creates, Deletes, and target-changing Updates from the plan, allowing only TXT registry label updates (e.g. groupID) to be applied
- Inactive group controllers now call `applyChanges(registryOnly=true)` to write accurate groupID labels to TXT registry entries without modifying DNS records
- Fixes stale group values in TXT records that prevented the active controller's `unpublishInactiveGroups` from correctly identifying and cleaning up inactive group records
- Propagates errors from inactive registry updates through status conditions for faster requeue on failure

### 3. Test: integration test for groupID registry self-correction (#780)
- Adds a test that tampers TXT registry group labels on an inactive controller's authoritative record and verifies the inactive controller corrects them via `registryOnly` applyChanges
- Adds `tamperRegistryGroup` helper to simulate stale/incorrect group labels by modifying authoritative record TXT endpoints
- Updates `filterDNSEndpoints` to clear Labels for stable endpoint comparisons

Fixes #778, #779, #780

## Test plan
- [x] Unit tests pass (`make test-unit`)
- [x] Integration tests pass (`make test-integration`) — all 46 specs
- [x] New integration test verifies inactive controllers write groupID to registry without modifying zone DNS records
- [x] New integration test verifies inactive controllers correct tampered registry group entries

## Summary by CodeRabbit

- **Breaking Changes**
  - Removed the `queuedAt` and `validFor` fields from DNS record status information and documentation.
  - Removed the `--valid-for` configuration option and corresponding environment variable.

- **Improvements**
  - Reconciliation now uses rate-limited exponential backoff between configured minimum and maximum requeue intervals.
  - Added validation for requeue interval limits.
  - Inactive grouped records keep registry entries synchronised without altering authoritative DNS endpoints.
  - Improved handling of reconciliation conflicts, registry update failures and unnecessary status-only updates.


internal/controller/dnsrecord_controller.go
```diff
@@ -507,73 +496,32 @@ func (r *DNSRecordReconciler) SetupWithManager(mgr ctrl.Manager, maxRequeue, val
...
 	}
-	// double the duration. Return the max timeout if overshoot
-	newRequeue := lastRequeue * 2
-	if newRequeue > defaultRequeueTime {
-		return defaultRequeueTime
+	if e.ObjectNew.GetDeletionTimestamp() != nil {
+		return true
 	}
-	return newRequeue
+	return false
```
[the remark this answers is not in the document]
philbrookes: The deletion state machine needs status-patch-driven watch events to progress efficiently. This suggestion would slow deletion down without meaningful benefit since deletion is a short-lived process, not a steady-state loop.

internal/controller/dnsrecord_controller.go
```diff
@@ -388,80 +385,72 @@ func (r *DNSRecordReconciler) updateStatus(ctx context.Context, previous, curren
...
 // SetupWithManager sets up the controller with the Manager.
-func (r *DNSRecordReconciler) SetupWithManager(mgr ctrl.Manager, maxRequeue, validForDuration, minRequeue time.Duration, healthProbesEnabled, allowInsecureHealthCert bool) error {
+func (r *DNSRecordReconciler) SetupWithManager(mgr ctrl.Manager, maxRequeue, minRequeue time.Duration, healthProbesEnabled, allowInsecureHealthCert bool) error {
+	if minRequeue > maxRequeue {
+		return fmt.Errorf("minRequeue (%s) must not exceed maxRequeue (%s)", minRequeue, maxRequeue)
+	}
+
 	defaultRequeueTime = maxRequeue
-	validFor = validForDuration
 	defaultValidationRequeue = minRequeue
```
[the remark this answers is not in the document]
philbrookes: This is addressed in the latest changes.

internal/controller/base_dnsrecord_reconciler.go
```diff
@@ -195,8 +198,30 @@ func (r *BaseDNSRecordReconciler) applyChanges(ctx context.Context, dnsRecord DN
...
+				filteredOld = append(filteredOld, oldEp)
+				filteredNew = append(filteredNew, newEp)
+			}
+		}
+		recordPlan.Changes.UpdateOld = filteredOld
+		recordPlan.Changes.UpdateNew = filteredNew
+	} else {
+		dnsRecord.SetStatusDomainOwners(recordPlan.Owners)
+		dnsRecord.SetStatusEndpoints(specEndpoints)
+	}
```
mikenairn: This seems like the wrong place to do this, it's adding logic into `applyChanges` that should be elsewhere. We already have a `GroupRegistry` that is used to manipulate endpoint input to the plan and has it's own `ApplyChanges` that gets called if adapting the registry for group functionality. I'd have expected that to be the place where logic relating to group specific behaviour would live and can be more easily tested in isolation.

I'd suggest passing the IsActive status to GroupRegistry when it's constructed:
```
recordRegistry = GroupRegistry{
    Registry: recordRegistry,
    Group:    dnsRecord.GetGroup(),
    IsActive: dnsRecord.IsActive(),  // New field
}
```
And during `GroupRegistry.ApplyChanges()` handle filtering the plan changes when inactive: 
```
func (g GroupRegistry) ApplyChanges(ctx context.Context, changes *plan.Changes) error {
      // If inactive group, filter to registry-only changes
      if g.Group.IsSet() && !g.IsActive {
          //filter plan changes as desired
      }
      return g.Registry.ApplyChanges(ctx, changes)
  }
```
This removes the registryOnly parameter entirely and keeps group-specific logic within the group registry wrapper.

internal/controller/base_dnsrecord_reconciler.go
```diff
@@ -218,20 +243,20 @@ func (r *BaseDNSRecordReconciler) updateStatus(ctx context.Context, client clien
...
+// and returns the given reconcile.Result.
+func (r *BaseDNSRecordReconciler) updateStatusAndRequeue(ctx context.Context, c client.Client, previous, current DNSRecordAccessor, result reconcile.Result) (reconcile.Result, error) {
 	logger := log.FromContext(ctx)
 	patch := client.MergeFrom(previous.GetDNSRecord())
 	// update the record after setting the status
 	if !equality.Semantic.DeepEqual(previous.GetStatus(), current.GetStatus()) {
 		if updateError := c.Status().Patch(ctx, current.GetDNSRecord(), patch); updateError != nil {
 			if apierrors.IsConflict(updateError) {
-				return ctrl.Result{RequeueAfter: time.Second}, nil
+				return ctrl.Result{Requeue: true}, nil
```
mikenairn: `Requeue: true` is Deprecated, reason it was using `RequeueAfter` https://github.com/kubernetes-sigs/controller-runtime/blob/v0.21.0/pkg/reconcile/reconcile.go#L30

internal/controller/base_dnsrecord_reconciler.go
```diff
@@ -218,20 +243,20 @@ func (r *BaseDNSRecordReconciler) updateStatus(ctx context.Context, client clien
 }
 
 // updateStatusAndRequeue will update the status of the record if the current and previous status is different
-// and returns a reconcile.result that re-queues at the given time.
-func (r *BaseDNSRecordReconciler) updateStatusAndRequeue(ctx context.Context, c client.Client, previous, current DNSRecordAccessor, requeueTime time.Duration) (reconcile.Result, error) {
+// and returns the given reconcile.Result.
+func (r *BaseDNSRecordReconciler) updateStatusAndRequeue(ctx context.Context, c client.Client, previous, current DNSRecordAccessor, result reconcile.Result) (reconcile.Result, error) {
```
mikenairn: I don't see a good reason to change `updateStatusAndRequeue` from taking time.Duration to taking Result. The original design was clearer:
  - Need to requeue? Call updateStatusAndRequeue(ctx, client, prev, curr, 5*time.Second)
  - Don't need to requeue? Call updateStatus(ctx, client, prev, curr, err)

Now both functions essentially do the same thing (pass through a Result), which defeats the purpose of having two separate functions. The name distinction no longer provides value if the caller has to construct the Result themselves anyway.

internal/controller/dnsrecord_controller.go
```diff
@@ -388,80 +385,72 @@ func (r *DNSRecordReconciler) updateStatus(ctx context.Context, previous, curren
...
 	probesEnabled = healthProbesEnabled
 	allowInsecureCert = allowInsecureHealthCert
 
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&v1alpha1.DNSRecord{}).
+		For(&v1alpha1.DNSRecord{}, builder.WithPredicates(specOrDeletionChangedPredicate{})).
+		WithOptions(controller.Options{
+			RateLimiter: workqueue.NewTypedItemExponentialFailureRateLimiter[reconcile.Request](minRequeue, maxRequeue),
```
mikenairn: Hmm ok, i think i see now why you wanted to use `Requeue: true`, unfortunately  this is deprecated, and i don't think we should be building a solution around an API that will be removed. 

https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/reconcile/reconcile.go#L42

It looks like `Requeue` and `RequeueAfter` do not behave the same, one(RequeueAfter) bypasses the rate limiter apparently since it was never intended for non error cases.

philbrookes: yeah, this is why. The alternatives as I see, is to either reset the backoff ratelimiting using RequeueAfter, or implement the ratelimiting ourselves again :\

mikenairn:
This has the same concern I raised in #733,  with recordReceivedPrematurely removed, AwaitingValidation is set and then immediately cleared on the next reconcile (5 seconds later). There's no meaningful validation window anymore. What purpose does AwaitingValidation serve with these changes?

philbrookes:
> This has the same concern I raised in #733, with recordReceivedPrematurely removed, AwaitingValidation is set and then immediately cleared on the next reconcile (5 seconds later). There's no meaningful validation window anymore. What purpose does AwaitingValidation serve with these changes?

This is true, I'd completely forgotten that issue. I wonder if we could have the awaiting validation stay true until the last write was over a X time ago and there are no changes made in this reconcile?

philbrookes:
> > This has the same concern I raised in #733, with recordReceivedPrematurely removed, AwaitingValidation is set and then immediately cleared on the next reconcile (5 seconds later). There's no meaningful validation window anymore. What purpose does AwaitingValidation serve with these changes?
> 
> This is true, I'd completely forgotten that issue. I wonder if we could have the awaiting validation stay true until the last write was over a X time ago and there are no changes made in this reconcile?

I see that this is exactly what we discussed before. I'm not sure if the validation loop still has enough value to be worth the effort of recoding it that way though.

Maybe the better answer here is to remove the concept of a DNSRecord verification step?

internal/controller/helper_test.go
```diff
@@ -145,6 +147,90 @@ func createDefaultDNSProviderSecret(ctx context.Context, namespace, zoneDomainNa
...
+	records, err := p.Records(ctx)
+	Expect(err).NotTo(HaveOccurred())
+	return records
+}
+
+// readRegistryMap reads zone records and parses TXT records into a RegistryMap
+func readRegistryMap(ctx context.Context, zoneDomainName string) *externaldnsregistry.RegistryMap {
+	records := readZoneRecords(ctx, zoneDomainName)
+	return externaldnsregistry.TxtRecordsToRegistryMap(records, txtRegistryPrefix, txtRegistrySuffix, txtRegistryWildcardReplacement, []byte(txtRegistryEncryptAESKey))
+}
```
maksymvavilov: I don't think this one is used anywhere

internal/controller/dnsrecord_controller.go
```diff
@@ -317,10 +313,21 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
 	if !dnsRecord.IsActive() {
+		if dnsProvider, err := r.getDNSProvider(ctx, dnsRecord); err != nil {
+			logger.Error(err, "Failed to load DNS provider for inactive group registry update")
+			dnsRecord.SetStatusCondition(string(v1alpha1.ConditionTypeReady), metav1.ConditionFalse,
+				string(v1alpha1.ConditionReasonProviderError), fmt.Sprintf("The dns provider could not be loaded: %v", err))
+			return r.updateStatus(ctx, previous, dnsRecord, false, err)
+		} else if _, err := r.applyChanges(ctx, dnsRecord, dnsProvider, false, true); err != nil {
+			logger.Error(err, "Failed to update registry for inactive group")
+			dnsRecord.SetStatusCondition(string(v1alpha1.ConditionTypeReady), metav1.ConditionFalse,
+				"RegistryError", fmt.Sprintf("Failed to update registry for inactive group: %v", err))
```
maksymvavilov: The "RegistryError" - should it be a type in [here](https://github.com/Kuadrant/dns-operator/blob/main/api/v1alpha1/conditions.go)?

internal/controller/helper_test.go
```diff
@@ -145,6 +147,90 @@ func createDefaultDNSProviderSecret(ctx context.Context, namespace, zoneDomainNa
 	return secret
 }
 
+// readZoneRecords reads all endpoints (DNS + TXT) from the inmemory provider's shared zone
```
maksymvavilov, line 150: We should delete `RandomizeValidationDuration()` and it's test - this is a dead code now

philbrookes:
> > > This has the same concern I raised in #733, with recordReceivedPrematurely removed, AwaitingValidation is set and then immediately cleared on the next reconcile (5 seconds later). There's no meaningful validation window anymore. What purpose does AwaitingValidation serve with these changes?
> > 
> > 
> > This is true, I'd completely forgotten that issue. I wonder if we could have the awaiting validation stay true until the last write was over a X time ago and there are no changes made in this reconcile?
> 
> I see that this is exactly what we discussed before. I'm not sure if the validation loop still has enough value to be worth the effort of recoding it that way though.
> 
> Maybe the better answer here is to remove the concept of a DNSRecord verification step?

Mike and I discussed this offline, and came to the conclusion that we could implement a backoff using the lastTransitionTime of the ready condition, and avoid using the Requeue: true.

I will implement and push up some changes to reflect this.

cmd/main.go
```diff
@@ -287,6 +281,15 @@ func main() {
...
 	}
 
+	if minRequeueTime < 5*time.Second {
+		setupLog.Error(fmt.Errorf("min-requeue-time (%s) must be at least 5s to avoid provider throttling", minRequeueTime), "invalid configuration")
+		os.Exit(1)
+	}
+	if maxRequeueTime > 15*time.Minute {
+		setupLog.Error(fmt.Errorf("max-requeue-time (%s) must not exceed 15m", maxRequeueTime), "invalid configuration")
+		os.Exit(1)
+	}
```
maksymvavilov: I'd imagine if the user wants to override this with something insane, they should be able to. Plus, for testing purposes, I'd put here seconds

CLAUDE.md
```diff
@@ -283,10 +282,9 @@ Common log metadata:
 - Prevents conflicts when multiple operators manage same zone
 
 **Reconciliation Flow:**
-- Records marked valid for 14m by default
 - Min requeue time: 5s (prevents API throttling)
 - Max requeue time: 15m (ensures eventual consistency)
-- Exponential backoff between min and max
+- Exponential backoff via SDK rate limiter between min and max
```
mikenairn: Probably no longer valid

cmd/main.go
mikenairn, line 291: Seems reasonable to add validation here but should it be documented?

internal/controller/base_dnsrecord_reconciler.go
```diff
@@ -218,20 +221,20 @@ func (r *BaseDNSRecordReconciler) updateStatus(ctx context.Context, client clien
...
+// and returns a reconcile.Result that re-queues after the given duration.
+func (r *BaseDNSRecordReconciler) updateStatusAndRequeue(ctx context.Context, c client.Client, previous, current DNSRecordAccessor, requeueAfter time.Duration) (reconcile.Result, error) {
 	logger := log.FromContext(ctx)
 	patch := client.MergeFrom(previous.GetDNSRecord())
 	// update the record after setting the status
 	if !equality.Semantic.DeepEqual(previous.GetStatus(), current.GetStatus()) {
 		if updateError := c.Status().Patch(ctx, current.GetDNSRecord(), patch); updateError != nil {
 			if apierrors.IsConflict(updateError) {
-				return ctrl.Result{RequeueAfter: time.Second}, nil
+				return ctrl.Result{RequeueAfter: defaultValidationRequeue}, nil
```
mikenairn: What is `defaultValidationRequeue`? This was just set to a second since it should just try again straight away. Maybe this is fine, but not if its a really long time which i can't tell from here.

internal/controller/dnsrecord_controller.go
```diff
@@ -317,8 +311,21 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+			dnsRecord.SetStatusCondition(string(v1alpha1.ConditionTypeReady), metav1.ConditionFalse,
+				string(v1alpha1.ConditionReasonProviderError), fmt.Sprintf("The dns provider could not be loaded: %v", err))
+			return r.updateStatus(ctx, previous, dnsRecord, false, err)
+		}
+		if _, err := r.publishRecord(ctx, dnsRecord, dnsProvider); err != nil {
+			logger.Error(err, "Failed to update registry for inactive group")
+			dnsRecord.SetStatusCondition(string(v1alpha1.ConditionTypeReady), metav1.ConditionFalse,
+				"RegistryError", fmt.Sprintf("Failed to update registry for inactive group: %v", err))
+			return r.updateStatus(ctx, previous, dnsRecord, false, err)
+		}
```
mikenairn: I don't think you need any of this? Can we not just remove all of this `dnsRecord.IsActive()` condition and have it update what needs updated through the normal `publishRecord` call below?

The requeue time is `InactiveGroupRequeueTime` which is every 15 seconds. Is there a need to poll that often?

internal/controller/dnsrecord_controller.go
```diff
@@ -382,86 +379,61 @@ func (r *DNSRecordReconciler) updateStatus(ctx context.Context, previous, curren
...
+	return r.updateStatusAndRequeue(ctx, r.Client, previous, current, requeueAfter)
 }
 
 // SetupWithManager sets up the controller with the Manager.
-func (r *DNSRecordReconciler) SetupWithManager(mgr ctrl.Manager, maxRequeue, validForDuration, minRequeue time.Duration, healthProbesEnabled, allowInsecureHealthCert bool) error {
+func (r *DNSRecordReconciler) SetupWithManager(mgr ctrl.Manager, maxRequeue, minRequeue time.Duration, healthProbesEnabled, allowInsecureHealthCert bool) error {
+	if minRequeue > maxRequeue {
+		return fmt.Errorf("minRequeue (%s) must not exceed maxRequeue (%s)", minRequeue, maxRequeue)
+	}
+
```
mikenairn: Any reason we validate this here, but do other validations on these values in main?

internal/controller/dnsrecord_controller.go
```diff
@@ -382,86 +379,61 @@ func (r *DNSRecordReconciler) updateStatus(ctx context.Context, previous, curren
...
 	defaultRequeueTime = maxRequeue
-	validFor = validForDuration
 	defaultValidationRequeue = minRequeue
 	probesEnabled = healthProbesEnabled
 	allowInsecureCert = allowInsecureHealthCert
 
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&v1alpha1.DNSRecord{}).
+		For(&v1alpha1.DNSRecord{}, builder.WithPredicates(specOrDeletionChangedPredicate{})).
```
mikenairn: Could `predicate.GenerationChangedPredicate{}` be used here? You might need additional predicate for the delete, not sure.

My understanding of this change is to prevent reconciles when the status updates (we do it a lot) is that correct?

internal/controller/dnsrecord_controller.go
```diff
@@ -507,73 +479,32 @@ func (r *DNSRecordReconciler) SetupWithManager(mgr ctrl.Manager, maxRequeue, val
...
+	}
+	if e.ObjectNew.GetGeneration() != e.ObjectOld.GetGeneration() {
+		return true
 	}
-	// double the duration. Return the max timeout if overshoot
-	newRequeue := lastRequeue * 2
-	if newRequeue > defaultRequeueTime {
-		return defaultRequeueTime
+	if e.ObjectNew.GetDeletionTimestamp() != nil {
+		return true
```
mikenairn: I think this will cause all status updates to trigger a reconcile when deleting, should it check for the timestamp changing?

internal/controller/dnsrecord_controller.go
```diff
@@ -382,86 +379,61 @@ func (r *DNSRecordReconciler) updateStatus(ctx context.Context, previous, curren
...
-			requeueTime = exponentialRequeueTime(current.GetStatus().ValidFor)
-
-			// reset requeue time if we changed healthcheck spec but no updates were needed to the provider
-			if generationChanged(current.GetDNSRecord()) {
-				requeueTime = defaultValidationRequeue
-			}
+			// stable — derive backoff from time since last condition transition
+			requeueAfter = min(max(defaultValidationRequeue, time.Since(readyCond.LastTransitionTime.Time)), defaultRequeueTime)
 		}
 	}
```
mikenairn, line 401: Should this logic be centralised somewhere so it can be re-used and tested properly (unit tests)?

internal/controller/dnsrecord_controller.go
```diff
@@ -382,86 +379,61 @@ func (r *DNSRecordReconciler) updateStatus(ctx context.Context, previous, curren
...
-			// we are giving precedence to AwaitingValidation
-			// meaning we are doubling not randomized value
-			requeueTime = exponentialRequeueTime(current.GetStatus().ValidFor)
-
-			// reset requeue time if we changed healthcheck spec but no updates were needed to the provider
-			if generationChanged(current.GetDNSRecord()) {
-				requeueTime = defaultValidationRequeue
-			}
+			// stable — derive backoff from time since last condition transition
+			requeueAfter = min(max(defaultValidationRequeue, time.Since(readyCond.LastTransitionTime.Time)), defaultRequeueTime)
```
maksymvavilov: This is elegant! I would, however, have it separated into a tiny function elsewhere. Two reasons: we could test it and, more importantly, explain what it does. 3 different values and a number or min/max gave me a pause to figure out what this is. 
I was debating whether this should be tested or not for a while, and it is, technically, tested in a "conflict" integration test, so maybe separating it for the sake of making it more "easy to digest" is not enough of a reason

maksymvavilov (dismissed):

internal/controller/base_dnsrecord_reconciler.go
philbrookes: I know, the names are confusing. I had to find where they came from as well. This is defaultValidationRequeue = minRequeueTime, and defaultRequeueTime = maxRequeueTime.

I'd be happy to rename them, just didn't want to get into that as part of this PR.

internal/controller/dnsrecord_controller.go
philbrookes: The inactive requeue needs to be more frequent than the usual backoff, as this is essentially the limiting factor on failover speed, plus it's quite cheap, just doing a DNS lookup on the active groups TXT record, and testing that it's group entries in the registry are accurate.

I might be missing something here, but if I change this to use the regular publish code below, the inactive groups would back off until 15m intervals?

internal/controller/dnsrecord_controller.go
philbrookes: I did originally write this to it's own function and it felt like overkill. I googled "idiomatic go clamp" and this format was the most idiomatic.

internal/controller/dnsrecord_controller.go
philbrookes: do you mean the requeueAfter = min max...? If so, I don't see much value in changing this, from:
```
requeueAfter = min(max(defaultValidationRequeue, time.Since(readyCond.LastTransitionTime.Time)), defaultRequeueTime)
```
to:
```
requeueAfter = requeueBackoff(readyCond.LastTransitionTime.Time, defaultValidationRequeue, defaultRequeueTime)
...
func requeueBackoff(lastTransition time.Time, minRequeue, maxRequeue time.Duration) time.Duration {
	return min(max(minRequeue, time.Since(lastTransition)), maxRequeue)
}
```

cmd/main.go
philbrookes: I'm going to remove it, to allow more flexible testing setup (i.e. minRequeue of a few milliseconds in integration tests, etc. to avoid long timeouts).

internal/controller/dnsrecord_controller.go
philbrookes: discussing this with Mike here, too: https://github.com/Kuadrant/dns-operator/pull/809/changes#r3748770328

internal/controller/dnsrecord_controller.go
philbrookes: yeah, that's right.

I'll remove the custom predicate and update this to use:
```
  builder.WithPredicates(predicate.Or(
      predicate.GenerationChangedPredicate{},
      deletionPredicate{},
  ))
```

internal/controller/dnsrecord_controller.go
philbrookes: yeah, we want that on a deleting object right? That's the same thing CodeRabbit said here: https://github.com/Kuadrant/dns-operator/pull/809/changes#r3712716960

cmd/main.go
```diff
@@ -287,6 +281,11 @@ func main() {
 		os.Exit(1)
 	}
 
+	if minRequeueTime > maxRequeueTime {
+		setupLog.Error(fmt.Errorf("min-requeue-time (%s) must not exceed max-requeue-time (%s)", minRequeueTime, maxRequeueTime), "invalid configuration")
+		os.Exit(1)
+	}
+
```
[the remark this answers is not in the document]
philbrookes: I have addressed this now.

internal/controller/dnsrecord_controller.go
mikenairn: Something like that, so you can test it and use it outside of the dnsrecord controller. It's only being used in the dnsrecord controller, but you could define it in the in the base. Should remote records be using this also? I'm not sure if they should, they obviously didn't before, but that was more because it was so convoluted and required those additional status fields.

internal/controller/base_dnsrecord_reconciler.go
mikenairn: OK, i would say we should rename all these, in a different PR, the names are just confusing.

I'm still not sure about using a variable for this, the explicit `time.Second` for these cases made sense to me.

internal/controller/dnsrecord_controller.go
mikenairn: OK, I'm not convinced you couldn't leave this to the normal publish, and alter the status update based on !dnsRecord.IsActive(). That said, this should work. I would however consider just moving the  `getDNSProvider` setup above here so you don't need to duplicate that at least https://github.com/Kuadrant/dns-operator/pull/809/files#diff-adacb6a4589c65c2d853c0f59ca02d6b8fb50bb388b41de1d13e99b0d128cfe1R333-R339

internal/external-dns/registry/group.go
```diff
@@ -19,13 +20,33 @@ var _ registry.Registry = &GroupRegistry{}
...
+				newEp.SetIdentifier == oldEp.SetIdentifier &&
+				newEp.RecordTTL == oldEp.RecordTTL &&
+				reflect.DeepEqual(newEp.ProviderSpecific, oldEp.ProviderSpecific) {
+				filteredOld = append(filteredOld, oldEp)
+				filteredNew = append(filteredNew, newEp)
+			}
+		}
+		changes.UpdateOld = filteredOld
+		changes.UpdateNew = filteredNew
+	}
```
mikenairn, line 53: Small comment on why this is doing this might be useful for 6 months time

internal/controller/dnsrecord_controller.go
philbrookes: ok, you and @maksymvavilov both feel the same way about this, so I'll move it to a small function and add some tests.

internal/controller/dnshealthcheckprobe_reconciler.go
```diff
@@ -72,7 +72,7 @@ func (r *DNSProbeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 		controllerutil.RemoveFinalizer(dnsProbe, DNSHealthCheckFinalizer)
 		if err = r.Update(ctx, dnsProbe); client.IgnoreNotFound(err) != nil {
 			if apierrors.IsConflict(err) {
-				return ctrl.Result{Requeue: true}, nil
+				return ctrl.Result{RequeueAfter: 100 * time.Millisecond}, nil
```
mikenairn: nit: probably should be consistent in what we are using for these conflict specific `RequeAfter` values https://github.com/Kuadrant/dns-operator/pull/809/files#diff-1e120ba24b1b5c9323db706038e38899f10a77d61d4a9c568683a7161a74975aR232

`time.Second` always seemed fine to me :shrug:

internal/controller/dnsrecord_controller.go
```diff
@@ -507,73 +467,28 @@ func (r *DNSRecordReconciler) SetupWithManager(mgr ctrl.Manager, maxRequeue, val
...
-			}
-		}
-		// prematurely is true here. return false in case we need full reconcile
-		return isHealthy == allProbesHealthy, requeueIn
-	}
-
-	return prematurely, requeueIn
+// ClampTime clamps desiredTime between minTime and maxTime.
+func ClampTime(minTime, maxTime, desiredTime time.Duration) time.Duration {
+	return min(max(minTime, desiredTime), maxTime)
```
mikenairn: nit: Could be in `base_dnsrecord_reconciler` for re-use by all reconcilers (Although they could from here anyway)

mikenairn (commented):
Looks fine, will approve once [this](https://github.com/Kuadrant/dns-operator/pull/810) is merged as i want to use this to test that helm chart sync.

mikenairn (dismissed):

mikenairn (approved):

PR #815, opened 2026-08-13
CLI dry-run mode for remove-active-group

### LINKED ISSUE #782: CLI dry-run mode for remove-active-group

Add a --dry-run or impact preview mode to the remove-active-group CLI command (#663) that uses the registry data structure to list which DNS endpoints will be removed or modified, with user confirmation before proceeding.

philbrookes:
## Summary
- Adds a `--dry-run` flag to the `remove-active-group` CLI command that previews the impact of removing a failover group without making DNS changes
- Uses the TXT registry data structure to analyse which DNS endpoints would be deleted or modified, displaying results in a table
- Refactors confirmation prompt to use the shared `inputYes` helper for consistency with `add-active-group`

Closes #782

## Test plan
- [x] Unit tests for `AnalyseGroupRemovalImpact` covering: empty endpoints, single-group DELETE, remaining-group no-impact, mixed-group MODIFY, ungrouped-owner preservation, non-managed record type filtering, no-registry-entry filtering, multiple endpoints
- [x] `RemainingGroups` field assertions in all test cases
- [x] `go build`, `go vet`, `make test-unit` all pass
- [x] Verify `--dry-run` flag is registered with help text
- [x] Verify no DNS mutations occur in dry-run path (`continue` before `ApplyChanges`)
- [ ] Manual test: run `remove-active-group --dry-run` against a real provider to verify output formatting

## Summary by CodeRabbit

* **New Features**
  * Added a dry-run option to preview active-group removal without making changes.
  * Preview results show whether DNS TXT records would be deleted or updated, along with affected endpoints.
* **Bug Fixes**
  * Legacy, unrecognised and irrelevant TXT records are now safely skipped.
  * Targets shared with active or ungrouped records are retained during inactive-group cleanup.
  * Improved interactive confirmation handling for active-group removal.
  * Improved parsing of target lists in registry records.
* **Tests**
  * Added coverage for endpoint deletion, updates, retention, ignored records and multiple affected endpoints.

### DIFF
--- a/cmd/plugin/failover/shared.go
+++ b/cmd/plugin/failover/shared.go
@@ -13,6 +13,9 @@ import (
 	"sigs.k8s.io/external-dns/endpoint"
 
 	"github.com/kuadrant/dns-operator/cmd/plugin/common"
+	internalcommon "github.com/kuadrant/dns-operator/internal/common"
+	"github.com/kuadrant/dns-operator/internal/external-dns/registry"
+	"github.com/kuadrant/dns-operator/types"
 )
 
 const (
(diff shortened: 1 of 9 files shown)


cmd/plugin/failover/shared.go
```diff
@@ -141,3 +143,98 @@ func GetDomainRegexp(domain string) (*regexp.Regexp, error) {
...
 	return domainRegexp, nil
 }
+
+// Must be kept in sync with internal/controller/dnsrecord_controller.go
+const (
+	txtRegistryPrefix              = "kuadrant-"
+	txtRegistrySuffix              = ""
+	txtRegistryWildcardReplacement = "wildcard"
+	txtRegistryEncryptAESKey       = ""
+)
```
DaliborD45, line 153: Wouldnt it be better to export these constants from one file, so we keep single source of truth ?

cmd/plugin/failover/shared.go
```diff
@@ -141,3 +143,98 @@ func GetDomainRegexp(domain string) (*regexp.Regexp, error) {
...
+		registryHost, ok := registryMap.Hosts[ep.DNSName]
+		if !ok {
+			continue
+		}
+
+		if !registryHost.HasAnyGroup(postRemovalGroups) && len(registryHost.UngroupedOwners) == 0 {
+			impacts = append(impacts, EndpointImpact{
+				DNSName:    ep.DNSName,
+				RecordType: ep.RecordType,
+				Action:     "DELETE",
```
DaliborD45:
```
type Action string

const (
    Delete Action = "DELETE"
    Modify Action = "MODIFY"
)

type EndpointImpact struct {
    DNSName    string
    RecordType string
    Action     Action
    OldTargets []string
    NewTargets []string
}
```
Than, you can use: `Action: Delete` instead of writing constant every time, so when you change the variable value, it propagates everywhere. I dont know how much idiomatic is this in GO though.

DaliborD45 (changes requested):

internal/controller/dnsrecord_groups.go
```diff
@@ -420,13 +421,13 @@ func (r *BaseDNSRecordReconciler) unpublishInactiveGroups(ctx context.Context, c
...
-		// inactive targets is the targets of all inactive groups
 		inactiveTargets := registryHost.GetOtherGroupsTargets(activeGroups)
+		activeTargets := registryHost.GetGroupsTargets(activeGroups)
+		ungroupedTargets := registryHost.GetUngroupedTargets()
 
-		// remove all inactive targets from endpoint
 		newTargets := []string{}
 		for _, t := range endpoint.Targets {
-			if !slices.Contains(inactiveTargets, t) {
+			if !slices.Contains(inactiveTargets, t) || slices.Contains(activeTargets, t) || slices.Contains(ungroupedTargets, t) {
```
DaliborD45, line 430: This if condition is the very same condition as in `shared.go` file. This should be refactored into one function and both ifs should use that same function, because we are trying to mirror the behaviour of actual run. If someone changed the condition in one file, not changing it in other place would cause a bug

internal/controller/dnsrecord_groups.go
```diff
@@ -420,13 +421,13 @@ func (r *BaseDNSRecordReconciler) unpublishInactiveGroups(ctx context.Context, c
...
 		}
 
-		// inactive targets is the targets of all inactive groups
 		inactiveTargets := registryHost.GetOtherGroupsTargets(activeGroups)
+		activeTargets := registryHost.GetGroupsTargets(activeGroups)
+		ungroupedTargets := registryHost.GetUngroupedTargets()
 
-		// remove all inactive targets from endpoint
 		newTargets := []string{}
 		for _, t := range endpoint.Targets {
```
DaliborD45, line 429: Also here we are changing controller behavior, which seems to me is kinda out of scope of this PR, so maybe it would be nice to mention it in the Title of the PR, or even better add it in separate commit (maybe I am just nitpicking at this point, so sorry for that :D )

DaliborD45 (changes requested):

DaliborD45 (approved):

PR #824, opened 2026-08-19
fix: allow metadata changes through watch predicate

philbrookes:
## Summary

- Replaces `GenerationChangedPredicate` with `specOrMetadataChangedPredicate` in the DNSRecord controller watch setup
- Allows finalizer and label updates to trigger immediate reconciliation instead of waiting for 5s requeue delays

The `GenerationChangedPredicate` (introduced in #809) filtered all non-spec update events, including metadata-only changes like finalizer and label additions. This caused the reconciler to rely on `RequeueAfter` delays between each intermediate step (finalizer → labels → publish+status), resulting in a ~15s minimum before `OwnerID` appeared in status — exceeding downstream test timeouts.

The new predicate allows updates through when generation, finalizers, labels, or annotations change, while still filtering status-only updates to prevent tight reconciliation loops. This is consistent with #821 which fixed the same class of issue for the deletion path.

## Test plan

- [x] dns-operator unit tests pass
- [x] All 31 kuadrant-operator DNS integration tests pass (previously failing test now completes in ~6.6s)

## Summary by CodeRabbit

- **Bug Fixes**
  - DNS record updates now reconcile correctly when relevant metadata, labels, annotations, or finaliser changes occur.
  - Status-only updates continue to be ignored, reducing unnecessary processing.
  - Deletion handling now reflects the updated update-event behaviour.

### DIFF
--- a/internal/controller/dnsrecord_controller.go
+++ b/internal/controller/dnsrecord_controller.go
@@ -20,6 +20,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"maps"
+	"slices"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


mikenairn (approved):
Tested locally, fixes the kuadrant-operator test issues.

Saying that, i am a bit concerned we aren't just getting them working now because we might have it cycle round quicker due to other changes. Some of the updates are just in the status, ownerID etc.., which should really cause it to come back round straight away after it returns. If i understand the current logic correctly this won't happen for anything that just update the status.

philbrookes:
> Tested locally, fixes the kuadrant-operator test issues.
> 
> Saying that, i am a bit concerned we aren't just getting them working now because we might have it cycle round quicker due to other changes. Some of the updates are just in the status, ownerID etc.., which should really cause it to come back round straight away after it returns. If i understand the current logic correctly this won't happen for anything that just update the status.

The problem was adding the finalizer and then labels (and no generation change) made the backoff logic take too long for the tests, changing these predicates means that those changes cause an immediate reconcile too.

Having said that, I don't see any reason not to also reconcile on status updates.

PR #826, opened 2026-08-20
Fix domain filtering and TXT record quotes for failover cli commands

averevki:
## Fixes

**Domain filtering fix (GCP, Azure):** The per-zone provider in `add-active-group` and `remove-active-group` CLI commands used `RegexDomainFilter` with pattern `^zone.example.com$`, which only matches the exact zone name. The active groups TXT record name is `kuadrant-active-groups.zone.example.com` — a subdomain that doesn't match the regex. Azure's `UpdateRecords()`/`DeleteRecords()` silently skips records that don't match the domain filter, so writes appeared to succeed but never actually applied.  GCP had the same filtering issue. Switched to list-based `DomainFilter{Filters: []string{domain}}` which matches subdomains. The `get-active-groups` command already used the correct filter.

**TXT record quoting fix (AWS):** The shared failover code wrapped TXT record values in literal double quotes (`"\"version=1;groups=...\""`). AWS Route53 requires this quoting at the API level, but GCP and Azure APIs accept raw, with literal quotes get stored as data, corrupting the value. Moved quoting responsibility into the AWS provider: `quoteTXTRecord()` ensures values are quoted before writing to Route53, `stripTXTQuotes()` normalizes them when reading back.

## Verification

```bash
# Before fix: add-active-group reports success on GCP/Azure but record is never created
kubectl-kuadrant_dns add-active-group test-group -d gcp.example.com --providerRef kuadrant/gcp-credentials -y
# added group "test-group" to active groups of "azure.example.com" zone
kubectl-kuadrant_dns get-active-groups -d gcp.example.com --providerRef kuadrant/gcp-credentials
# Output: no active groups found

# After fix: add-active-group correctly creates the record
kubectl-kuadrant_dns add-active-group test-group -d azure.example.com --providerRef kuadrant/azure-credentials -y
# added group "test-group" to active groups of "azure.example.com" zone
kubectl-kuadrant_dns get-active-groups -d azure.example.com --providerRef kuadrant/azure-credentials
# Output: `test-group` shown as an active group
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved failover handling by applying exact domain matching for per-zone provider operations.
  * Corrected endpoint retrieval error messaging.
  * Standardised failover TXT record formatting, including compatibility with existing quoted records.
  * Improved Route 53 TXT record handling by normalising values when reading and formatting them correctly when updating.
* **Tests**
  * Added coverage for quoted and unquoted TXT record values and legacy-version compatibility.

### DIFF
--- a/internal/external-dns/provider/aws/aws.go
+++ b/internal/external-dns/provider/aws/aws.go
@@ -405,7 +405,11 @@ func (p *AWSProvider) records(ctx context.Context, zones map[string]types.Hosted
 				if len(r.ResourceRecords) > 0 {
 					targets := make([]string, len(r.ResourceRecords))
 					for idx, rr := range r.ResourceRecords {
-						targets[idx] = aws.ToString(rr.Value)
+						val := aws.ToString(rr.Value)
+						if r.Type == types.RRTypeTxt {
+							val = stripTXTQuotes(val)
+						}
+						targets[idx] = val
 					}
 
 					ep := endpoint.NewEndpointWithTTL(wildcardUnescape(aws.ToString(r.Name)), string(r.Type), ttl, targets...)
@@ -740,6 +744,9 @@ func (p *AWSProvider) newChange(action types.ChangeAction, ep *endpoint.Endpoint
 		}
 		change.ResourceRecordSet.ResourceRecords = make([]types.ResourceRecord, len(ep.Targets))
 		for idx, val := range ep.Targets {
+			if change.ResourceRecordSet.Type == types.RRTypeTxt {
+				val = quoteTXTRecord(val)
+			}
 			change.ResourceRecordSet.ResourceRecords[idx] = types.ResourceRecord{
 				Value: aws.String(val),
 			}
@@ -1063,4 +1070,18 @@ func zoneTypeString(z *types.HostedZone) string {
 		return "public"
 	}
 	return "private"
+}
+
+func quoteTXTRecord(val string) string {
+	if strings.HasPrefix(val, "\"") && strings.HasSuffix(val, "\"") {
+		return val
+	}
+	return "\"" + val + "\""
+}
+
+func stripTXTQuotes(val string) string {
+	if strings.HasPrefix(val, "\"") && strings.HasSuffix(val, "\"") {
+		return val[1 : len(val)-1]
+	}
+	return val
 }
--- a/cmd/plugin/failover/add-active-group.go
+++ b/cmd/plugin/failover/add-active-group.go
@@ -106,7 +106,7 @@ func addActiveGroup(_ *cobra.Command, args []string) error {
 		var endpoints []*externaldnsendpoint.Endpoint
 
 		providerForZone, err = common.GetProviderForConfig(ctx, resourceRef, provider.Config{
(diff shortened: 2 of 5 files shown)


internal/external-dns/provider/aws/aws.go
```diff
@@ -1064,3 +1071,17 @@ func zoneTypeString(z *types.HostedZone) string {
 	}
 	return "private"
 }
+
+func quoteTXTRecord(val string) string {
+	if strings.HasPrefix(val, "\"") && strings.HasSuffix(val, "\"") {
+		return val
+	}
+	return "\"" + val + "\""
```
philbrookes: would this be better as:
```go
if !strings.HasPrefix(val, "\"") {
	val = "\"" + val
}
if !strings.HasSuffix(val, "\"") {
	val = val + "\""
}
return val
```

internal/external-dns/provider/aws/aws.go
```diff
@@ -1064,3 +1071,17 @@ func zoneTypeString(z *types.HostedZone) string {
...
+}
+
+func stripTXTQuotes(val string) string {
+	if strings.HasPrefix(val, "\"") && strings.HasSuffix(val, "\"") {
```
philbrookes: I think this would be better as:
```go
func stripTXTQuotes(val string) string {
    return strings.Trim(val, "\"")
}
```

Though I'm not convinced we need a whole function for this, it could just be done in place?

philbrookes: Example here: https://go.dev/play/p/B6fRgYzOi0L

philbrookes (commented):
A few comments, but generally fine.

philbrookes (approved):

philbrookes (approved):

internal/external-dns/provider/aws/aws.go
```diff
@@ -405,7 +405,12 @@ func (p *AWSProvider) records(ctx context.Context, zones map[string]types.Hosted
 				if len(r.ResourceRecords) > 0 {
 					targets := make([]string, len(r.ResourceRecords))
 					for idx, rr := range r.ResourceRecords {
-						targets[idx] = aws.ToString(rr.Value)
+						val := aws.ToString(rr.Value)
+						if r.Type == types.RRTypeTxt {
+							// route53 stores TXT values quoted
+							val = strings.Trim(val, "\"")
```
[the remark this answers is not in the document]
averevki: This can be valid, but rather an edge-case with quotations placement not introduced by this PR. Might be worth fixing in the future

internal/external-dns/provider/aws/aws.go
```diff
@@ -740,6 +745,10 @@ func (p *AWSProvider) newChange(action types.ChangeAction, ep *endpoint.Endpoint
 		}
 		change.ResourceRecordSet.ResourceRecords = make([]types.ResourceRecord, len(ep.Targets))
 		for idx, val := range ep.Targets {
+			if change.ResourceRecordSet.Type == types.RRTypeTxt {
+				// route53 requires TXT values to be quoted
+				val = "\"" + strings.Trim(val, "\"") + "\""
```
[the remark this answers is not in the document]
averevki: I would consider this group name not valid. Might be worth fixing in the future, not introduced by this PR

